PR #690, opened 2014-12-16
test(rxSortableColumn): All-column functions

Droogans:
We seem to be experimenting some techmological differences. View the travis logs to see more, as well as the long-form comments in https://github.com/rackerlabs/encore-ui/commit/c298e4fade292e821a5b3958498e3727a43324ab

Useful for when you need all column names, or their sorts by table.

Includes some refactoring for column sorting, which now gets reused.


parlarjb:
LGTM

PR #716, opened 2015-01-08
test(rxForm): Prefer options' text over allOptions

Droogans:
I have no idea why I exposed an `allOptions` call that returned every option as an object. This replaces that with a more suitable `options` call that returns all option's text, and does away with `allOptions` altogether.

This does make it a breaking change, but I'm almost certain no one was using it before. It can wait until other breaking changes are planned for a release to avoid excessive version increments.


parlarjb:
Can you update the commit message so it contains a "BREAKING CHANGE" footer, as described here? https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.gbbngquhe0qa

This will ensure that the auto-generated change log makes mention of the breaking change.

parlarjb:
LGTM

PR #720, opened 2015-01-09
feat(rxDOMHelpers): add a ".find" method

halkon:
Adds a `.find` method that allows the developer to use selectors supported via `element.querySelector`.

https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_Started/Selectors


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/720/gallery.html)

parlarjb:
LGTM

PR #722, opened 2015-01-11
E2e multiline comments everywhere, with minor breaking changes

Droogans:
This started as a chore to clean up comments, but it ended up becoming more. I realized that my editor makes it easy to make multiline comments by creating many single-line comments, and that this is a bit hard on the eyes. I'm switching over to multiline comments because they look better.

Also, since I submitted a pull request with breaking changes in it earlier this week, I thought I might as well get as many of them as I can for the minor version change that the first will cause.

### DIFF
--- a/src/rxApp/rxApp.page.js
+++ b/src/rxApp/rxApp.page.js
@@ -9,32 +9,16 @@ var rxApp = {
         get: function () { return '.collapsible-toggle'; }
     },
 
-    lblNavTitle: {
-        get: function () { return this.rootElement.$('.site-title'); }
-    },
-
-    lblNavSectionTitle: {
-        get: function () { return this.rootElement.$('.nav-section-title'); }
-    },
-
-    eleSiteNav: {
-        get: function () { return this.rootElement.$('.rx-app-nav'); }
-    },
-
     btnCollapseToggle: {
         get: function () { return this.rootElement.$(this.cssCollapseButtonSelector); }
     },
 
-    lnkLogout: {
-        get: function () { return this.rootElement.$('.site-logout'); }
-    },
-
     title: {
-        get: function () { return this.lblNavTitle.getText(); }
+        get: function () { return this.rootElement.$('.site-title').getText(); }
     },
 
     sectionTitle: {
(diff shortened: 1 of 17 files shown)


src/rxApp/rxApp.page.js
```diff
@@ -4,37 +4,23 @@ var Page = require('astrolabe').Page;
...
         get: function () { return this.rootElement.$(this.cssCollapseButtonSelector); }
     },
 
-    lnkLogout: {
-        get: function () { return this.rootElement.$('.site-logout'); }
-    },
-
     title: {
-        get: function () { return this.lblNavTitle.getText(); }
+        get: function () { return this.rootElement.$('.site-title').getText(); }
```
parlarjb: What's your thinking behind this change?

parlarjb:
Can you make sure the final squashed commit message contains a "BREAKING CHANGE" footer, as described here? https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.gbbngquhe0qa

This will ensure that the auto-generated change log makes mention of the breaking change.

src/rxApp/rxApp.page.js
Droogans: The commit message is a bit buried in the changeset...here's the relevant snippet:

> Only use element properties on a page object if the element is used in more than one place in the functions exposed by the page object. In cases where this is only one reference to a button, link, etc., just use an inline element selector.

Droogans:
@parlarjb I made it `BREAKING CHANGES` (plural). Is that going to be seen by the change log generator?

parlarjb:
I don't think so. https://github.com/ajoslin/conventional-changelog/blob/master/lib/git.js#L97

Droogans:
@parlarjb I just changed it to two separate `BREAKING CHANGE:` paragraphs in one commit, and on second reading of the source code you linked to, I don't think it's going to catch that either.

All in all, I think it's fine. The end result is that one commit will have one breaking change, which is the way it should be, really. I'll update that commit and split it up.

parlarjb:
LGTM

PR #724, opened 2015-01-12
E2e rx option form table

Droogans:


parlarjb:
@Droogans Could you fix the merge conflict on this?

src/rxForm/rxForm.page.js
```diff
@@ -171,3 +221,286 @@ exports.rxForm = {
...
+        }
+    },
+
+    selections: {
+        /*
+          Return a list of row indexes that are currently selected.
+          Get the row yourself if you need more information about the row's contents.
+        */
+        get: function () {
+            return this.tblRows.map(function (rowElement, index) {
```
parlarjb: Why didn't you just do a filter here, instead of `reject(map(...))`?

Droogans: If I were to go with something like

```
return this.tblRows.filter(optionIsSelected);
```

I would not be getting the index back. If I were to expand it a little to get the index, I could only use it to evaluate whether or not I'd need to keep it.

```
return this.tblRows.filter(function (rowElement, index) {
    return optionIsSelected(rowElement).then(function (selected) {
        if (selected) {
            return index; // is this a conditional? that's all .filter cares about
        }
    });
});
```

`.map` allows me to iterate over a list, and return a custom value back for each "good" value. `filter` only gives back a list of web elements. The Protractor docs show the difference in the "Returns" section at the bottom of each of these pages.

http://angular.github.io/protractor/#/api?view=ElementArrayFinder.prototype.map

http://angular.github.io/protractor/#/api?view=ElementArrayFinder.prototype.filter

Droogans: I should also elaborate a bit on why I went with index numbers instead of the row objects themselves. I thought most people (?) would only care that row number `{X}` was selected by default or something, and if they wanted the cell value for some row (which could be variable based on the test), you could create the row yourself and ask if `row.cell('Created By')` equals something you care about.

That's definitely up for debate. Tagging @jj4th and @SyntaxStacks for their input here.

parlarjb: Sorry, I totally blanked on the fact that you were returning the index. Makes sense.

src/rxForm/templates/rxFormOptionTable.html
```diff
@@ -29,7 +29,7 @@
                 </label>
             </td>
             <td ng-repeat="column in columns">
-                <label for="{{fieldId}}_{{$parent.$index}}">
+                <label for="{{column.label}}_{{$parent.$index}}">
```
parlarjb: Have you checked if this could break anyone?

Droogans: Nope. I doubt any developers have unit tests that rely on this configuration, as it's not very informative as is. As far as I know, this pattern is not used by any testers for that very reason, so I am nearly certain that no one is relying on it there. If they were, they could replace their entire approach with one line from these page objects.

I'll update my repo clones and grep through any usage of the directive, and see if I can't glean any hints about code dependent on this.

parlarjb:
LGTM

jj4th:
Andrew asked me to do the merge on this since I already looked at it... QE-LGTM!

PR #725, opened 2015-01-12
Account info under breadcrumbs

parlarjb:
Previously `rxAccountInfo` was a big blue box that you could place anywhere on the page. That version is still supported to give people a chance to move this version, but now if you pass `account-info-banner="true"` to the directive, it will draw with the new style.

You should no longer explicitly place `<rx-account-info>` on your pages. It is now part of `rxPage`, and will automatically be inserted if you pass `account-number="..."` to `<rx-page>`

When used on a Cloud page, the account info will look like this:
screen shot 2015-01-12 at 4 37 00 pm

When used on any other page, it will be drawn without the user select drop-down
screen shot 2015-01-12 at 4 39 45 pm

### DIFF
--- a/src/rxAccountInfo/README.md
+++ b/src/rxAccountInfo/README.md
@@ -1,9 +1,10 @@
 [![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
 
-Functionality for populating an `rxInfoPanel` with a user's account details.
+This component is used to draw an account info box at the top of each page, directly underneath the breadcrumbs. `rxPage` (through `rxApp`) integrates it directly into its template, and you activate it by passing `account-number="..."` to `<rx-page>`.
 
-This component is very specific to internal Rackspace needs, but also lives as a demonstration on how to make
-use of the `rxInfoBox`.
+While you could theoretically use this component elsewhere, its design and style were done with the intention of sitting underneath the breadcrumbs.
 
-This directive requires that `SupportAccount`, `Encore`, and `Teams` services are available. These are not provided by this project,
+When placed on a page that has `:user` in its route parameters, this component will also draw a drop-down user selector, to allow the Racker to change which user they're looking at for the given account. At this time, this user-selection is *only* available for products under Cloud. If you need it for additional products, please let us know. 
+
+This directive requires that `SupportAccount`, `Encore`, `AccountStatusGroup`, and `Teams` services are available. These are not provided by this project,
 but are available in an internal Rackspace repository.
--- a/src/rxAccountInfo/rxAccountInfo.js
+++ b/src/rxAccountInfo/rxAccountInfo.js
@@ -1,7 +1,31 @@
 angular.module('encore.ui.rxAccountInfo', [])
-.directive('rxAccountInfo', function (SupportAccount, Teams, Encore, rxNotify) {
+
+/**
+ * @ngdoc directive
+ * @name encore.ui.rxAccountInfo:rxAccountInfo
+ * @restrict E
+ * @scope
+ * @description
+ * Responsible for drawing an account info box.
+ *
+ * There are two different styles of account info box supported. The "old" one, which appears
+ * wherever you want it to be, and a new one that is intended to be placed underneath the breadcrumbs.
+ * To use the new one, pass `account-info-banner="true"` to this directive
+ *
+ * @param {string} accountNumber - The account number to load and retrieve data for
+ * @param {string} [teamId] - Optional team ID, used for loading team badges
+ * @param {string} [notifyStack] - Optional notifications stack to put errors on. Defaults to `page`
+ * @param {string} [accountInfoBanner] - Set to "true" to use the new under-the-breadcrumbs style
+ */
+.directive('rxAccountInfo', function (Teams, SupportAccount, Encore, rxNotify, encoreRoutes,
+                                    AccountStatusGroup) {
     return {
-        templateUrl: 'templates/rxAccountInfo.html',
+        templateUrl: function (elem, attr) {
+            if (attr.accountInfoBanner === 'true') {
+                return 'templates/rxAccountInfoBanner.html';
+            }
+            return 'templates/rxAccountInfo.html';
+        },
         restrict: 'E',
         transclude: true,
         scope: {
@@ -16,6 +40,12 @@ angular.module('encore.ui.rxAccountInfo', [])
                 return ['<span class="tooltip-header">', badge.name,
                         '</span><p>', badge.description, '</p>'].join('');
             };
+
+            // Currently, the only time we should show the `Current User` area is
+            // if the Racker is on the Cloud page
+            encoreRoutes.isActiveByKey('cloud').then(function (isCloud) {
+                scope.showCurrentUser = isCloud;
+            });
 
             scope.accountPageUrl = _.template('/accounts/<%= accountNumber %>', scope);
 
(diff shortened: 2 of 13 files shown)


src/rxAccountInfo/README.md
```diff
@@ -1,9 +1,10 @@
 [![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
 
-Functionality for populating an `rxInfoPanel` with a user's account details.
+This component is used to draw an account info box at the top of each page, directly underneath the breadcrumbs. `rxPage` (through `rxApp`) integrates it directly into its template, and you activate it by passing `account-number="..."` to `<rx-page>`.
```
halkon: I'll make this comment elsewhere, but wouldn't you be able to just change this to `show-account-info="true"` and you pull the `accountNumber` directly from the `$routeParams`?

halkon: Nevermind I think as I start reading through, you already have it set, and just allow to override?

parlarjb: We can't assume that it exists in `$routeParams`. In TQ, for instance, it's not there.

src/rxAccountInfo/rxAccountInfo.js
```diff
@@ -41,13 +71,37 @@ angular.module('encore.ui.rxAccountInfo', [])
...
+})
+
+/**
+ * @ngdoc service
+ * @name encore.ui.rxAccountInfo:rxAccountInfoStatusClass
+ * @description Given an account status string, this function returns
+ * the correct class to be used to style the status string.
+ *
+ */
+.factory('rxAccountInfoStatusClass', function () {
```
halkon: something makes me think this better belongs in `encore-ui-svcs` rather than `encore-ui`, however, it would be way easier to deploy this through as an `encore-ui` module. 

That being said it is generic enough to be in encore ui, just... :unamused:

src/rxAccountInfo/rxAccountInfo.less
```diff
@@ -22,3 +22,80 @@
...
+                    margin: 5px 2px;
+                }
+            }
+        }
+    }
+
+    ul.account-info-badges {
+        overflow: hidden;
+        li {
+        }
```
halkon: forget something?

src/rxAccountInfo/rxAccountInfo.js
parlarjb: That's a good idea. I didn't feel overly comfortable putting this here, I don't know why I didn't think about encore-ui-svcs.

And since `rxAccountInfo` already requires `encore-ui-svcs` to be present, we'll be fine.

glynnis:
My only thoughts from a design perspective is that the demos here don't actually show what this will look like in the context of the app -- i.e. the way breadcrumbs, account info, and rxNotifications will actually interact with each other visually. I think that might be worth noting and including a screen shot (as I assume it's difficult to mimic in a demo that doesn't have hierarchy re: breadcrumbs).

src/rxAccountInfo/rxAccountInfo.less
parlarjb: Paging @glynnis

glynnis: derpyderpderp. You can just remove the lines 98-99 (that deal with the empty `li`). Probably deleted a property and forgot to clean up. Sorry!

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

src/rxAccountInfo/README.md
halkon: you know... `scope.accountNumber || $routeParams.accountNumber || throw new Error('...` (Ok maybe not that, just saying maybe make it optional? though that may not really be necessary at this point)

src/rxAccountInfo/rxAccountInfo.js
halkon: :+1: if you put it there I'll review your PR there... may even do a release...

src/rxAccountInfo/rxAccountInfo.less
halkon: derp!

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

src/rxAccountInfo/README.md
parlarjb: I'm going to keep this as-is. I don't know that there's a benefit in implicitly grabbing the account number

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/725/gallery.html)

src/rxAccountInfo/README.md
halkon: :+1: just making your head hurt

src/rxAccountInfo/rxAccountInfo.js
```diff
@@ -1,7 +1,31 @@
...
+ */
+.directive('rxAccountInfo', function (Teams, SupportAccount, Encore, rxNotify, encoreRoutes,
+                                    AccountStatusGroup) {
     return {
-        templateUrl: 'templates/rxAccountInfo.html',
+        templateUrl: function (elem, attr) {
+            if (attr.accountInfoBanner === 'true') {
+                return 'templates/rxAccountInfoBanner.html';
+            }
+            return 'templates/rxAccountInfo.html';
```
halkon: Look at you fancy pants :jeans:

halkon:
LGTM

src/rxAccountInfo/rxAccountInfo.spec.js
```diff
@@ -56,15 +83,17 @@ describe('rxAccountInfo', function () {
 
         // load templates
         module('templates/rxAccountInfo.html');
+        module('templates/rxAccountInfoBanner.html');
 
         // Inject in angular constructs
         inject(function ($location, $rootScope, $compile, $q, Encore, SupportAccount,
```
TheNando: I'm a fan of using the pattern of _$serviceName_ in the inject function params, then $serviceName = _$serviceName in the body so it's always blatantly obvious that you are using the service. That or only specifying $injector in the params, then using $serviceName = $injector.get('$serviceName').

I know this code was already there and I'm not suggesting you change anything for your PR, but I'd like to get everyone's feed back and if we all agree, we can go on a spec cleaning spree to make sure everything is consistent.

TheNando:
LGTM, and in this case the 'G' stands for 'Great', but that only applies to the visual 'L' or 'Look'. The code also looks good to me.

src/rxAccountInfo/rxAccountInfo.spec.js
parlarjb: D'oh, I completely forgot that `_$serviceName_` is a thing. That's a much nicer idea, and I'll start doing everything like that from now on.

PR #727, opened 2015-01-13
test(rxStatusColumn): Page objects

Droogans:
BREAKING CHANGE: Completely overhauls the page objects for
rxStatusColumn, anything that was there before is almost certainly
invalid now. The only exception to this is the `.api` property.


parlarjb:
LGTM

PR #728, opened 2015-01-13
test(rxAccountInfo): Remove `.all` badges function

Droogans:
BREAKING CHANGE: Avoid the use of any function that evaluates every
single sub component into one large collection of objects. This replaces
it with a more modest "get all names" call for times when mock data must
be quickly verified.


parlarjb:
Thanks for this. I'll be pulling the work into https://github.com/rackerlabs/encore-ui/pull/725

PR #729, opened 2015-01-13
E2e rx notify no by type

Droogans:


src/rxNotify/docs/rxNotify.midway.js
```diff
@@ -51,65 +45,31 @@ describe('rxNotify', function () {
...
 
-            it('should all be type "info"', function () {
-                _.forEach(info, function (infoMessage) {
-                    expect(infoMessage.type).to.eventually.equal('info');
-                });
+            it('should have helpful information', function () {
+                expect(info.text).to.eventually.equal('Helpful Information');
             });
 
-            describe('first notification', function () {
```
parlarjb: How come you took this out?

Droogans: I was going through what was nearly the same set of tests twice. I have already covered the `text` property, `isDismissable` (true and false), as well as `hasSpinner` true and false in tests before/after this one. So I removed it.

I'm just testing each property and function in one place to save some time on end to end testing. This probably saved a couple of seconds here.

src/rxNotify/rxNotify.page.js
```diff
@@ -76,31 +76,31 @@ var rxNotify = {
         }
     },
 
-    messages: {
```
parlarjb: Can you update the commit message to include a note about this breaking change? (And the `byType` change below)

Droogans: Is there something wrong with the current one?

> BREAKING CHANGE: Avoid the use of functions that evaluate many sub components into a single object. `.messages` and `.byType` with a more precise `.byText`, limiting the number of notifications that can be returned to one.

I notice I accidentally a word in the second sentence. It should say "Replace `.messages` and `.byType`..."

parlarjb: Sorry, somehow I didn't see that you'd already done it :/

src/rxNotify/rxNotify.page.js
```diff
@@ -76,31 +76,31 @@ var rxNotify = {
...
             var page = this;
-            return this.tblNotifications.each(function () {
-                notification(page.tblNotifications.get(-1)).dismiss();
+            return this.tblNotifications.map(function (notificationElement, index) {
+                return notification(notificationElement).isDismissable().then(function (dimissable) {
+                    if (dimissable) {
+                        return index;
+                    }
+                });
+            }).then(function (dismissableIndexes) {
```
jj4th: What is the motivation for dismissing them here instead of dismissing them where you return the index?

Droogans: You have to iterate over the list in reverse, otherwise you could have the `[0, 2]` indexes to dismiss, close the `[0]`th element, and run into an error when you try to later close the (non-existent) index `[2]`. Doing it backwards means you'll find each index in the correct position.

parlarjb:
LGTM

src/rxNotify/rxNotify.page.js
jj4th: I understand that, but that wasn't my question, my question was why wouldn't this work?
```
return this.tblNotifications.map(function (notificationElement, index) {
   return notification(notificationElement).isDismissable().then(function (dismissable) {
       if (dismissable) {
           notification(notificationElement).dismiss();
       }
   });
})
```

jj4th: Nevermind, I see that protractor is dumb about this, it appears to preserve the binding to repeater, so when you remember the notification, is gone from the this.tblNotifications.

PR #730, opened 2015-01-13
test(tabs): Remove `.all`, use `.names` instead

Droogans:
BREAKING CHANGE: Avoid functions that iterate over a collection of sub
components and return them as an object. It's slow and typically not
used. Prefer the use of the term `.names` (as is used everywhere else),
instead of `.allNames`.


parlarjb:
LGTM

PR #735, opened 2015-01-14
fix(rxSortableColumn): Put sort arrow in <button> and show how to sort rxStatusColumn

### LINKED ISSUE #732: rxStatusColumn tooltip not showing when no icon is present
Only the ERROR, INFO and WARNING boxes are displaying tooltips, because the tooltip is tied to the icon.

### LINKED ISSUE #733: Support rxSortableColumn with rxStatusColumn

parlarjb:
This commit puts the sort icon into the <button> that rxSortableColumn
creates, allowing us to click on the sort arrow itself.

This also updates rxStatusColumn to allow for sorting of the status
column. It adds documentation to the rxStatusColumn demo page to show
how to do this.

Finally, we also fixed a problem where tooltips would not show for
ACTIVE and PENDING status column cells. It was slightly related to
the sorting issue, so it has been bundled in to this commit.

Closes #732
Closes #733

### DIFF
--- a/src/rxStatusColumn/docs/rxStatusColumn.js
+++ b/src/rxStatusColumn/docs/rxStatusColumn.js
@@ -1,7 +1,7 @@
 /*jshint unused:false*/
 
 // This file is used to help build the 'demo' documentation page and should be updated with example code
-function rxStatusColumnCtrl ($scope, rxStatusMappings) {
+function rxStatusColumnCtrl ($scope, rxStatusMappings, rxSortUtil) {
     $scope.servers = [
         { status: 'ACTIVE', title: 'ACTIVE status' },
         { status: 'ERROR', title: 'ERROR status' },
@@ -22,4 +22,8 @@ function rxStatusColumnCtrl ($scope, rxStatusMappings) {
 
     rxStatusMappings.addAPI('fooApi', { 'DELETING': 'PENDING' });
     rxStatusMappings.mapToPending('SomeApiSpecificStatus', 'fooApi');
+    $scope.sortCol = function (predicate) {
+        return rxSortUtil.sortCol($scope, predicate);
+    };
+    $scope.sort = rxSortUtil.getDefault('status');
 }
--- a/src/rxStatusColumn/docs/rxStatusColumn.midway.js
+++ b/src/rxStatusColumn/docs/rxStatusColumn.midway.js
@@ -1,8 +1,11 @@
+var _ = require('lodash');
 var Page = require('astrolabe').Page;
 
 var rxStatusColumn = require('../rxStatusColumn.page.js').rxStatusColumn;
+var rxSortableColumn = require('../../rxSortableColumn/rxSortableColumn.page.js').rxSortableColumn;
 
 // an anonymous page object to demonstrate table and cell creation
+var repeaterString = 'server in servers';
 var tablePageObject = Page.create({
 
     rootElement: {
@@ -13,7 +16,17 @@ var tablePageObject = Page.create({
 
     tblServers: {
         get: function () {
-            return this.rootElement.all(by.repeater('server in servers'));
+            return this.rootElement.all(by.repeater(repeaterString));
+        }
+    },
+
+    column: {
+        value: function (columnName) {
+            // Only 'Status' is supported -- 'Title' is unsortable.
+            if (columnName === 'Status') {
+                var columnElement = this.rootElement.$('rx-sortable-column[sort-property="status"]');
+                return rxSortableColumn.initialize(columnElement, repeaterString);
+            }
         }
     },
 
@@ -206,6 +219,33 @@ describe('rxStatusColumn', function () {
 
             });
 
+        });
+
+    });
+
+    describe('sorting', function () {
+        var column;
+        var ascendingOrder = _.values(statuses).sort();
+
+        var statusCellData = function (cellElements) {
+            return cellElements.map(function (cellElement) {
+                return rxStatusColumn.initialize(cellElement).byType;
+            });
+        };
+
+        before(function () {
+            column = tablePageObject.column('Status');
+        });
+
+        it('should support sorting ascending', function () {
+            column.sortDescending(); // FIXME
+            expect(column.getDataUsing(statusCellData, '[rx-status-column]')).to.eventually.eql(ascendingOrder);
+        });
+
+        it('should support sorting descending', function () {
+            var descendingOrder = ascendingOrder.reverse();
+            column.sortAscending(); // FIXME
+            expect(column.getDataUsing(statusCellData, '[rx-status-column]')).to.eventually.eql(descendingOrder);
         });
 
     });
--- a/src/rxStatusColumn/docs/rxStatusColumn.midway.js
+++ b/src/rxStatusColumn/docs/rxStatusColumn.midway.js
@@ -75,11 +75,11 @@ describe('rxStatusColumn', function () {
             });
 
             it('should not have a tooltip', function () {
-                expect(status.tooltip.exists).to.eventually.be.false;
+                expect(status.tooltip.exists).to.eventually.be.true;
             });
 
-            it('should have no tooltip text', function () {
-                expect(status.tooltip.text).to.eventually.be.null;
+            it('should have tooltip text', function () {
+                expect(status.tooltip.text).to.eventually.equal('ACTIVE');
             });
 
         });
@@ -87,7 +87,7 @@ describe('rxStatusColumn', function () {
         describe('error cell', function () {
 
             before(function () {
-                status = tablePageObject.row(1).status;
+                status = tablePageObject.row(3).status;
             });
 
             it('should have a status by type', function () {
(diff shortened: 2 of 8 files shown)


src/rxStatusColumn/docs/rxStatusColumn.js
```diff
@@ -22,4 +22,8 @@ function rxStatusColumnCtrl ($scope, rxStatusMappings) {
 
     rxStatusMappings.addAPI('fooApi', { 'DELETING': 'PENDING' });
     rxStatusMappings.mapToPending('SomeApiSpecificStatus', 'fooApi');
+    $scope.sortCol = function (predicate) {
+        return rxSortUtil.sortCol($scope, predicate);
+    };
+    $scope.sort = rxSortUtil.getDefault('status');
```
Droogans: I know it's not cannon, but ticketing has severities associated with each status, and those are sorted numerically. It may allow for more control in how we organize "status importance" in the UI, instead of relying on alphabetical sorting.

I understand that this may become tricky if the statuses are pulled from the framework and placed inside of encore-svcs.

src/rxStatusColumn/docs/rxStatusColumn.midway.js
```diff
@@ -200,6 +200,10 @@ describe('rxStatusColumn', function () {
                     expect(status.api).to.eventually.equal('fooApi');
                 });
 
+                it('should have tooltip text', function () {
+                    expect(status.tooltip.text).to.eventually.equal('DELETING');
+                });
+
```
Droogans: I'll go in real fast and add some sorting tests to make sure it all works there too.

src/rxStatusColumn/docs/rxStatusColumn.js
parlarjb: I'm not quite sure what you mean. And note that the `status` being talked about here is different than the `account status` we were looking at earlier. These statuses are for things like `server.status`.

In addition, the sorting function is customizable, you can pass in whatever function you like to determine how sorting is performed.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/735/gallery.html)

src/rxStatusColumn/docs/rxStatusColumn.js
Droogans: Well, it's a bit tricky for me to explain without sounding like I have no idea what I'm talking about, but let's go for it. I'm not afraid to look dumb in public.

1. rxStatusColumn exposes abstract colors, icons and status transformations that can be arranged any way the user wants, theoretically.
2. rxStatusColumn now exposes sorting, again, in any way the user wants to implement it in their app.

At some point I wonder if there's going to be a discussion about unifying the different statuses, and making sure that sorting is the same across every app's usage of the rxStatusColumn sort feature. One way to do that is to add a numerical value for each status, and create a default sort that uses the numerical value instead.

For instance, green "Active" (or "Healthy", "OK", or whatever else it gets names) statuses would be given a numerical value of zero in the app, so that if you use it, it'll automatically put it at the top of your ascending sorts, unless you override the sort function built into the directive. At the bottom would be the red "warning" icon statuses. Between those two, of course is a really fuzzy grey area, and so that's where I'll leave it. That's why we have UX folks in the world.

parlarjb: An important thing to note is that we're not sorting by the generic status name (which are ACTIVE, INFO, WARNING, ERROR and PENDING). We're instead sorting by whatever the internal status is for the server/loadbalancer/etc.

The dev might have both "REBOOTING" and "SHUTTINGDOWN" map to our generic "PENDING" status colour, but the tooltip that a user sees, and the sort order, will use "REBOOTING" and "SHUTTINGDOWN" instead. 

This ensures that the sort is grouping the actual related things together. It will put all the REBOOTING servers together, all the SHUTTINGDOWN servers together, etc. And the dev can choose how they want those to be ordered, based on their chosen sort function.

glynnis: @Droogans +1 for this approach, as I do see that as offering a better experience. Added a note here: https://jira.rax.io/browse/EN-1501 and making an issue in the framework, as I think there's enough on @parlarjb's plate (and mine) that we should try and tackle it separately.

Droogans: Here's a snapshot of how the columns get sorted by default:

screen shot 2015-01-14 at 14 47 42

I think it should be all green statuses, all blue statuses, then yellow "pending" statuses, then yellow "warning" statuses, and finally, red statues. Perhaps some code like this?

```js
/**
 * @ngdoc service
 * @name encore.ui.rxStatusColumn:rxStatusSorts
 * @description
 * The recommended sorting function for applying rxSortableColumn to
 * rxStatusColumn. Prefers class identifiers first, followed by icons
 * as a secondary sort.
 *
 */
.factory('rxStatusSorts', function () {
  
    // Leave space between each value for any icons that offset the sort.
    var colorSortValues = {
        ACTIVE: 0,
        INFO: 4,
        PENDING: 8,
        WARNING: 12,
        ERROR: 16
    } || 20; // unknown statuses are always placed to the bottom

    // Missing and unknown icons add nothing to the sorting precendence.
    var iconSortValues = {
        'fa-circle': 1,
        'fa-exclamation-triangle': 2,
        'fa-ban': 3
    } || 0;

    var getSortValue = function (statusCellElement) {
        // There's some missing logic for extracting the important parts of the class names we care about...
        var icon = statusCellElement.find('i')
        return colorSortValues[statusCellElement.attr('class')] + iconSortValues[icon.attr('class')];
    }
  
});
```

You could tell devs to provide something like this to their rxSortableColumn sorting logic if they aren't planning on doing a lot of custom work to it.

Droogans: @parlarjb, @glynnis this should be done in a separate bit of work, absolutely. I like what's here so far. Just looking out for what's probably going to happen later.

parlarjb: I get what you're saying, but my point is that we don't want to sort on the `colorSortValues`. The colours are independent of the actual statuses from the APIs. 'REBOOTING' and 'SHUTTINGDOWN' will both likely be mapped to 'PENDING', but I don't want the rebooting and shutting down servers to be randomly intermingled when sorted. I would want all the 'REBOOTING' servers to be together, and all the 'SHUTTINGDOWN' servers together.

And this is where a custom sort function comes in, because the devs know what the order should be for their given API.

And as I put in https://github.com/rackerlabs/encore-ui/issues/737, we could make this easier on the devs by letting them pass in their desired order for their API statuses, but it still comes back to the fact that we're *not* sorting on "ACTIVE", "ERROR", "INFO", "WARNING" and "PENDING", we're sorting on the actual API statuses.

Droogans: I see. That will require users to hover over statuses to see where they are in the sort, whereas colors and symbols are always visible. Since I am just some person with an opinion, I am officially getting out of the way so @glynnis can do what she wants with this.

glynnis:
Design: LGTM :sparkles: :sparkling_heart: :sparkles:

halkon:
Dev: LGTM :1234: :v: :traffic_light: :vertical_traffic_light: 

I will point out that while @glynnis did add an update to a JIRA story in cloud, Andrew's comment is  definitely something that deserves discussion/merit (of implementation). 

I can see many cases where we would want to not just sort on `status` (for example) but be specific about which `statuses` come first, and which go last (asc -> desc or viceversa)

But it is for a later discussion

glynnis:
Just to document, I made an issue in the framework to continue the prioritization of statuses discussion Andrew kicked off (I thought I'd posted the link here, but looks like I forgot): https://github.com/rackerlabs/encore-ui/issues/737

Thanks for the suggestion, Andrew! I hope that's something we can prioritize soon. As we all seem to agree, it would definitely improve UX, particularly where sorting by status is concerned. :+1:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/735/gallery.html)

src/rxStatusColumn/docs/rxStatusColumn.midway.js
```diff
@@ -200,8 +213,39 @@ describe('rxStatusColumn', function () {
...
+        it('should support sorting ascending', function () {
+            column.sortDescending(); // FIXME
+            expect(column.getDataUsing(statusCellData, '[rx-status-column]')).to.eventually.eql(ascendingOrder);
+        });
 
+        it('should support sorting descending', function () {
+            var descendingOrder = ascendingOrder.reverse();
+            column.sortAscending(); // FIXME
+            expect(column.getDataUsing(statusCellData, '[rx-status-column]')).to.eventually.eql(descendingOrder);
```
Droogans: @parlarjb note the // FIXMEs in the test code. I have similar ones in the billing app too.

It's strange because the rxSortableColumn tests all behave normally, but I'm seeing implementations lately that sort backwards...

PR #740, opened 2015-01-15
feat(rxSortEmptyTop): DCXAPPS-1335 Add filter for placing empty values first

rogerza:
Adds the rxSortEmptyTop filter in module rxSortableColumn to allow moving empty values to the top of the list in ascending order, and the bottom of the list in descending order.

### DIFF
--- a/src/rxSortableColumn/docs/rxSortableColumn.html
+++ b/src/rxSortableColumn/docs/rxSortableColumn.html
@@ -1,6 +1,5 @@
 <div ng-controller="rxSortableColumnCtrl">
-    <p>The demo here is shown with the table also using <code>rx-floating-header</code>, but this is not required. We've only done this to illustrate
-    that <code>rxSortableColumn</code> works properly with <code>rxFloatingHeader</code></p>
+    <p>Note: The demo table is also using <code>rx-floating-header</code>, which is not required. We've only done this to illustrate that <code>rxSortableColumn</code> works properly with <code>rxFloatingHeader</code>. The table is also using <code>rxSortEmptyTop</code>.</p>
     <table rx-floating-header>
         <thead>
             <tr>
@@ -25,7 +24,7 @@
             </tr>
         </thead>
         <tbody id="talentPoolData">
-            <tr ng-repeat="resource in talentPool | orderBy:sort.predicate:sort.reverse">
+            <tr ng-repeat="resource in talentPool | rxSortEmptyTop:sort.predicate:sort.reverse">
                 <th scope="row" class="talent-name">
                     {{resource.name}}
                 </th>
--- a/src/rxSortableColumn/rxSortableColumn.js
+++ b/src/rxSortableColumn/rxSortableColumn.js
@@ -60,4 +60,33 @@ angular.module('encore.ui.rxSortableColumn', [])
     };
 
     return util;
-});
+})
+/**
+* @ngdoc filter
+* @name encore.ui.rxSortableColumn:rxSortEmptyTop
+* @description
+* Filter that moves rows with an empty predicate to the top of the column in ascending order,
+  and to the bottom in descending order.
+*
+* @example
+* <pre>
+* [{ name: { firstName: 'Adam' } }, { }] | rxSortEmptyTop 'name.firstName':false
+* Will sort as [{}, { name: { firstName: 'Adam' } }].
+
+* [{ name: { firstName: 'Adam' } }, { name: { firstName: null } ] | rxSortEmptyTop 'name.firstName':true
+* Will sort as [{ name: { firstName: 'Adam' } }, {}]
+
+* </pre>
+*/
+.filter('rxSortEmptyTop', ['$filter', '$parse', function ($filter, $parse) {
+    return function (array, key, reverse) {
+
+        var predicateGetter = $parse(key);
+
+        var sortFn = function (item) {
+            return predicateGetter(item) || '';
+        };
+
+        return $filter('orderBy')(array, sortFn, reverse);
+    };
+}]);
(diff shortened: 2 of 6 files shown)


parlarjb:
Could you please add an example usage to the docs for `rxSortableColumn`, and describe the new functionality in the README?

rogerza:
But of course.

rogerza:
image

rogerza:
image

rogerza:
Fixing the failures. These included some FIXME's in rxStatusColumn that had to do with the 'asc' and 'desc' being reversed.

rogerza:
@parlarjb anything else needed for this?

src/rxSortableColumn/rxSortableColumn.js
```diff
@@ -60,4 +60,55 @@ angular.module('encore.ui.rxSortableColumn', [])
...
+* Will sort as [{ name: { firstName: 'Adam' } }, {}]
+
+* </pre>
+*/
+.filter('rxSortEmptyTop', function () {
+    return function (array, key, reverse) {
+
+        // Checks whether the property is in the collection.
+        // It will split on . to check nested properties.
+        var hasDeepIn = function (collection, propertyPath) {
```
parlarjb: This should be defined outside of the function that's being returned. It doesn't reference `array`, `key` or `reverse`, thus you don't need a closure. Right now, you're doing the work of recreating this function every time.

src/rxSortableColumn/rxSortableColumn.js
```diff
@@ -60,4 +60,55 @@ angular.module('encore.ui.rxSortableColumn', [])
...
+            var properties = propertyPath.split('.');
+            var has = true;
+            _.forEach(properties, function (property) {
+                if (_.has(collection, property) || _.isObject(collection) && property in collection) {
+                    if (_.isEmpty(collection[property])) {
+                        has = false;
```
parlarjb: Since `has` can never go back to `true`, you should just `return false` here to shortcut the rest of the function

src/rxSortableColumn/docs/rxSortableColumn.html
```diff
@@ -25,7 +25,7 @@
             </tr>
         </thead>
         <tbody id="talentPoolData">
-            <tr ng-repeat="resource in talentPool | orderBy:sort.predicate:sort.reverse">
+            <tr ng-repeat="resource in talentPool | orderBy:sort.predicate:sort.reverse | rxSortEmptyTop:sort.predicate:sort.reverse">
```
parlarjb: I'm feeling wary of this. Doing it this way means we need to add a separate full iteration over our collection. Is there a way to add this functionality in to `rxSortUtil.sortCol`? I'd rather have the default sort function be "smarter" about handling empty/null values, than have to pass everything through a separate filter

rogerza:
I did spend some time trying that approach, but didn't get it to work quite right. A filter would also be easier to use, from a user's standpoint, I think, since they wouldn't need to do anything in the controller.

parlarjb:
You shouldn't need to do anything if you use `sortCol` either. I believe most controllers already do something like `rxSortUtil.sortCol(scope, predicate);` to generate their sort method. If it just natively knows how to deal with empty values, then you shouldn't have to change anything on the controller side.

src/rxStatusColumn/docs/rxStatusColumn.midway.js
```diff
@@ -238,13 +238,12 @@ describe('rxStatusColumn', function () {
         });
 
         it('should support sorting ascending', function () {
-            column.sortDescending(); // FIXME
             expect(column.getDataUsing(statusCellData, '[rx-status-column]')).to.eventually.eql(ascendingOrder);
         });
```
Droogans: I'd prefer it if the sort ascending call were used explicitly, instead of relying on a default sort from page load.

I'd also rather see this fix pulled out ahead of the rest of this pull request as well.

rogerza: Alright, I can do that.

rogerza:
@parlarjb I'll take another look at that, then.

rogerza:
@parlarjb I refactored so that only one filter (and one iteration) is needed. I left the bug fix from #744 in there for now. Once that is merged, I can rebase this branch.

src/rxSortableColumn/rxSortableColumn.js
```diff
@@ -60,4 +60,34 @@ angular.module('encore.ui.rxSortableColumn', [])
...
+    return function (array, key, reverse) {
+
+        var predicateGetter = $parse(key);
+
+        var sortFn = function (item) {
+            return predicateGetter(item) || '';
+        };
+
+        var sorted = $filter('orderBy')(array, sortFn);
+        return reverse ? sorted.reverse() : sorted;
```
parlarjb: In cases where `reverse` is true, this is causing an extra full iteration, because you sort in the `$filter()` line, and then run through it all again when calling `reverse.

Instead, try this: `return $filter('orderBy')(array, sortFn, reverse)`

src/rxSortableColumn/templates/rxSortableColumn.html
```diff
@@ -4,7 +4,7 @@
         <span ng-transclude></span>
         <i class="sort-icon"
             ng-style="{visibility: predicate === '{{sortProperty}}' && 'visible' || 'hidden'}"
-            ng-class="{'desc': !reverse, 'asc': reverse}">
+            ng-class="{'desc': reverse, 'asc': !reverse}">
```
parlarjb: Just to be clear, you'd like #744 to go in first, and then you'll rebase this work?

src/rxSortableColumn/rxSortableColumn.js
rogerza: Ah, good tip. That seems to work!

src/rxSortableColumn/templates/rxSortableColumn.html
rogerza: Yes, I'd prefer that.

parlarjb:
LGTM

PR #741, opened 2015-01-16
DON'T MERGE chore(npm): Publish rx-page-objects to npm

Droogans:
This needs documentation for @parlarjb but I've confirmed it works.

https://www.npmjs.com/package/rx-page-objects

These changes mean `grunt rxPageObjects` should only be used when running :shipit:.

I believe this package will still continue to be published in the CDN, but then again I'm not sure. I'll work on the documentation around this Monday.


Droogans:
Update documentation around publishing the page objects, including where to find the keys, where they go, etc.

parlarjb:
Can you put `grunt.task.requires('cloudfiles:production')` into the `rxPageObjects` task? That will ensure the task can only be run if we've successfully pushed to production, i.e. if we're running in the `shipit` task

CITguy:
LGTM

parlarjb:
LGTM, we'll put in dependencies later

PR #744, opened 2015-01-20
fix(rxColumnSort): Fix reversed ascending and descending sort

rogerza:
Fixes reversed ascending and descending sort in rxColumnSort.

### DIFF
--- a/src/rxSortableColumn/docs/rxSortableColumn.midway.js
+++ b/src/rxSortableColumn/docs/rxSortableColumn.midway.js
@@ -28,7 +28,7 @@ describe('rxSortableColumn', function () {
     });
 
     it('should return all sorts in the table', function () {
-        expect(columns.sorts).to.eventually.eql([0, -1]);
+        expect(columns.sorts).to.eventually.eql([1, -1]);
     });
     // https://github.com/rackerlabs/encore-ui/issues/694 -- End odd behavior.
 
@@ -37,8 +37,8 @@ describe('rxSortableColumn', function () {
         expect(roleColumn.rootElement.isDisplayed()).to.eventually.eq.true;
     });
 
-    it('should have a descending sort shown by default for the name column', function () {
-        expect(nameColumn.currentSortDirection).to.eventually.eq(0);
+    it('should have an ascending sort shown by default for the name column', function () {
+        expect(nameColumn.currentSortDirection).to.eventually.eq(1);
     });
 
     it('should have no sort shown by default for the job title column', function () {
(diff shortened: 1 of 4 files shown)


src/rxSortableColumn/docs/rxSortableColumn.midway.js
```diff
@@ -37,8 +37,8 @@ describe('rxSortableColumn', function () {
         expect(roleColumn.rootElement.isDisplayed()).to.eventually.eq.true;
     });
 
-    it('should have a descending sort shown by default for the name column', function () {
-        expect(nameColumn.currentSortDirection).to.eventually.eq(0);
+    it('should have an ascending sort shown by default for the name column', function () {
+        expect(nameColumn.currentSortDirection).to.eventually.eq(1);
     });
```
Droogans: This is going to conflict with https://github.com/rackerlabs/encore-ui/pull/746

Droogans:
LGTM. I'll rebase mine once @parlarjb clears [the template change](https://github.com/rackerlabs/encore-ui/pull/744/files#diff-9b7f8e58e453e652b1a72a8f3616d2e9L7).

src/rxSortableColumn/templates/rxSortableColumn.html
```diff
@@ -4,7 +4,7 @@
         <span ng-transclude></span>
         <i class="sort-icon"
             ng-style="{visibility: predicate === '{{sortProperty}}' && 'visible' || 'hidden'}"
-            ng-class="{'desc': !reverse, 'asc': reverse}">
+            ng-class="{'desc': reverse, 'asc': !reverse}">
```
parlarjb: This is going to be a breaking change. Would you mind running the Cloud midways against this, to see how many tests we're going to break? I'd at least like to give people a head's up.

rogerza: Sure thing. I'll do that now.

rogerza: Looks like 26 failures. I can work on a PR to address those.

rogerza: @parlarjb @jj4th I have a branch that fixes the Cloud errors resulting from this fix. I should wait until this is released to make a PR for it, right?

Droogans: I think it's fine, breaking changes happen. These are superficial "arrow is pointed wrong way" failures that don't actually affect end users, just tests.

I say release and fix, just like we always do.

parlarjb: Yeah, hold off on your Cloud PR until we have this in a release. Could you please update the commit message here to have the following at the bottom of the message:

```
BREAKING CHANGE: The sort directions were incorrect. This has been corrected, but it will break existing tests.
```

rogerza: Sounds good. The message is updated.

parlarjb:
LGTM

PR #747, opened 2015-01-20
Color docs

glynnis:
Add color documentation to "Basics" page of Framework's Design Style Guide.

We'll continue to build upon this as we receive more questions/see more need to explain color usage.


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/747/gallery.html)

demo/assets/generated_demo.css
```diff
@@ -0,0 +1,81 @@
...
+}
+ul.colors .status-disabled {
+  background-color: #cfd2d4;
+}
+ul.colors .action-system {
+  background-color: #db7820;
+}
+ul.colors .action-delete {
+  background-color: #fc0f1d;
+}
```
Droogans: I have a pet peeve of having no newline between definitions...my editor navigates by paragraph, and without a line break, I will jump from the top to the bottom. I like it better when I can jump in intervals of one less definition at a time, instead of just a single line.

CITguy:
I like the swatches. :+1:

glynnis:
Added line breaks and squashed. @Droogans wasn't sure what you meant by "repeater" -- is there an example of somewhere else we do this that I could refer to? Does it make sense given the varying explanations/hex codes for each color?

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/747/gallery.html)

demo/styleguide/basics.html
```diff
@@ -16,6 +16,89 @@ <h2 class="title lg clear">Basics</h2>
...
+            </li>
+            <div class="color-logic">
+                <ul>
+                    <li>This red is used to mean "error", either as part of a response or status message.</li>
+                    <li>Red is also used for link colors and icons associated with destructive/destroy actions that a user can take. Note that irreversible changes (like rebuilding a server) are not considered destructive actions.</li>
+                </ul>
+            </div>
+        </ul>
+    </ul>
+
```
Droogans: If you're repeating the same divs with the same classes, just with different values for each, you could probably pack all of it up into a javascript object and [iterate over it using ng-repeat](https://docs.angularjs.org/api/ng/directive/ngRepeat).

For the `ul.colors` section, you have many `li`'s with a `div` underneath it with colors, values and text:

```html
<ul class="colors">
  <li class="color-swatch">
      <div class="swatch action-accept"></div>
      <div class="hexcode">24b56a</div>
  </li>
  <div class="color-logic">
      <ul>
          <li>This green is used for links and icons that deal explicitly with create and add actions.</li>
          <li>Note that this green should not be used to communicate "active" or "success", the distinction being that these are states, not actions a user can take.</li>
      </ul>
  </div>
</ul>    
```

If you had to template this out, it might look like this:

```html
<ul class="colors" ng-repeat="color in colors">
  <li class="color-swatch">
      <div class="swatch {{color.className}}"></div>
      <div class="hexcode">{{color.hexValue}}</div>
  </li>
  <div class="color-logic">
      <ul ng-repeat="description in color.descriptions">
          <li>{{color.description}}</li>
      </ul>
  </div>
</ul>
```

I'd recommend throwing the color swatch data [into a controller](https://docs.angularjs.org/guide/controller), since this is just for docs.

```js
$scope.colors = [{
    className: 'action-accept',
    hexcode: '24b56a',
    description: [
        'This green is used for links and icons that deal explicitly with create and add actions.',
        'Note that this green should not be used to communicate "active" or "success", the distinction being that these are states, not actions a user can take.'
    ]
}, {
    className: 'status-success',
    hexcode: 'c8e5bf',
    description: [
        'This green is used for statuses that communicate success.'
    ]    
}];
```

This should be a separate improvement outside of the PR. It introduces a lot of new ideas into the documentation section of the codebase, such as how we should organize the code that generates this HTML, and other things that @parlarjb is probably too busy to want to spend a lot of time on. The issue of spending a lot of time on this gets worse considering there's not a lot gained by turning this section into a repeater. However, I feel that if there's some effort put into it, there may be dividends paid down the line when the time comes to add more documentation that may be easier to write and maintain as Angular code, as opposed to straight HTML.

TL; DR -- probably best to leave this as is and investigate later how to manage the documentation HTML as Angular code once there's more justification for it.

CITguy: Is there any advantage in doing this versus plain html?

Droogans: > documentation [...] may be easier to write and maintain as Angular code, as opposed to straight HTML.

That's the only real advantage I can think of. Another might be that it's a fairly simple and straightforward initiative, making it a good opportunity for those new to Angular to pick up some skills with it.

CITguy: IMO, angular is good for dynamic content, not so much for static content.  Considering that the docs shouldn't change all that often, I'm not sure applying angular functionality would gain much of an advantage.  Honestly, I think it may confuse more than it helps. `</my-2-cents>`

parlarjb:
LGTM

PR #751, opened 2015-01-26
fix(rxFormOptionTable): Change color of empty message text.

megalord:
This is part of @glynnis's work to enforce an empty message in all tables in Encore UI.  Here, the text color is simply changing from red to the default color.

screen shot 2015-01-26 at 10 04 38 am

### DIFF
--- a/src/rxForm/rxForm.page.js
+++ b/src/rxForm/rxForm.page.js
@@ -331,7 +331,7 @@ var rxOptionFormTable = {
 
     lblEmptyWarningMessage: {
         get: function () {
-            return this.rootElement.$('.msg-warn');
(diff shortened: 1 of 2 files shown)


parlarjb:
Can you please update the commit message to include something like:

```
BREAKING CHANGE: Any page objects dependent on the presence of `.msg-warn` in rxFormOptionTable will now break. Update them to use `.empty-data` instead
```

glynnis:
LGTM design-wise.

FWIW, when I've made this change in other tables, I've just taken off the `.msg-warn`, but I understand if an additional class is needed here for testing/JS purposes.

PR #757, opened 2015-01-28
fix(rxDOMHelper): Properly wrap a select node instead of its options.

### LINKED ISSUE #756: Cannot use normal option tags in rx-form-item select.
The following use of `rx-form-item`
```html
<rx-form-item label="simple option tag" description="does not work :(">
    <span class="field-select">
    	<select ng-model="property1">
        	<option value="0" selected>Option 0</option>
        </select>
    </span>
</rx-form-item>
```
generates
```html
<rx-form-item label="simple option tag" description="does not work :(" class="ng-scope ng-isolate-scope">
    <div class="form-item" ng-class="{'text-area-label': isTextArea}">
        <label class="field-label ng-binding" for="field-008">simple option tag:</label>
        <div class="field-content">
            <!-- ngIf: prefix -->
          	<span class="field-input-wrapper" ng-transclude="">
              	<span class="field-select ng-scope">
                	<select ng-model="property1" class="ng-pristine ng-valid" id="field-008">
          		  		<option value="? undefined:undefined ?"></option>
                  		<span class="field-input">
          					<option value="0" selected="">Option 0</option>
          				</span>
            		</select>
           		</span>
            </span>
            <!-- ngIf: description -->
            <div ng-if="description" class="field-description ng-binding ng-scope" ng-bind-html="description">does not work :(</div>
            <!-- end ngIf: description -->
        </div>
    </div>
</rx-form-item>
```

The problem is that the `<option>` tags are wrapped by `<span class="field-input">` rather than the `<select>` tag.  I do not see this but when using `<option ng-repeat="...">` (like the [docs](http://rackerlabs.github.io/encore-ui/#/component/rxForm)) or `<select ng-options="...">`.

megalord:
Fixes #756 
The issue is that a `select` element has a `length` property, which returns the number of options it has.  That was causing the `wrapAll` method to thing that the element was actually an array of elements.  Further, it is possible to access a specific option by its index, e.g. `selectElement[0] === optionElement`.

### DIFF
--- a/src/rxFloatingHeader/rxFloatingHeader.js
+++ b/src/rxFloatingHeader/rxFloatingHeader.js
@@ -193,8 +193,8 @@ angular.module('encore.ui.rxFloatingHeader', [])
     var wrapAll = function (newParent, elms) {
 
         // Figure out if it's one element or an array
-        var isNonEmptyArray = (elms.length && elms.tagName !== 'SELECT');
-        var el = isNonEmptyArray ? elms[0] : elms;
+        var isGroupParent = ['SELECT', 'FORM'].indexOf(elms.tagName) !== -1;
+        var el = (elms.length && !isGroupParent) ? elms[0] : elms;
 
         // cache the current parent node and sibling 
         // of the first element
@@ -208,8 +208,10 @@ angular.module('encore.ui.rxFloatingHeader', [])
         // If there are other elements, wrap them. Each time
         // it will remove the element from its current parent,
         // and also from the `elms` array
-        while (isNonEmptyArray) {
-            newParent.appendChild(elms[0]);
+        if (!isGroupParent) {
+            while (elms.length) {
+                newParent.appendChild(elms[0]);
+            }
         }
 
         // If there was a sibling to the first element,
--- a/src/rxFloatingHeader/rxFloatingHeader.js
+++ b/src/rxFloatingHeader/rxFloatingHeader.js
@@ -193,7 +193,8 @@ angular.module('encore.ui.rxFloatingHeader', [])
     var wrapAll = function (newParent, elms) {
 
         // Figure out if it's one element or an array
-        var el = elms.length ? elms[0] : elms;
+        var isNonEmptyArray = (elms.length && elms.tagName !== 'SELECT');
+        var el = isNonEmptyArray ? elms[0] : elms;
 
         // cache the current parent node and sibling 
         // of the first element
@@ -207,7 +208,7 @@ angular.module('encore.ui.rxFloatingHeader', [])
         // If there are other elements, wrap them. Each time
         // it will remove the element from its current parent,
         // and also from the `elms` array
-        while (elms.length) {
+        while (isNonEmptyArray) {
             newParent.appendChild(elms[0]);
         }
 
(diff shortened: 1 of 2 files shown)


Droogans:
I'm checking out if this fixes our problem in billing right now.

Droogans:
LGTM. Marking high priority for @parlarjb.

halkon:
Aw I had found this and my solution was definitely way less elegant than this, so I never did anything with it: https://github.com/rackerlabs/encore-ui/compare/rx-form-item-select-fix

halkon:
LGTM

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -207,7 +208,7 @@ angular.module('encore.ui.rxFloatingHeader', [])
         // If there are other elements, wrap them. Each time
         // it will remove the element from its current parent,
         // and also from the `elms` array
-        while (elms.length) {
+        while (isNonEmptyArray) {
```
halkon: Actually, just early in the morning but, wouldn't you want to re-assign this again? given the comment and the way it was before, `elms.length` always got modified when the element was moved to the new parent.  So this array would eventually stop. 

In this case, if we had 
```
elms.length = 5;
elms.tagName = 'DIV';
```
`isNonEmptyArray` stays as true forever right?

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -207,7 +208,7 @@ angular.module('encore.ui.rxFloatingHeader', [])
...
             newParent.appendChild(elms[0]);
         }
```
Droogans: Could there be an explicit skip step here, instead of hiding the skipping logic inside of `isNonEmptyArray`?

```js
if (els.tagName.indexOf(['SELECT']) == -1) {
    while (elms.length) {
        newParent.appendChild(elms[0]);
    }
}
```

megalord: Yeah, good call.  This will also fix what @halkon commented on.  Any reason you chose to use
```javascript
if (elms.tagName.indexOf(['SELECT']) == -1) {
```
instead of
```javascript
if (elms.tagName === 'SELECT') {
```
?

src/rxFloatingHeader/rxFloatingHeader.js
Droogans: The expression `(elms.length && elms.tagName !== 'SELECT');` is always evaluated, apparently. I think that seems a bit counter-intuitive, but it appears that the DOM is stateful. I agree, I typically see `isSomeCondition()` invoked as a function. Maybe it should be a function instead?

src/rxFloatingHeader/rxFloatingHeader.js
Droogans: Yes. Freddy and I went back and forth and figured `['SELECT', 'UL', 'OL', 'DL', 'TBODY', 'THEAD']` would likely get added to this list as it gets used in more places.

Not sure if now is a good time to do it. Also, if you do go with this approach, please mark clearly in the docs that this will not wrap "collection container" tags (think of a better wording though, I'm not so great at it).

megalord: Hmmm, there might be some confusion.  This change is a bugfix that *will* wrap collection container tags.  Previously, the `wrapAll` method would ignore the element and just wrap it's children.  If that is the desired outcome, then the method should be called as
```javascript
rxDOMHelper.wrapAll(selectElement, optionElements);
```

Droogans: I see. I wasn't reading close enough, and missed this crucial line:

```js
var el = (elms.length && !isGroupParent) ? elms[0] : elms;
```

I forgot that `el` is the primary focus of `wrapAll`, I was caught up in the details of `elms`.

Droogans:
LGTM.

halkon:
LGTM

PR #760, opened 2015-01-29
test(rxForm): currencyToPennies returns ints only

### LINKED ISSUE #759: rxForm.currencyToPennies has binary rounding errors.
This runs into a binary rounding error.

```js
rxForm.currencyToPennies('$1.10')
```

There's a few ways to deal with this:

```js
return parseFloat((resFloat * 100).toFixed(4));
```

This assumes there will never be more than [4 decimal places in a fraction of a dollar](http://en.wikipedia.org/wiki/Mill_%28currency%29#Finance), which is probably safe, and easy to do.

There's also the option of parsing out the precision of the decimals, and making no promises about rounding errors that can occur if you provide a currency string featuring fractions of a penny. This is also pretty safe and easy. I think I'd most likely do this one.

The last one is using a [decimal library](https://www.npmjs.com/package/decimal), which I'm not opposed to, even if it means forcing new types from third party libraries onto users. But additional dependencies should be avoided for one function use cases like this.

Droogans:
Closes #759

### DIFF
--- a/src/rxForm/docs/rxForm.midway.js
+++ b/src/rxForm/docs/rxForm.midway.js
@@ -319,8 +319,8 @@ describe('rxForm', function () {
                 expect(pennies('$0.01')).to.equal(1);
             });
 
-            it('should convert a fraction of a penny to a float', function () {
-                expect(pennies('$0.001')).to.equal(0.1);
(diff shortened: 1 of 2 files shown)


parlarjb:
LGTM

PR #763, opened 2015-01-30
fix(style): rxActionMenu width

CITguy:
* width should roughly match cog icon width

### DIFF
--- a/src/rxActionMenu/rxActionMenu.less
+++ b/src/rxActionMenu/rxActionMenu.less
@@ -3,7 +3,7 @@
 // containers for actionmenus that spawn modals
 .action-menu-container {
     position: relative;
-    width: 5px;
+    width: 15px;
     & > i {
         cursor: pointer;
         color: @subduedText;


CITguy:
### Without Fix

Table constrained by screen width: Icon overflows container :-1: 
screen shot 2015-01-30 at 5 14 37 pm

Table not constrained by screen width: No overflow :+1: 
screen shot 2015-01-30 at 5 15 06 pm


### With Fix
Constrained by screen width: No overflow :+1: 
screen shot 2015-01-30 at 5 20 20 pm

Unconstrained by screen width: No overflow :+1: 
screen shot 2015-01-30 at 5 19 58 pm

src/rxActionMenu/rxActionMenu.less
```diff
@@ -3,7 +3,7 @@
 // containers for actionmenus that spawn modals
 .action-menu-container {
     position: relative;
-    width: 5px;
+    width: 15px;
```
CITguy: This value should roughly match the width of the cog icon.

glynnis:
Design: LGTM :+1: 

Thanks for this! :cake: :ice_cream: :fish_cake: :icecream: :shaved_ice:

PR #765, opened 2015-02-02
docs(statusColumn): update demo to reflect real status usage

glynnis:
Updated statuses here to provide more real-world examples of how we use them, so new teams are more likely to understand and utilize status mappings correctly.

### DIFF
--- a/src/rxStatusColumn/rxStatusColumn.page.js
+++ b/src/rxStatusColumn/rxStatusColumn.page.js
@@ -100,12 +100,15 @@ exports.rxStatusColumn = {
 
     statuses: {
         active: 'ACTIVE',
-        build: 'BUILD',
+        deleted: 'DELETED',
         deleting: 'DELETING',
         error: 'ERROR',
-        inProgress: 'INPROGRESS',
-        reboot: 'REBOOT',
-        suspended: 'SUSPENDED'
+        migrating: 'MIGRATING',
+        rebuild: 'REBUILD',
+        rescue: 'RESCUE',
+        resize: 'RESIZE',
+        suspended: 'SUSPENDED',
+        unknown: 'UNKNOWN'
     },
 
     icons: {
--- a/src/rxStatusColumn/README.md
+++ b/src/rxStatusColumn/README.md
@@ -62,7 +62,7 @@ To accommodate different statuses, the `rxStatusMappings` factory includes metho
     rxStatusMappings.addGlobal({
         'RUNNING': 'ACTIVE',
         'STANDBY': 'INFO',
-        'REBOOTING': 'WARNING',
+        'SUSPENDED': 'WARNING',
         'FAILURE': 'ERROR'
     })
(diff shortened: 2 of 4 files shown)


src/rxStatusColumn/rxStatusColumn.page.js
```diff
@@ -100,12 +100,15 @@ exports.rxStatusColumn = {
...
-        reboot: 'REBOOT',
-        suspended: 'SUSPENDED'
+        deleted: 'DELETED',
+        unknown: 'UNKNOWN',
+        rescue: 'RESCUE',
+        suspended: 'SUSPENDED',
+        rebuild: 'REBUILD',
+        resize: 'RESIZE',
+        migrating: 'MIGRATING',
+        deleting: 'DELETING'
```
Droogans: `<that-guy>`

Can you sort these alphabetically?

`</that-guy>`

Droogans:
@parlarjb normally, these status changes would break the constants spread throughout many tests via `rxStatusColumn.statuses`, but in this case it's not a breaking change because no one has had a chance to really use it.

src/rxStatusColumn/rxStatusColumn.page.js
eric-dunham: It looks like they originally were? :P :+1:

parlarjb:
@Droogans I thought that `statuses` was only there to support the midway tests for the framework, not for other products?

The reason is that there are possibly an infinite number of statuses. The framework only knows about five, and products must create explicit mappings between their own statuses and the five inside the framework

Droogans:
I created those to mirror what the directive supports in the user interface. Since there's going to be defaults, the tests should at least recognize them. If folks add their own, it'll be on them to name them and support them themselves.

At this point, I'm just rubber-ducking my decision here...does it make sense to push these default mappings in rx-page-objects?

src/rxStatusColumn/rxStatusColumn.page.js
glynnis: Done!

parlarjb:
The directive doesn't support those in the interface. The only statuses the directive knows about are active/warning/error/info/pending. 

A product will have its own large set of statuses, and the goal of this directive is to map each of those statuses to one of the five above.

So you don't want to put these in rx-page-objects, because they may have no relation whatsoever to the statuses available in a particular product

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/765/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/765/gallery.html)

parlarjb:
LGTM

PR #775, opened 2015-02-04
feat(rxPaginate): Persist items per page accross applications.

### LINKED ISSUE #742: rxPaginate should persist the user-selected "items per page"`
Right now when a user selects a value for "items per page", it is global to the app it was selected it, but it is not persisted.

This should become global to *all* apps, and be persisted with LocalStorage. By persisting with LocalStorage, it will automatically become global to all apps.

megalord:
Fixes #742

### DIFF
--- a/src/rxPaginate/README.md
+++ b/src/rxPaginate/README.md
@@ -2,6 +2,10 @@
 
 Used for adding pagination around a data object.
 
+# Persistence
+
+The user's preference for the number of items to display per page will be persisted across applications using localstorage. This preference is set whenever the use selects a new number to show.
+
 # Hiding the pagination
 
 In some instances, the pagination should be hidden if there isn't enough data to require it. For example, if you have `itemsPerPage` set to 10, but only have 7 items of data (so only one page). Hiding the pagination is pretty simple:
--- a/src/rxPaginate/README.md
+++ b/src/rxPaginate/README.md
@@ -4,7 +4,7 @@ Used for adding pagination around a data object.
 
 # Persistence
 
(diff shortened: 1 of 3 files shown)


src/rxPaginate/rxPaginate.js
```diff
@@ -111,8 +109,10 @@ angular.module('encore.ui.rxPaginate', [])
             itemSizeList.splice(index, 0, itemsPerPage);
         }
 
+        var selectedItemsPerPage = parseInt(LocalStorage.getItem('rxItemsPerPage'));
+
         // If the user has chosen a desired itemsPerPage, make sure we're respecting that
-        if (!_.isUndefined(selectedItemsPerPage) && _.contains(itemSizeList, selectedItemsPerPage)) {
+        if (!_.isNaN(selectedItemsPerPage) && _.contains(itemSizeList, selectedItemsPerPage)) {
```
parlarjb: There's actually a pre-existing bug here, maybe you can fix it. If `opts.itemsPerPage` exists (i.e. the dev has requested a specific number to show up on this given page), that value will be ignored.

Would you mind putting a fix for that into this PR?

megalord: most certainly

src/rxPaginate/README.md
```diff
@@ -2,6 +2,10 @@
 
 Used for adding pagination around a data object.
 
+# Persistence
+
+The user's preference for the number of items to display per page will be persisted across applications using localstorage. This preference is set whenever the use selects a new number to show.
```
parlarjb: You're missing an "r" here in "user": `whenever the use selects`

parlarjb:
LGTM

PR #776, opened 2015-02-04
feat(table): Add styles and demo for nested tables.

### LINKED ISSUE #761: Add styles/directive for parent-child tables
There are some existing styles for this present in Cloud and the Dedicated Load Balancer tool. There's also a separate (non-preferred) implementation in vCenters here: https://staging.encore.rackspace.com/virt/vcenters

We should consider bringing these styles and/or making a directive for this in the Framework, since it's a pattern being utilized across several projects. Note that we should also add an additional layout to the demo that features this pattern.

Existing implementation in Cloud: https://github.com/rackerlabs/encore-cloud-ui/pull/445
screen shot 2015-01-29 at 3 35 27 pm

megalord:
Fixes #761 

screen shot 2015-02-04 at 10 26 57 am


glynnis:
Changed title to reflect that this shouldn't be merged until Framework freeze to support Billing project has ended.

glynnis:
Spoke with @megalord just now about this, and the only thing that needs adding here for an LGTM is a demo/documentation update to include an example of using non-table data inside a child (like metadata styles), then we'll be good to go!

parlarjb:
LGTM

megalord:
Demo using metadata instead of a table for the nested content.  The odd thing about this design is the amount of whitespace in the subrow, but it is possible that there is not any more information to display.  Also, the location seems a bit arbitrary, though I did not care much for the centered version either.  What do you think?

screen shot 2015-02-18 at 12 38 40 pm

glynnis:
Design: this LGTM. I'm aware of design work needed for non-table displays in child rows (as @megalord pointed out). Making an issue for that since it requires some thinking. Currently only the LBS tool uses non-table displays of data in child rows, so I think it's important to have it documented even if it needs some work.

Mergemergemerge!

glynnis:
New issue here: https://github.com/rackerlabs/encore-ui/issues/804

PR #777, opened 2015-02-04
chore(rxMisc): centralize rxDOMHelper

### LINKED ISSUE #734: Centralize rxDOMHelper
`rxDOMHelper` is currently defined within `rxFloatingHeader` but its functionality is used across directives.  Please move `rxDOMHelper` to its own file in a centralized location for injection by any directive.  I'd suggest using a namespace such as `encore.ui.util`.

CITguy:
* initialize `encore.ui.rxMisc` as a namespace for shared logic
* move `rxDOMHelper` to `rxMisc`
* Fixes #734


CITguy:
Title changed to reflect "billing-only" moratorium.

parlarjb:
How does this currently look on the demo page? Is it showing up on the left side with the other components?

Ideally, we'll at least have a README file with some documentation, and will show up with the other components, otherwise people coming to the framework won't be aware that this exists

parlarjb:
I just pulled your branch and checked. Because there's no README, this isn't showing up in the components list on the demo page

CITguy:
Ah, yes. That got pushed out of my short term memory. I'll hit that first thing tomorrow.

CITguy:
Added documentation. Waiting on Travis.

parlarjb:
LGTM

halkon:
My only question is, whether `rxMisc` describes this best or not.  Given that this deal with UI elements I agree this staying in `encore-ui`, but if we're going to create an `rxMisc` module, just seems too open to everything, would `rxHelpers` be better? (In lieu of my original thought of `rxUtil` when I first saw it, but I'm sure that may have been already thought about).

halkon:
Aside from the name discussion which, really it's all about the module name more than anything:

LGTM - and thank you @CITguy for doing this, I can sleep a little better at night now :grin: 
LGTM

Issue #782, opened 2015-02-09
rxBillingSearch button needs height adjustment 


velveteer:
Because of the added options to the `rxBillingSearch` template, the entire div is now vertically expanded, which means that the magnifying glass image is going to expand to fill it (`height: 100%`): https://github.com/rackerlabs/encore-ui/blob/master/src/rxApp/rxApp.less#L398

The issue as it shows in the framework currently: 
![issue](https://cloud.githubusercontent.com/assets/3134804/6110188/a3cf5448-b045-11e4-99a3-6c56e0b58758.png)

This was how I fixed it in Billing-UI, but we need a better solution in the framework:
```
fieldset > .search-action { 
    height: 40% !important;
}
```

@glynnis Any ideas?

CITguy:
The `<button>` within the `<fieldset>` is set to `position: absolute;` which is attaching on the first ancestor up the tree with a `position: relative;`.  In this case, the `<div class="rx-app-search">` is where the button is attaching its positions, not the `<fieldset>` (where it should be attaching). This can be corrected by adding `position: relative;` for the `<fieldset>`.  I can get a PR in shortly.

CITguy:
@velveteer, was the 1.6.3 release satisfactory in resolving this issue?

velveteer:
@CITguy Yep.

PR #783, opened 2015-02-09
fix(css): rx-app-search icon

CITguy:
* In regards to #782
* Fix `position: absolute;` attachment

### DIFF
--- a/src/rxApp/rxApp.less
+++ b/src/rxApp/rxApp.less
@@ -360,6 +360,10 @@
         font-size: 14px;
         color: #babab9;
 
+        fieldset {
+          position: relative;
+        }
+
         // NOTE: We have to have this duplication because CSS doesn't like combining the placholder selectors
         .placeholder() {
            font-style: italic;


Droogans:
@velveteer has his setup all ready for pulling in changes in encore-ui to reflect his local build, and seeing how this is his problem specifically, should probably be merged by him.

velveteer:
LGTM. This fixes the issue locally. My only concern is are we ok with changing `fieldset` behavior across the entire framework?

parlarjb:
For my own edification, this fix is taking advantage of this? http://css-tricks.com/absolute-positioning-inside-relative-positioning/

CITguy:
@parlarjb: yes, exactly

CITguy:
@velveteer: this change only affects fieldsets within the `.rx-app-search` scope, so it's not globally modifying `fieldset` behavior.

velveteer:
@CITguy Ah I see that now after viewing the full file. Then this definitely LGTM.

CITguy:
@parlarjb, @Droogans, who should merge this?

Droogans:
I nominate @glynnis...she's a neutral third party (who, unlike me, actually understands the full ramifications of this kind of a change).

parlarjb:
Let's merge this in as soon as the framework freeze is over

CITguy:
This is billing-specific. Will that affect when it's merged?

glynnis:
LGTM :+1: Interestingly enough, no other search boxes in the nav are using `<fieldset>` around `<input>`s within their `<form>`s. Should that someday be the case, though, @CITguy is correct -- this is correctly scoped and shouldn't affect other fields around Encore.

:dancer: :dancer: :dancer:

PR #784, opened 2015-02-09
docs(currencyToPennies): Remove stale example

Droogans:


src/rxForm/rxForm.page.js
```diff
@@ -91,7 +91,7 @@ exports.rxForm = {
             resFloat = -resFloat;
         }
 
-        // 0.001 -> 0.1
+        // 0.001 -> 1
```
CITguy: If I run `0.001` through the `parseInt(...)` function, I get `0` not `1`.

Droogans: Serves me right by relying on my memory. I'm just going to delete both lines. The main documentation covers it fine.

Thanks.

parlarjb:
LGTM

Issue #792, opened 2015-02-13
Examine design issue of multiple greys and disabled grey color matching encore background


glynnis:
The case of many greys! Right now there are a lot of them, and in the case of our color for "disabled" grey, there is not enough contrast for it to be used successfully on the context of the grey background we use for the body of all content pages.

Investigate, document, and propose possible solutions.

kobowley:
Will solving this issue help with a solution for 758

glynnis:
@kobowley I see #758 as related, since it does use gray as a color, but I think this issue isn't a blocker for #758. Rather, this issue is a really high-level design question (how we use colors in Encore), whereas #758 is a request for a specific new directive.

CITguy:
Closing. Captured in JIRA https://jira.rax.io/browse/FRMW-87

PR #800, opened 2015-02-17
Bulk select support for Cloud

parlarjb:
This implements the necessary changes to support https://github.com/rackerlabs/encore-cloud-ui/pull/465

Namely:

 1. Modifies `rxFloatingHeader` to support the appearing/disappearing "message header row" 
 2. Brings in `progressbar` from Angular Bootstrap.
 3. Adds the ability do disable "press Escape to close" for `rxModalAction` 
 4. Has `rxStatusColumn` explicitly watch for changes in status, and update accordingly
 5. Modifies some padding values for `rxModalAction` and `rxForm`

### DIFF
--- a/src/progressbar/progressbar.less
+++ b/src/progressbar/progressbar.less
@@ -1,0 +1,188 @@
+@import 'vars';
+//
+// Progress bars
+// --------------------------------------------------
+
+.progress-bar-variant(@color) {
+  background-color: @color;
+
+  // Deprecated parent class requirement as of v3.2.0
+  .progress-striped & {
+    #gradient > .striped();
+  }
+}
+
+//** Background color of the whole progress component
+@progress-bg:                 #f5f5f5;
+//** Progress bar text color
+@progress-bar-color:          #fff;
+//** Variable for setting rounded corners on progress bar.
+@progress-border-radius:      4px;
+
+//** Default progress bar color
+@progress-bar-bg:             @siteBrandingBg;
+//** Success progress bar color
+@progress-bar-success-bg:     #5cb85c;
+//** Warning progress bar color
+@progress-bar-warning-bg:     @infoOrange;
+//** Danger progress bar color
+@progress-bar-danger-bg:      @warnRed;
+//** Info progress bar color
+@progress-bar-info-bg:        @infoBlue;
+@font-size-base: @appFontSize;
+@font-size-large:         ceil((@font-size-base * 1.25));
+@font-size-small:         ceil((@font-size-base * 0.85)); 
+//** Unit-less `line-height` for use in components like buttons.
+@line-height-base:        1.428571429; // 20/14
+//** Computed "line-height" (`font-size` * `line-height`) for use with `margin`, `padding`, etc.
+@line-height-computed:    floor((@font-size-base * @line-height-base)); // ~20px
+
+// Bar animations
+// -------------------------
+
+// WebKit
+@-webkit-keyframes progress-bar-stripes {
+  from  { background-position: 40px 0; }
+  to    { background-position: 0 0; }
+}
+
+// Spec and IE10+
+@keyframes progress-bar-stripes {
+  from  { background-position: 40px 0; }
+  to    { background-position: 0 0; }
+}
+
+
+// Bar itself
+// -------------------------
+
+// Outer container
+.progress {
+  overflow: hidden;
+  height: @line-height-computed;
+  margin-bottom: @line-height-computed;
+  background-color: @progress-bg;
+  border-radius: @progress-border-radius;
+  .box-shadow(inset 0 1px 2px rgba(0,0,0,.1));
+}
+
+// Bar of progress
+.progress-bar {
+  float: left;
+  width: 0%;
+  height: 100%;
+  font-size: @font-size-small;
+  line-height: @line-height-computed;
+  color: @progress-bar-color;
+  text-align: center;
+  background-color: @progress-bar-bg;
+  .box-shadow(inset 0 -1px 0 rgba(0,0,0,.15));
+  .transition(width .6s ease);
+}
+
+// Striped bars
+//
+// `.progress-striped .progress-bar` is deprecated as of v3.2.0 in favor of the
+// `.progress-bar-striped` class, which you just add to an existing
+// `.progress-bar`.
+.progress-striped .progress-bar,
+.progress-bar-striped {
+  #gradient > .striped();
+  background-size: 40px 40px;
+}
+
+// Call animation for the active one
+//
+// `.progress.active .progress-bar` is deprecated as of v3.2.0 in favor of the
+// `.progress-bar.active` approach.
+.progress.active .progress-bar,
+.progress-bar.active {
+  .animation(progress-bar-stripes 2s linear infinite);
+}
+
+
+// Variations
+// -------------------------
+
+.progress-bar-success {
+  .progress-bar-variant(@progress-bar-success-bg);
+}
+
+.progress-bar-info {
+  .progress-bar-variant(@progress-bar-info-bg);
+}
+
+.progress-bar-warning {
+  .progress-bar-variant(@progress-bar-warning-bg);
+}
+
+.progress-bar-danger {
+  .progress-bar-variant(@progress-bar-danger-bg);
+}
+
+
+
+#gradient {
+
+  // Horizontal gradient, from left to right
+  //
+  // Creates two color stops, start and end, by specifying a color and position for each color stop.
+  // Color stops are not available in IE9 and below.
+  .horizontal(@start-color: #555; @end-color: #333; @start-percent: 0%; @end-percent: 100%) {
+    background-image: -webkit-linear-gradient(left, @start-color @start-percent, @end-color @end-percent); // Safari 5.1-6, Chrome 10+
+    background-image: -o-linear-gradient(left, @start-color @start-percent, @end-color @end-percent); // Opera 12
+    background-image: linear-gradient(to right, @start-color @start-percent, @end-color @end-percent); // Standard, IE10, Firefox 16+, Opera 12.10+, Safari 7+, Chrome 26+
+    background-repeat: repeat-x;
+    filter: e(%("progid:DXImageTransform.Microsoft.gradient(startColorstr='%d', endColorstr='%d', GradientType=1)",argb(@start-color),argb(@end-color))); // IE9 and down
+  }
+
+  // Vertical gradient, from top to bottom
+  //
+  // Creates two color stops, start and end, by specifying a color and position for each color stop.
+  // Color stops are not available in IE9 and below.
+  .vertical(@start-color: #555; @end-color: #333; @start-percent: 0%; @end-percent: 100%) {
+    background-image: -webkit-linear-gradient(top, @start-color @start-percent, @end-color @end-percent);  // Safari 5.1-6, Chrome 10+
+    background-image: -o-linear-gradient(top, @start-color @start-percent, @end-color @end-percent);  // Opera 12
+    background-image: linear-gradient(to bottom, @start-color @start-percent, @end-color @end-percent); // Standard, IE10, Firefox 16+, Opera 12.10+, Safari 7+, Chrome 26+
+    background-repeat: repeat-x;
+    filter: e(%("progid:DXImageTransform.Microsoft.gradient(startColorstr='%d', endColorstr='%d', GradientType=0)",argb(@start-color),argb(@end-color))); // IE9 and down
+  }
+
+  .directional(@start-color: #555; @end-color: #333; @deg: 45deg) {
+    background-repeat: repeat-x;
+    background-image: -webkit-linear-gradient(@deg, @start-color, @end-color); // Safari 5.1-6, Chrome 10+
+    background-image: -o-linear-gradient(@deg, @start-color, @end-color); // Opera 12
+    background-image: linear-gradient(@deg, @start-color, @end-color); // Standard, IE10, Firefox 16+, Opera 12.10+, Safari 7+, Chrome 26+
+  }
+  .horizontal-three-colors(@start-color: #00b3ee; @mid-color: #7a43b6; @color-stop: 50%; @end-color: #c3325f) {
+    background-image: -webkit-linear-gradient(left, @start-color, @mid-color @color-stop, @end-color);
+    background-image: -o-linear-gradient(left, @start-color, @mid-color @color-stop, @end-color);
+    background-image: linear-gradient(to right, @start-color, @mid-color @color-stop, @end-color);
+    background-repeat: no-repeat;
+    filter: e(%("progid:DXImageTransform.Microsoft.gradient(startColorstr='%d', endColorstr='%d', GradientType=1)",argb(@start-color),argb(@end-color))); // IE9 and down, gets no color-stop at all for proper fallback
+  }
+  .vertical-three-colors(@start-color: #00b3ee; @mid-color: #7a43b6; @color-stop: 50%; @end-color: #c3325f) {
+    background-image: -webkit-linear-gradient(@start-color, @mid-color @color-stop, @end-color);
+    background-image: -o-linear-gradient(@start-color, @mid-color @color-stop, @end-color);
+    background-image: linear-gradient(@start-color, @mid-color @color-stop, @end-color);
+    background-repeat: no-repeat;
+    filter: e(%("progid:DXImageTransform.Microsoft.gradient(startColorstr='%d', endColorstr='%d', GradientType=0)",argb(@start-color),argb(@end-color))); // IE9 and down, gets no color-stop at all for proper fallback
+  }
+  .radial(@inner-color: #555; @outer-color: #333) {
+    background-image: -webkit-radial-gradient(circle, @inner-color, @outer-color);
+    background-image: radial-gradient(circle, @inner-color, @outer-color);
+    background-repeat: no-repeat;
+  }
+  .striped(@color: rgba(255,255,255,.15); @angle: 45deg) {
+    background-image: -webkit-linear-gradient(@angle, @color 25%, transparent 25%, transparent 50%, @color 50%, @color 75%, transparent 75%, transparent);
+    background-image: -o-linear-gradient(@angle, @color 25%, transparent 25%, transparent 50%, @color 50%, @color 75%, transparent 75%, transparent);
+    background-image: linear-gradient(@angle, @color 25%, transparent 25%, transparent 50%, @color 50%, @color 75%, transparent 75%, transparent);
+  }
+}
+
+
+.animation(@animation) {
+  -webkit-animation: @animation;
+       -o-animation: @animation;
+          animation: @animation;
+}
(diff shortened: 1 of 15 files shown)


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -5,13 +5,32 @@
...
+            };
+        },
         link: function (scope, table) {
+            var state, seenFirstScroll, trs, ths, clones, inputs, maxHeight, header, singleThs, maxThWidth;
 
-            var state = 'fixed',
+            var setup = function () {
+
+                if (clones && clones.length) {
+                    _.each(clones, function (clone) {
```
halkon: Given the comment about the possible memory leak, would this be better:

```
_.forEachRight(clones, function (clone) {
    clone.remove();
    clone = undefined;
});
``` or somehow remove the instance from the list?

Comment (parlarjb) [inline src/rxFloatingHeader/rxFloatingHeader.js]: I could also clear the list afterwards. But my concern isn't with this list, so much as internal parts of Angular keeping references. My concern comes from here: http://stackoverflow.com/a/27016855

Comment (halkon) [inline src/rxFloatingHeader/rxFloatingHeader.js]: gotcha, given that `clones` are only html elements themselves, I think this is fine.  A bit creepy and scary but it's most likely fine.

:+1:

Erasing the list afterwards is most likely a good idea also

File: src/rxFloatingHeader/rxFloatingHeader.js [hunk 2]
```diff
@@ -27,24 +52,87 @@ angular.module('encore.ui.rxFloatingHeader', [])
...
+                    // This <tr> only had one <th> in it. Grab that <th> and its clone
+                    // Also grab the width of the <th>, and compare it to our max width.
+                    // We need to do this because if a <th> was hidden, and then made to
+                    // appear while floating, its width will be too short, and will need
+                    // to be updated
+                    if (thsInTr.length === 1) {
+                        var th = thsInTr[0];
+                        var width = rxDOMHelper.width(th);
+                        if (width !== 'auto') {
+                            var numeric = _.parseInt(width.replace('px', ''));
```
halkon: do you really have to do `width.replace('px', '')`?

Native `parseInt`:
```
> parseInt('123px')
123
```

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -27,24 +52,87 @@ angular.module('encore.ui.rxFloatingHeader', [])
...
+                            }
+                        }
+
+                        singleThs.push([th, angular.element(clone.find('th'))]);
+                    }
+                });
+
+                // Explicitly set the width on every <th> that is the *only*
+                // <th> in its <tr>
+                maxThWidth = maxThWidth.toString() + 'px';
```
halkon: is the explicit `.toString()` here really necessary? it already converts to string when concatenating a string.

```
> parseInt('123px') + 'px'
'123px'
```

That being said... I may just be looking at this too closely for anything :grin:

halkon:
Even with my comments - LGTM
LGTM

src/rxStatusColumn/rxStatusColumn.js
```diff
@@ -22,15 +22,33 @@ angular.module('encore.ui.rxStatusColumn', [])
...
-            scope.statusIcon = rxStatusColumnIcons[scope.mappedStatus] || 'fa-exclamation-circle';
-            element.addClass('status');
-            element.addClass('status-' + scope.mappedStatus);
-            element.addClass('rx-status-column');
+
+            var lastStatusClass = '';
+
+            var setTooltip = function (tooltip, status) {
+                scope.tooltipText = tooltip || status || '';
+            };
```
CITguy: Multiple argument arity to this function is unnecessary.  Without looking at the internal logic for this function, a developer has no idea which argument trumps the other.  I'd redefine this to accept a single argument, like so:

```javascript
var setTooltip = function (tooltip) {
  scope.tooltipText = tooltip || '';
};
```
so that I can explicitly define which value to pass, via the business logic. ...like so:

```javascript
var tooltip = thisValue || thatValue || theOtherValue;
setTooltip(tooltip);

var tooltip2 = thisValue || thatValue;
setTooltip(tooltip);
```
This keeps the arity consistent across all calls to the function and removes any doubt around argument precedence.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: At first I was all "ewww", but apparently stripping non-numeric portions is part of the spec. :+1:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: I refuse to allow implicit conversions when adding ints and strings. REFUSE.

halkon: REFUSED! :anger: :angel: :innocent: :camel: :smiling_imp:

halkon: That's fine, like I said, I was just grasping at straws to find anything :stuck_out_tongue:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -27,24 +52,87 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
                 inputs = [],
                 maxHeight,
                 header = angular.element(table.find('thead'));
+                
+                // Are we currently floating?
+
```
rogerza: is this newline needed?

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -27,24 +52,87 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
...
+                    // into the DOM whenever the original <tr> elements need to float. This keeps the
+                    // height of the table correct, and prevents it from jumping up when we put
+                    // the <tr> elements into a floating state.
+                    // It also makes sure the column widths stay correct, as the widths of the columns
+                    // are determined by the current fixed header, not the floating header.
+                    var clone = tr.clone();
+                    clones.push(clone);
+
+                    if (floating) {
+                        // We're currently floading, so add the class back, and
```
rogerza: "floating"

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -80,7 +168,7 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
                         var topOffset = 0;
                         _.each(trs, function (tr, index) {
                             tr.addClass('rx-floating-header');
-                            tr.css({ 'top': topOffset });
+                            tr.css({ 'top': topOffset + 'px' });
                             topOffset += trHeights[index];
```
rogerza: and I quote: "I refuse to allow implicit conversions when adding ints and strings. REFUSE." :)

halkon: :grin: sadly this is still `int` to `int`

```
...
                            trHeights.push(parseInt(rxDOMHelper.height(tr)));
                         });
 
                         // Grab the current widths of each `th` that we want to float

                         var topOffset = 0;
...
                             topOffset += trHeights[index];
```

rogerza:
DEV-LGTM

sundip:
DEV: LGTM :+1:

src/rxFloatingHeader/rxFloatingHeader.js
halkon: But the line above... YES (nice catch)

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -80,7 +168,7 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
                         var topOffset = 0;
                         _.each(trs, function (tr, index) {
                             tr.addClass('rx-floating-header');
-                            tr.css({ 'top': topOffset });
+                            tr.css({ 'top': topOffset + 'px' });
```
halkon: What @rogerza said in: https://github.com/rackerlabs/encore-ui/pull/800#discussion_r25268297 

:grin: :grin: :grin: :grin:

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: noooooooooooooo _universe collapses in on itself_

src/progressbar/progressbar.less
```diff
@@ -0,0 +1,201 @@
...
+//** Default progress bar color
+@progress-bar-bg:             @brand-primary;
+//** Success progress bar color
+@progress-bar-success-bg:     @brand-success;
+//** Warning progress bar color
+@progress-bar-warning-bg:     @brand-warning;
+//** Danger progress bar color
+@progress-bar-danger-bg:      @brand-danger;
+//** Info progress bar color
+@progress-bar-info-bg:        @brand-info;
```
halkon: doing a quick search I see we're not really using any of the `@brand-` LESS variables aside from inside of here. 

Any chance we can map these to: https://github.com/rackerlabs/encore-ui/blob/master/src/styles/vars.less#L55-L68?

As well as get rid of the `@brand-` variables.

src/progressbar/progressbar.less
```diff
@@ -0,0 +1,201 @@
...
+@font-size-small:         ceil((@font-size-base * 0.85)); 
+//** Unit-less `line-height` for use in components like buttons.
+@line-height-base:        1.428571429; // 20/14
+//** Computed "line-height" (`font-size` * `line-height`) for use with `margin`, `padding`, etc.
+@line-height-computed:    floor((@font-size-base * @line-height-base)); // ~20px
+
+//** By default, this inherits from the `<body>`.
+@headings-font-family:    inherit;
+@headings-font-weight:    500;
+@headings-line-height:    1.1;
```
halkon: These `@headings-` are not used at all

src/progressbar/progressbar.less
```diff
@@ -0,0 +1,201 @@
...
+}
+
+@brand-primary: @siteBrandingBg;
+@brand-success:         #5cb85c;
+@brand-info:            #5bc0de;
+@brand-warning:         #f0ad4e;
+@brand-danger:          #d9534f;
+@border-radius-base:        4px;
+@border-radius-large:       6px;
+@border-radius-small:       3px;
```
halkon: `@border-radius-` are not used anywhere else

src/progressbar/progressbar.less
```diff
@@ -0,0 +1,201 @@
...
+//** Background color of the whole progress component
+@progress-bg:                 #f5f5f5;
+//** Progress bar text color
+@progress-bar-color:          #fff;
+//** Variable for setting rounded corners on progress bar.
+@progress-border-radius:      @border-radius-base;
```
halkon: Replace with explicit value as the variable is only used in this place

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -8,10 +8,29 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
 .directive('rxFloatingHeader', function (rxDOMHelper) {
     return {
         restrict: 'A',
-        scope: {},
+        controller: function ($scope) {
+            this.update = function () {
+                // It's possible for a child directive to try to call this
+                // before the rxFloatingHeader link function has been run,
+                // meaning $scope.update won't have been configured yet
+                if (_.isFunction($scope.update)) {
```
halkon: I think you and I have talked about this before but:

https://docs.angularjs.org/api/ng/service/$compile#-bindtocontroller-

However I do think this requires a higher version of angular than what we are currently using (also have you tried any of this in `1.3`?)

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -27,24 +52,86 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
...
-                    inputs.push(input);
-                }
-            });
+                    // We are going to clone all the <tr> elements in the <thead>, and insert them
+                    // into the DOM whenever the original <tr> elements need to float. This keeps the
+                    // height of the table correct, and prevents it from jumping up when we put
+                    // the <tr> elements into a floating state.
+                    // It also makes sure the column widths stay correct, as the widths of the columns
+                    // are determined by the current fixed header, not the floating header.
+                    var clone = tr.clone();
```
halkon: What's the behavior that happens when you clone the row with the `rx-floating-header` class on it (and removing it after cloning?)

src/rxFloatingHeader/rxFloatingHeader.js
```diff
@@ -124,6 +211,9 @@ angular.module('encore.ui.rxFloatingHeader', ['encore.ui.rxMisc'])
                 scope.updateHeaders();
             });
 
+            scope.update = function () {
+                setup();
```
halkon: I'm always wishy washy about these, so I'll ask it in a different way, is `setup()` a variable that is bound to be changed?

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: It's not available in 1.2.x

halkon: I knooooooOoooowwww

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: Good catch. It seems to work fine to just do 
```
var clone = tr.clone();
clones.push(clone);
if (floating) {
    clone.removeClass('rx-floating-header');
}
```

src/rxFloatingHeader/rxFloatingHeader.js
parlarjb: Nope. Whatcha getting at?

src/rxFloatingHeader/rxFloatingHeader.js
halkon: :+1:

src/rxFloatingHeader/rxFloatingHeader.js
halkon: More of an opinion thing than actually an issue or a "big deal", I would rather prefer to just do: 

```
scope.update = setup;
```
since there's really nothing else being done with `setup()` itself.  

Ehhh, I think I'm just more in code clean up mode and these stand out a lot more than they really matter.

image

parlarjb: The reason I tend to avoid that is because JavaScript. Namely, by doing it the way I have it, I don't worry about `setup` being bound to an object, possibly affecting its `this`. Not that I use `this` in it, but I want to be very clear about what the call environment is

halkon: Good point, gladly these days that's a lot easier to work with using `_.bind` or `Function.bind`, but that's not the issue here.   

Either way explicit is good, that's why I was half and half on it.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/progressbar/progressbar.less
```diff
@@ -0,0 +1,188 @@
...
+@progress-bg:                 #f5f5f5;
+//** Progress bar text color
+@progress-bar-color:          #fff;
+//** Variable for setting rounded corners on progress bar.
+@progress-border-radius:      4px;
+
+//** Default progress bar color
+@progress-bar-bg:             @siteBrandingBg;
+//** Success progress bar color
+@progress-bar-success-bg:     #5cb85c;
```
CITguy: Where did this color come from? I'm not seeing it in `src/styles/vars.less`.

parlarjb: Collapsing variables. Bootstrap originally gave us `@brand-success:         #5cb85c;`, but since it was only being used in one place, @halkon asked me to just use it directly.

If there's a colour from `vars.less` that we should use, I'd be happy to, but for now, this is default bootstrap.

CITguy: I'd suggest `@actionGreen`, but I think that's meant for links. I'm fine with `#5cb85c` for now, but this needs to be referenced for #792.

CITguy:
LGTM - But I think we need to have a separate PR for new components (i.e. `progressbar`).

parlarjb:
Yeah, I'd normally bring in new components separately, but this is all to facilitate the bulk select in Cloud

halkon:
OMG

src/progressbar/progressbar.less
glynnis: FWIW, I'm aware the progress bar needs some design love to make it more Encore-y, w/r/t color as well as the border radius stuff that was mentioned. We can iterate on the specific styles for it moving forward -- more concerned about getting this baby merrrrrggeeedd :steam_locomotive:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

src/rxModalAction/rxModalAction.less
```diff
@@ -38,7 +38,7 @@
     left: 50%;
     z-index: 1050;
     width: 560px;
-    padding-right: @modalPadding;
+    //padding-right: @modalPadding;
```
halkon: OOOOOOOOOOOH COMMENTED CODE!!!

src/rxModalAction/rxModalAction.js
```diff
@@ -152,6 +155,10 @@ angular.module('encore.ui.rxModalAction', ['ui.bootstrap'])
                 // The eval will execute function (if it exists)
                 scope.$eval(attrs.preHook);
 
+                if (_.has(attrs, 'disableEsc')) {
+                    attrs.keyboard = false;
```
halkon: :+1:

halkon:
Sigh....


....

....


LGTM
LGTM

src/rxModalAction/rxModalAction.less
parlarjb: DAMN

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/800/gallery.html)

PR #808, opened 2015-02-20
fix(rxAccountUsers): Switch to $route from $routeParams

parlarjb:
$routeParams is not ready until the current route has successfully
resolved. This was causing a race condition if the nav directive
tried to use it before the route finished. $route.current.params
will always be ready


julianbonilla:
LGTM :+1:

halkon:
LGTM :+1:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/808/gallery.html)

PR #816, opened 2015-02-25
fix(style): last cell in subtable too small

CITguy:


parlarjb:
Would you mind putting up some before/after screenshots?

CITguy:
*Before*
screen shot 2015-02-25 at 9 48 01 am

*After*
screen shot 2015-02-25 at 9 48 07 am

Styling of subtable cell width should be handled by the application, not the framework.

glynnis:
Removing the width may break action cog cells for data tables -- I think that's why this width was in here? Worth investigating to ensure it doesn't affect that.

Agreed that if this is handled framework-side that it should be linked to a class that can be applied to tables that have action cogs (rather than all tables), but it'll take some effort to get that rolled out properly to all the various projects.

parlarjb:
This is limited to `.expanded-container`, so shouldn't we be ok w.r.t. normal data tables?

CITguy:
The use case I'm trying to apply this pattern to doesn't have action cogs in the last column. It's purely a two column subtable for additional data.  

Without this CSS modification, I would require adding a single empty column so my data displays properly.  Is this the intended pattern to follow?

_no empty column at end_
screen shot 2015-02-25 at 10 29 13 am

_extra, empty column added_
screen shot 2015-02-25 at 10 30 12 am

_Note: screenshots are of an in-progress prototype._

eric-dunham:
image
This is an action subtable inside an expandable table with the style removed. Please ignore the fact that the cogs don't exist. I just copy-pasted in the inspector and angular wouldn't render the `<rx-action-menu>` again.

If this warning from the style guide is followed, I don't think there will be problems:

> HEADS UP: In Encore Cloud's servers.less, we added the the actions class to the table header and styled the column at 15 pixels. This has not been implemented for EncoreUI.

As it stands, even the demo gives this example markup for an action table:
```html
<thead>
    <tr>
        <th>Name</th>
        <th>Occupation</th>
        <th class="actions" style="width: 15px;"></th>
    </tr>
</thead>
```

:-/

If I take out the style attribute, it looks like:
image

That leads me to believe that `.actions` for table headers hasn't been implemented at the framework level yet.

src/rxApp/common.less
```diff
@@ -105,6 +105,9 @@ table {
         background: @tableHeaderBg;
         color: @tableHeaderText;
         border: 1px solid @tableHeaderBorder;
+        &.actions {
+          width: 15px;
+        }
```
CITguy: _Note:_ This only applies to `table thead th.actions`

CITguy:
Addressing what @eric-dunham mentioned, I've added the `.actions` style for use with column headers on an action cog column. I've also updated the documentation and style guides to reflect this.

src/rxApp/common.less
```diff
@@ -153,7 +156,6 @@ table {
 
         th:last-child {
             border-right-width: 0;
-            width: 15px;
```
CITguy: _Note:_ this only applies to `.expanded-container table.subtable th:last-child`. So there shouldn't be any adverse effects to existing action columns.

CITguy:
@glynnis, for peace of mind, here's a table with action-menu cogs with the style change in place.

_no change to last column size_
screen shot 2015-02-25 at 1 47 53 pm
screen shot 2015-02-25 at 1 49 13 pm

glynnis:
@CITguy @eric-dunham Excellent! Thanks for the follow-up and screen shots.

LGTM.

Issue #814, opened 2015-02-24
Support for multi-screen modals


glynnis:
There are several use cases recently for multi-screen modals, i.e. modals that have several screens they progress through before a submit/close action is done. Think "wizard".

For an example of the kind of behavior I'm talking about: http://simpulton.github.io/angularjs-wizard/

There are three use cases for this in Encore currently:
- Batch actions in Cloud already does this (somewhat inelegantly according to @parlarjb :wink: )
- LBS has a similar implementation to Cloud's batch actions
- A use case has come up for making a payment/commenting on a payment in EOD

parlarjb:
The demo comes from this blog post: http://onehungrymind.com/build-angularjs-wizard-redux/

Issue #799, opened 2015-02-17
rxModalAction hide submission buttons


megalord:
In [this PR](https://github.com/rackerlabs/encore-cloud-ui/pull/465) to Encore Cloud, there is a new UX pattern in which the footer buttons are hidden once the form is submitted.  This prevents a user from exiting the modal while the process is ongoing.  Once the process has completed, a blue return button is displayed, allowing the user to return to the application.

This pattern is useful any time an operation does not complete when the initial request returns, and its status must be polled instead.

Example:
![blue return button](https://cloud.githubusercontent.com/assets/9376/6113847/1dc6e92e-b068-11e4-826f-aaedef20e8ad.png)

glynnis:
+1 getting this in the Framework

megalord:
I'll give it a go.

PR #817, opened 2015-02-25
feat(rxModalAction): Make footer content stateful.

megalord:
This PR addresses issue #799, and I suppose #814 to some degree.  It is intended for modals that kick of tracked processes to hide the submit and cancel buttons after the process is kicked off.  It could possibly be used/adapted for implementing a wizard in the modal.

This solution is backwards compatible with existing modals, meaning existing (and some future) uses of  `rxModalForm` do not have to worry about manipulating the modal's state.

screen shot 2015-02-25 at 9 48 42 am
screen shot 2015-02-25 at 9 48 49 am
screen shot 2015-02-25 at 9 49 07 am
screen shot 2015-02-25 at 9 48 56 am

### DIFF
--- a/src/rxModalAction/README.md
+++ b/src/rxModalAction/README.md
@@ -28,24 +28,19 @@ The `<rx-modal-form>` directive is helpful for providing a common format to form
 
 - Title
 - Subtitle
-- isLoading
- - whether the modal form should have a 'loading' message by default. This is usually tied in with a `pre-hook` to load data
-- submitText
- - Override of the 'submit' button text
-- cancelText
- - Override of the 'cancel' button text
-- returnText
- - Override of the 'return' button text
+- isLoading - whether the modal form should have a 'loading' message by default. This is usually tied in with a `pre-hook` to load data
+- submitText - Override of the 'submit' button text
+- cancelText - Override of the 'cancel' button text
+- returnText - Override of the 'return' button text
 
- This directive also provides an 'autofocus' mechanism, which will move the keyboard focus cursor to the first 'tabbable' input available in the form.
+This directive also provides an 'autofocus' mechanism, which will move the keyboard focus cursor to the first 'tabbable' input available in the form.
 
 ## rxModalFooter
 
 When a modal has multiple views or kicks off a process that should be tracked within the modal, the `<rx-modal-footer>` directive should be used.  Its attributes are:
 
 - state
-- global
- - This is just a flag and takes no value.
+- global - This is just a flag and takes no value.
 
 Modal Footers should be defined in the same template as the Modal Form unless the footer is global, in which case it should be loaded in `module.run()`.  Global footers can be used in any subsequent modal by changing to the state they were defined with.
 
--- a/src/rxModalAction/README.md
+++ b/src/rxModalAction/README.md
@@ -34,5 +34,20 @@ The `<rx-modal-form>` directive is helpful for providing a common format to form
  - Override of the 'submit' button text
 - cancelText
  - Override of the 'cancel' button text
+- returnText
+ - Override of the 'return' button text
 
  This directive also provides an 'autofocus' mechanism, which will move the keyboard focus cursor to the first 'tabbable' input available in the form.
+
+## rxModalFooter
+
+When a modal has multiple views or kicks off a process that should be tracked within the modal, the `<rx-modal-footer>` directive should be used.  Its attributes are:
+
+- state
+- global
+ - This is just a flag and takes no value.
+
+Modal Footers should be defined in the same template as the Modal Form unless the footer is global, in which case it should be loaded in `module.run()`.  Global footers can be used in any subsequent modal by changing to the state they were defined with.
+
+The modal's controller also inherits the `setState()` method on the scope, which should be used to toggle different views or footers. See the Multi-View Example of this design pattern's usage.
+The default `editing` state shows the standard submit and cancel buttons, and the only other state provided by the framework is `complete` (showing the return button).
--- a/src/rxModalAction/docs/rxModalAction.html
+++ b/src/rxModalAction/docs/rxModalAction.html
@@ -15,16 +15,17 @@
         <rx-modal-action
             classes="button"
             controller="rxModalStateCtrl"
-            template-url="multiview.html">
+            template-url="multiview.html"
+            disable-esc>
             Multi-View Example
         </rx-modal-action>
     </p>
 
     <script type="text/ng-template" id="changePassword.html">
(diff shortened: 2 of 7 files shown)


glynnis:
Marking this as urgent, since it'd be great to have for https://jira.rax.io/browse/EOD-273, which is in the current sprint.

src/rxModalAction/rxModalAction.js
```diff
@@ -28,10 +31,16 @@ angular.module('encore.ui.rxModalAction', ['ui.bootstrap'])
             isLoading: '=',
             submitText: '@',
             cancelText: '@',
+            returnText: '@',
```
parlarjb: This (and `submitText` and `cancelText` too, to be honest) should be `?@`, to show that they're optional.

And you'll need to add a `@param` above for this

src/rxModalAction/docs/rxModalAction.html
```diff
@@ -31,4 +40,29 @@
...
+            submit-text="Send Request"
+            return-text="Return to rxModalAction docs">
+            <rx-notifications stack="modal"></rx-notifications>
+            <div ng-switch="state">
+                <div ng-switch-when="editing">
+                    <p>Press "Send Request" to start the fake process.</p>
+                </div>
+                <div ng-switch-when="confirm">
+                    <h3 class="title">Are you sure you want to continue?</h3>
+                    <p ng-click="confirm()">This action cannot be undone.</p>
```
parlarjb: Is this `confirm()` supposed to be here?

src/rxModalAction/rxModalAction.js
```diff
@@ -28,10 +31,16 @@ angular.module('encore.ui.rxModalAction', ['ui.bootstrap'])
...
             defaultFocus: '@'
         },
+        controller: function ($scope, $element) {
+            _.assign($scope.$parent, _.pick($scope, ['submitText', 'cancelText', 'returnText']));
+            $compile(rxModalFooterTemplates.flush())($scope.$parent, function (clone) {
+                $element.children('div.modal-footer').append(clone);
```
parlarjb: Would you mind adding some comments here? And why are you putting it into a `controller`? Could this not go in the `link`?

src/rxModalAction/rxModalAction.js
```diff
@@ -43,30 +52,24 @@ angular.module('encore.ui.rxModalAction', ['ui.bootstrap'])
...
-                    $timeout(function () {
-                        // first check for an element with autofocus
-                        focusElement = formSelector.querySelector('[autofocus]');
-                        if (!focusElement) {
-                            focusElement = formSelector.querySelector(focusSelectors[focus]);
-                        }
-                        if (focusElement) {
-                            focusElement.focus();
-                        }
-                    }, 400);
```
parlarjb: Nice, I've been wanting to make a similar change for awhile

src/rxModalAction/rxModalAction.js
```diff
@@ -85,6 +88,62 @@ angular.module('encore.ui.rxModalAction', ['ui.bootstrap'])
     // cancel out of the modal if the route is changed
     $rootScope.$on('$routeChangeSuccess', $modalInstance.dismiss);
 })
+.factory('rxModalFooterTemplates', function () {
```
parlarjb: Could you add an `@ngdoc`to help explain what this is for?

src/rxModalAction/README.md
```diff
@@ -34,5 +34,20 @@ The `<rx-modal-form>` directive is helpful for providing a common format to form
...
\ No newline at end of file
+ This directive also provides an 'autofocus' mechanism, which will move the keyboard focus cursor to the first 'tabbable' input available in the form.
+
+## rxModalFooter
+
+When a modal has multiple views or kicks off a process that should be tracked within the modal, the `<rx-modal-footer>` directive should be used.  Its attributes are:
+
+- state
+- global
+ - This is just a flag and takes no value.
```
parlarjb: FYI, something in our styles for the demo app is causing it to not render this as a sub-item to `global`

screen shot 2015-02-25 at 1 30 15 pm

src/rxModalAction/docs/rxModalAction.js
```diff
@@ -9,4 +9,35 @@ function rxModalActionCtrl ($scope) {
...
+}
+
+function rxModalStateCtrl ($scope, $modalInstance, $timeout, rxNotify) {
+    function complete () {
+        $scope.loaded = true;
+        $scope.setState('complete');
+        rxNotify.add('Operation Success!', {
+            stack: 'modal',
+            type: 'success'
+        });
```
parlarjb: In the demo app, open the modal, click "Send Request", click "Yes", and then when the operation completes, press "Escape" on your keyboard instead of clicking "Return to rxModalAction docs".

If you do this, and then open the modal again, the "Operation Success" message will be in the first state of the modal,

screen shot 2015-02-25 at 1 33 18 pm

You can disable the Escape key using the `disable-esc` attribute I recently introduced, or figure out how to register for cleanup when a modal closes, and do your `rxNotify.clear('modal')` there. I want to make sure that our example is pristine.

src/rxModalAction/docs/rxModalAction.html
megalord: Ah good catch... leftover from debugging.

src/rxModalAction/rxModalAction.js
megalord: Sure thing!

src/rxModalAction/rxModalAction.js
megalord: Oh, I didn't know about the `?@` syntax. Cool!

parlarjb: And it might be that it's `@?`, i.e. question mark at the end, I can never remember

src/rxModalAction/rxModalAction.js
megalord: Ah sorry, this was from a previous iteration.  I see no reason why this cannot go in `link`, and that is more appropriate.

parlarjb: FYI, I've never seen a reason to use `controller` in a directive, *except* for when you're using it to define an API that other directives can use

src/rxModalAction/docs/rxModalAction.js
megalord: Good find.  I think using `disable-esc` is a better solution than using `$modalInstance.result.catch()`.

parlarjb:
LGTM

PR #827, opened 2015-03-01
E2e rx character count

Droogans:
@glynnis I was really happy to see that you wrote out end to end tests for your new component. Most devs don't do that :cry:.

However, there were some missing parts. I've gone ahead and addressed them one at a time, leaving a detailed commit messages overviewing each change so that you know what an ideal page object looks like.

I'll squash all these together once you get the gist of what's going on here, and you're good with merging it.

### DIFF
--- a/src/rxCharacterCount/rxCharacterCount.exercise.js
+++ b/src/rxCharacterCount/rxCharacterCount.exercise.js
@@ -3,12 +3,20 @@ var rxCharacterCount = require('./rxCharacterCount.page.js').rxCharacterCount;
 
 /**
    rxCharacterCount exercises.
+   @exports encore.exercise.rxCharacterCount
    @param {Object} [options=] - Test options. Used to build valid tests.
    @param {string} [options.cssSelector=] - Fallback selector string to initialize widget with.
+   @param {Number} [options.maxCharacters=254] - The total number of characters allowed.
+   @param {Number} [options.nearLimit=10] - The number of remaining characters needed to trigger the "near-limit" class.
+   @param {Boolean} [options.ignoreInsignificantWhitespace=false] - Whether or not the textbox ignores leading and
+   trailing whitespace when calculating the remaining character count.
    @example
    ```js
    describe('default exercises', encore.exercise.rxCharacterCount({
-       cssSelector: '.secondary-info rx-paginate', // select one of many widgets on page
+       cssSelector: '.demo-custom-max-characters', // select one of many widgets on page
+       maxCharacters: 25,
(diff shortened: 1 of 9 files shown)


src/rxCharacterCount/rxCharacterCount.exercise.js
```diff
@@ -26,11 +38,93 @@ exports.rxCharacterCount = function (options) {
...
+
+        it('should not set the over-limit class on an empty text box', function () {
+            expect(component.isOverLimit()).to.eventually.be.false;
+        });
+
+        var belowNearLimitLength = options.maxCharacters + 1 - options.lowerBoundary;
+        it('should set the near-limit class when ' + belowNearLimitLength + ' characters are entered', function () {
+            component.comment = Array(belowNearLimitLength).join('a');
+            expect(component.isNearLimit()).to.eventually.be.false;
+        });
```
Droogans: It should *not* set the near-limit class.

src/rxCharacterCount/rxCharacterCount.page.js
```diff
@@ -35,27 +51,40 @@ var rxCharacterCount = {
...
 
+    /**
+       @returns {Number} The remaining number of characters that can be entered.
+     */
     remaining: {
         get: function () {
-            return this.lblRemaining.getText();
+            return this.lblRemaining.getText().then(function (remainingCharacters) {
+                return parseInt(remainingCharacters, 10);
+            });
```
Droogans:
```js
return this.lblRemaining.getText().then(parseInt);
```

src/rxCharacterCount/rxCharacterCount.exercise.js
```diff
@@ -17,7 +25,11 @@ exports.rxCharacterCount = function (options) {
         options = {};
     }
 
-    options = _.defaults(options, { /* defaults go here */ });
+    options = _.defaults(options, {
+        maxCharacters: 254,
+        lowerBoundary: 10,
+        ignoreInsignificantWhitespace: true
+    });
```
Droogans: `lowerBoundary` isn't very intuitive. Pick something like `nearLimitCharacters` or something instead.

parlarjb:
Since you're already bundling some exercise changes here, could you also make sure to add the exercise file to this? https://github.com/rackerlabs/encore-ui/blob/master/grunt-tasks/component-template/rename.json Right now, the exercise file doesn't get named properly when running `grunt-init`

Droogans:
@parlarjb got it.

PR #829, opened 2015-03-03
fix(rxPageObjects): Exercise written with incompatible require path.

megalord:
The [regex rule for concatenation](https://github.com/rackerlabs/encore-ui/blob/master/grunt-tasks/options/concat.js#L46) does not support having the `.js` file extension.  This breaks v1.9.0 of the npm module.
```
$ node
> require('rx-page-objects')
Error: Cannot find module './rxCharacterCount.page.js'
```

### DIFF
--- a/grunt-tasks/options/concat.js
+++ b/grunt-tasks/options/concat.js
@@ -43,7 +43,7 @@ module.exports = {
             banner: removeFromExercises.join(''),
             process: function (src) {
                 // replace all exercise require statements from src/ directory to index.js (published version)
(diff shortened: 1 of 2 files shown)


Droogans:
I'll add another commit to catch `.js` endings in the regex.

Droogans:
@megalord I can't push to your repo please cherry-pick https://github.com/Droogans/encore-ui/commit/5d75fb365013b84d8e37cdf4b45a53c0c8425032 into this PR.

parlarjb:
LGTM

PR #833, opened 2015-03-03
fix(rxButton): Disabled style takes priority over cancel.

### LINKED ISSUE #771: Allow disabled styling for Cancel Button
The LBS project disables the cancel button during api operations. The is a request, please, to allow `disabled` styling to be applied to the cancel button.

megalord:
Fixes #771 

Now, applying the `disabled` attribute results in:
screen shot 2015-03-03 at 5 49 19 pm


glynnis:
@megalord Can you explain a little more about the need/use cases for this? While you have applied the "disabled" styles correctly and this is indeed what a cancel button would look like disabled, I'm a little concerned about when/why we would disable a cancel button in the first place, since usually the "Cancel" button should be a user's "way out" of an action. Could you explain?

megalord:
Ah, this was being used in a modal before the new footers were available.  Back then, we had wanted to disable the buttons while an operation was running to prevent the modal from being closed.  Frankly, I didn't give this much thought and was just crossing off an item from the issue list.  I agree though, we can probably just close this PR and the related issue - it seems they are no longer relevant.

PR #834, opened 2015-03-04
feat(rxSortUtil): No longer use 'name' as a default predicate.

### LINKED ISSUE #786: Remove "name" default sort on rxSortUtil.getDefault
https://github.com/rackerlabs/encore-ui/blob/7c3c4c58f13abbcb9df472dfabbdea3e39ab8c6f/src/rxSortableColumn/rxSortableColumn.js#L46

I feel that when I write

```js
$scope.sort = rxSortUtil.getDefault();
```

I am trying to say "sort no columns". In this case, it will sort a "Name" column for me. Instead, I have to write

```js
$scope.sort = rxSortUtil.getDefault('no matching key in repeater');
```

Or something that will result in `undefined`.

megalord:
Fixes #786

### DIFF
--- a/src/rxSortableColumn/docs/rxSortableColumn.js
+++ b/src/rxSortableColumn/docs/rxSortableColumn.js
@@ -2,7 +2,7 @@
 
 // This file is used to help build the 'demo' documentation page and should be updated with example code
 function rxSortableColumnCtrl ($scope, PageTracking, rxSortUtil) {
-    $scope.sort = rxSortUtil.getDefault();
+    $scope.sort = rxSortUtil.getDefault('name');
     $scope.pager = PageTracking.createInstance();
 
     $scope.sortCol = function (predicate) {
--- a/src/rxSortableColumn/rxSortableColumn.js
+++ b/src/rxSortableColumn/rxSortableColumn.js
@@ -43,8 +43,7 @@ angular.module('encore.ui.rxSortableColumn', [])
     var util = {};
 
     util.getDefault = function (property, reversed) {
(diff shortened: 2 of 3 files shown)


parlarjb:
I don't know how I feel about this one, or #786 in general. My opinion is that the two arguments to `getDefault()` are required args, and anything else is an error. I can be convinced otherwise though.

cc @Droogans

Droogans:
I feel that anyone expecting an unspecified default to match their column's sorting are relying on an undocumented, unintuitive feature. If you're going to sort on a property, you should specify that property.

parlarjb:
This goes back to your issue #786 though. You mention "sort no columns", but I don't know that that's a correct request

megalord:
I do find it odd to sort by a specific property as a default.  If the arguments are both "required", then it boils down to how we treat that error.  I would expect not passing them to result in the behavior of this change rather than doing a default sort on the `name` property.  That is regardless of whether sorting no columns is valid as an intentional use.

CITguy:
I'm with @megalord about having an arbitrary property being the default for sorting.  IMO, `rxSortableColum` should provide the _ability to sort_, while the controller specifies _how to sort_.

parlarjb:
Yeah, default property of `name` is weird, and I'd guess a relic of this starting in Cloud and then moving to the framework.

We don't have much precedent for raising errors in the framework, so the behaviour in this PR is as good as we'll get for now.

megalord:
> behaviour
:smile:

parlarjb:
LGTM

PR #835, opened 2015-03-04
feat(rxCharacterCount): Highlight characters over the limit.

### LINKED ISSUE #820: Add character highlighting when over limit for character count
This is something we didn't implement right off the bat for character count, but we should look into it for a future iteration, since it will provide more immediate feedback to the user.

We should implement this much like Twitter does:
screen shot 2015-02-26 at 1 28 27 pm

megalord:
Fixes #820 
There are some concerns about the accessibility of using a `contenteditable` element instead of an input, so instead a background is provided for the input.  A drawback of this approach is that the width of the background must be the same as the input.  Since we already direct developers to apply an overriding style to change the width (see comments), this seemed like an acceptable constraint.
screen shot 2015-03-04 at 9 16 58 am

### DIFF
--- a/src/rxCharacterCount/rxCharacterCount.less
+++ b/src/rxCharacterCount/rxCharacterCount.less
@@ -28,6 +28,7 @@
     position: relative;
     background-color: #ffffff;
     input, textarea {
+      position: relative; // Ensure the input is layered over the highlighting without z-index
       background-color: transparent;
       resize: none;
       .inputBorderPadding;
--- a/src/rxCharacterCount/rxCharacterCount.less
+++ b/src/rxCharacterCount/rxCharacterCount.less
@@ -7,7 +7,8 @@
 // custom wrappers with custom widths can be set.
 .character-count-wrapper {
     width: 375px;
-    textarea {
+    // These two selectors must always be styled with the same width.
+    .input-highlighting, textarea {
         width: 370px;
     }
 }
@@ -20,5 +21,27 @@
     }
     &.over-limit {
         color: #fc0f1d;
+    }
+}
+
+.counted-input-wrapper {
+    position: relative;
+    background-color: #ffffff;
+    input, textarea {
+      background-color: transparent;
+      resize: none;
+      .inputBorderPadding;
+    }
+    .input-highlighting {
+        position: absolute;
+        top: 4px;
+        left: 6px;
+        white-space: pre-wrap;
+        word-wrap: break-word;
+        color: transparent;
+        .over-limit-text {
+            background-color: @warnRed;
+            opacity: 0.3;
+        }
     }
 }
--- a/src/rxCharacterCount/README.md
+++ b/src/rxCharacterCount/README.md
@@ -11,3 +11,5 @@ The 254 and 10 values are both configurable. To change the maximum number of cha
 ### Leading and Trailing characters ###
 By default, any text field using `ng-model` has `ng-trim="true"` applied to it. This means that any leading and trailing spaces/blanks in your text field will be ignored. They will not count towards the remaining character count. If you want it to count leading/trailing spaces, then just add `ng-trim="false"` to your `<textarea>`.
 
+### Styling ###
+When specifying a width other than the default, you should style some built-in classes in addition to the text field itself. As in the demo, the `.input-highlighting` class should have the same width as the text field, and the `.character-count-wrapper` should be used to correctly position the counter.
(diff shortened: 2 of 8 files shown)


parlarjb:
An odd interaction bug with this: If you highlight any of the text using your mouse cursor, and hit "delete" (on a Mac), it'll send you back a page, instead of deleting the text. If you highlight the text with shift+cursor keys, then it works correctly.

megalord:
Hmmm interesting, I'll take a look.

megalord:
The highlighting was layered on top of the textarea, so selecting with the mouse was actually taking focus off of the input.

src/rxCharacterCount/rxCharacterCount.less
```diff
@@ -7,7 +7,8 @@
 // custom wrappers with custom widths can be set.
 .character-count-wrapper {
     width: 375px;
-    textarea {
+    // These two selectors must always be styled with the same width.
+    .input-highlighting, textarea {
```
parlarjb: Could you add a note to the README about this? (i.e. if someone wants to change the width they have to remember to include `.input-highlighting`)

glynnis:
With the README change @parlarjb requested, this has a design: LGTM.

Thanks Jordan! This is a nice win for UX. :clap: :clap:

src/rxCharacterCount/rxCharacterCount.less
megalord: Yeah, good point.  This is probably the biggest wart of the solution.

CITguy: Could you just apply a single css class to both the `.input-highlighting` and the `textarea`, say a `.highlighting-box` class to simplify?

CITguy:
With the increasing complexity of the HTML surrounding the counted element, I feel like this directive would be better defined as an element (`<rx-character-count>{element}</rx-character-count>`) rather than an attribute (`<{element} rx-character-count>`).

parlarjb:
@CITguy The stuff you crossed out in your comment seems to say the same thing as your "Edit" text. Is there a typo, or were you just rewording?

CITguy:
@parlarjb I initially thought that we could define `rx-character-count` as an element directive, but I realized we need access to `ng-model` for functionality. It wouldn't make sense to define `ng-model="something"` in both the element and a `rx-character-count` element.  So a better way to solve the complex markup would be to introduce `ngTransclude` functionality into the directive, add an HTML template, and remove as much markup from the javascript as possible.

I hope this clarifies things.  I don't know if this suggestion would fall within the scope of this PR, though.

parlarjb:
Ahh, ok. I thought your `ngTransclude` comment implied you still wanted an element directive, and as you mentioned, that doesn't really work with `ngModel`.

I'd like to merge this in as-is, and we could look at using `ngTransclude` in the future.

src/rxCharacterCount/rxCharacterCount.less
megalord: This is already possible if the dev gives their `textarea` or `input` the `.input-highlighting` class.  I feel this is already pretty transparent as far as what is going on and what is expected.

megalord:
FWIW, I like @CITguy 's suggestion for using a template instead of manipulating the DOM in the directive; there's quite a bit going on there.  However, I don't believe `ngTransclude` is a solution for this specific problem since this is an attribute directive with no nested content.  In the end, there may only be a slight win in terms of complexity.

CITguy:
@megalord you can use `transclude: 'element'` to do the transclusion for a self-closing element like `<input />`

parlarjb:
Is that still supported? [According to this](http://ng.malsup.com/#!/transclude-property), you also need `replace: true`, and I think that `replace: true` has been deprecated

CITguy:
`replace` will be removed with angular 2.0 (per [this documentation](https://docs.angularjs.org/api/ng/service/$compile)), but at that point, it'd be a rewrite anyway.

CITguy:
and yes, you would need `replace: true` as well

parlarjb:
@CITguy are you ok with merging as is, and we can look at the transclusion option later?

CITguy:
yeah, the transclusion can be a topic for later. I'm ok with merge :+1: LGTM

CITguy:
@parlarjb, I've created an issue to capture the transclusion topic.

parlarjb:
LGTM

scadams:
The behavior seems a little off for me on an input box:

image

Was this intentional?

scadams:
Disregard, Jay pointed me to the documentation that discusses how to handle this.

PR #836, opened 2015-03-04
feat(rxForm): Disable options

rogerza:
Add ability to disable options in rx-form-table-options

### DIFF
--- a/src/rxForm/README.md
+++ b/src/rxForm/README.md
@@ -32,7 +32,7 @@ Creates a field row wrapped in a fieldset. Used for fields which include multipl
 
 ## rxFormOptionTable
 
-Given an data object and an additional optional object for column labels, rxFormOptionTable creates a series of radio or checkbox buttons. 
+Given a data object and an additional optional object for column labels, rxFormOptionTable creates a series of radio or checkbox buttons.
 
 ### Attributes
 
@@ -42,4 +42,11 @@ Given an data object and an additional optional object for column labels, rxForm
 - `data`: A data object to prefill the radio/checkbox collection. For checkboxes, checked values default to true unless `value` and `falseValue` attributes are given. See the samples.
 - `model`: The AngularJS model to tie all radios/checkboxes together.
 - `field-id`: The ID of the elements.
-- `empty-message`: (string) A default message if the data attribute is empty. 
+- `empty-message`: (string) A default message if the data attribute is empty.
+- `disable-fn`: (Function) Optional callback function to determine if option should be disabled. Parameters `tableId`, `fieldId`, and `rowId` will be passed to the function.
+
+# Services
+
+## rxFormUtils
+
+Provides a set of utility functions for accessing form data
--- a/src/rxForm/docs/rxForm.js
+++ b/src/rxForm/docs/rxForm.js
@@ -47,6 +47,7 @@ function rxFormDemoCtrl ($scope) {
     ];
 
     $scope.volume = {
+        data: 0,
         isNameRequired: true,
         type: _.first($scope.types).value, // select the first type by default
         checked: [true, 'unchecked'] //example with first checkbox automatically checked
@@ -56,22 +57,32 @@ function rxFormDemoCtrl ($scope) {
 
     $scope.optionTableData = [
         {
+            'id': 'option1_id',
             'name': 'Option #1',
             'value': 0,
             'obj': {
                 'name': 'Nested Name 1'
             }
         }, {
+            'id': 'option2_id',
             'name': 'Option #2',
             'value': 1,
             'obj': {
                 'name': 'Nested Name 2'
             }
         }, {
+            'id': 'option3_id',
             'name': 'Option #3',
             'value': 2,
             'obj': {
                 'name': 'Nested Name 3'
+            }
+        }, {
+            'id': 'option4_id',
+            'name': 'Option #4',
+            'value': 3,
+            'obj': {
+                'name': 'Nested Name 4'
             }
         }
     ];
@@ -107,6 +118,11 @@ function rxFormDemoCtrl ($scope) {
     $scope.compressedLayout = { value: false };
 
     $scope.details = { email: '' };
+
+    $scope.disableOption = function (tableId, fieldId, rowId) {
+        return rowId === 'option4_id';
+    };
+
 }
 
 // A dummy directive only used within the rxForm demo page.
@@ -132,5 +148,5 @@ angular.module('encore.ui.rxForm')
             });
         }
     };
-    
+
 });
--- a/src/rxForm/rxForm.js
+++ b/src/rxForm/rxForm.js
@@ -140,6 +140,12 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
  * @param {Object} model - Value to bind input to using ng-model
  * @param {String} fieldId - Used for label and input 'id' attribute
  * @param {Object} required - Value passed to input's 'ng-required' attribute
+ * @param {Function} disableFn - Callback function to determine if option should be disabled.
+                                 Takes tableId, fieldId, and rowId as parameters.
+                                 Example:
+```
+ disable-fn="disableOption(tableId, fieldId, rowId)"
+```
  */
 .directive('rxFormOptionTable', function ($interpolate) {
     return {
@@ -153,15 +159,20 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
             model: '=',
             fieldId: '@',
             required: '=',
-            emptyMessage: '@'
+            emptyMessage: '@',
+            disableFn: '&?'
         },
-        controller: function ($scope) {
+        controller: function ($scope, $element) {
             var determineMatch = function (val1, val2) {
                 if (_.isUndefined(val1) || _.isUndefined(val2)) {
                     return false;
                 }
 
                 return (val1 == val2);
+            };
+
+            $scope.checkDisabled = function (row) {
+                return $scope.disableFn({ tableId: $element.attr('id'), fieldId: $scope.fieldId, rowId: row.id });
             };
 
             // Determines whether the row is the initial choice
@@ -171,7 +182,7 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
 
             // Determines whether the row is selected
             $scope.isSelected = function (val, idx) {
-                // row can only be 'selected' if it's not the 'current'' value
+                // row can only be 'selected' if it's not the 'current' value
                 if (!$scope.isCurrent(val)) {
                     if ($scope.type == 'radio') {
                         return (val == $scope.model);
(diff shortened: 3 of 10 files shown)


src/rxForm/docs/rxForm.js
```diff
@@ -107,6 +117,15 @@ function rxFormDemoCtrl ($scope) {
     $scope.compressedLayout = { value: false };
 
     $scope.details = { email: '' };
+
+    $scope.disableOption = function (tableId, fieldId, rowId) {
+        if (rowId === 'option4_id') {
+            return true;
+        } else {
+            return false;
+        }
```
megalord: How about `return rowId === 'option4_id'`?

src/rxForm/rxForm.js
```diff
@@ -153,7 +154,8 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
             model: '=',
             fieldId: '@',
             required: '=',
-            emptyMessage: '@'
+            emptyMessage: '@',
+            disableFn: '&'
```
megalord: Will you please add the new attributes to the README?

src/rxForm/README.md
```diff
@@ -42,4 +42,11 @@ Given an data object and an additional optional object for column labels, rxForm
 - `data`: A data object to prefill the radio/checkbox collection. For checkboxes, checked values default to true unless `value` and `falseValue` attributes are given. See the samples.
 - `model`: The AngularJS model to tie all radios/checkboxes together.
 - `field-id`: The ID of the elements.
-- `empty-message`: (string) A default message if the data attribute is empty. 
+- `empty-message`: (string) A default message if the data attribute is empty.
+- `disable-fn`: (Function) Callback function to determine if option should be disabled
```
parlarjb: Could you document here the parameters that will be passed to the function?

glynnis:
I'm not comfortable with where we are on design for this yet. I think we have a little more work to do to make sure the UX is good. Right now the contrast between "enabled" and "disabled" for an option is not high enough -- it's hard to tell from just the radio button alone that an option is disabled.

screen_shot_2015-03-06_at_3_49_11_pm

@ElementE Do we have any treatments that you've done in the past that would help here? I'm thinking that if the option is disabled, all the text for that radio button should show as the disabled gray color. What do you think?

rogerza:
@glynnis I totally agree. But there were no wireframes, since this was just logged as a defect in EC.

ElementE:
@glynnis Yeah thats definitely needed. I agree, full row text color to disabled.

rogerza:
How is this? The cursor also changes to the 'not-allowed' symbol over the radio, but the screenshot didn't capture it.
image

glynnis:
@rogerza This is better. The "not-allowed" symbol is a nice addition -- is it possible to put that over the entire row? For even better UX, you could show a tool-tip on mouseover (anywhere on the row), which can display a small amount of text explaining why something is disabled, so the user understands why an option isn't available (allowing them to understand how to enable it if it's actually the option they need).

The contrast between text color and the gray rows is pretty low, but I think that's probably okay given that the option is disabled. @ElementE What do you think?

ElementE:
This def works better. I think this is a solid treatment for disabled rows.

rogerza:
@glynnis I think the tooltip is a good idea, but the Angular-UI Bootstrap plugin for tooltips does funky things to the table if I apply it to the whole row (e.g. changes the row's bg and border color). Would you be okay without the tooltip for now? I forgot to mention that I also set the cursor to change over the disabled option - from the pointing hand to the default arrow, which helps the UX too. As far as the contrast goes, if you look at the screenshot from the EC PR, it's much more noticeable when there are multiple disabled rows

glynnis:
@rogerza Yeah, I think without the tooltip is fine for now :+1:

LGTM

src/rxForm/rxForm.js
```diff
@@ -268,4 +281,40 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+ */
+.service('rxFormUtils', function () {
+
+    // Returns the selected rows for all rxFormOptionTables with id: tableId
+    this.rxFormOptionTableSelected = function (tableId) {
+
+        var selectedRows = [];
+
+        // Find tables by id ($document doesn't support find by element id, so had to use document)
+        var tables = document.querySelectorAll('rx-form-option-table#' + tableId);
```
parlarjb: There should only be one instance of a given ID, how come you're doing `querySelectorAll`?

rogerza: This is in case you have multiple tables sharing the same id (which I did in EC). It makes finding the matching data for the row much simpler in the controller.

parlarjb: Two elements on the same page are not allowed to have the same ID

halkon: I would say that in that case we need to remove the duplicate ID's, but that's just me

halkon: Ok maybe not just say, I think we should remove all duplicate ID.

rogerza: sure, no problem. i can change it.

TheNando: For what it's worth: "What makes attributes of type ID special is that no two such attributes can have the same value in a conformant document, regardless of the type of the elements that carry them". It's strange that with that in the spec, Chrome and Firefox don't seem to complain when you reuse an ID. The querySelectorAll even works as 'expected'.

http://dev.w3.org/csswg/selectors-3/#id-selectors

src/rxForm/rxForm.js
```diff
@@ -268,4 +281,52 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+        return selectedRow;
+    };
+
+    // Returns the selected option within the tabset
+    // @param {String} tabsetId - The id of the tabset
+    // @returns {object} The tableId, fieldId, and rowId of the selected option
+    self.getSelectedOptionForTabSet = function (tabsetId) {
+        var selectedOption;
+        var xpathToTable = '//div[@id="' + tabsetId +
+            '"]//tr[@class="ng-scope selected"]//ancestor::rx-form-option-table';
```
parlarjb: Can you take the `ng-scope` out? I prefer not to rely on internal Angular implementation details. 

And is XPath necessary here? I don't think we use it anywhere else in the framework. What are you doing that a regular selector can't handle?

rogerza: Yeah, I can remove that class. I used xpath because it was way, way faster and cleaner than css in finding the table from the tabSet id. CSS doesn't allow you to query for a parent, so you'd have to call the parent 4 or 5 times, which is very slow.

parlarjb: :+1:

halkon: One of those rare moments (and :cry: :crying_cat_face: I had learned the fun of xpath only to never use it again, waaah )

src/rxForm/rxForm.js
parlarjb: Could you please mark this as '&?', to show that it's optional, and note in the README that it's an optional parameter?

src/rxForm/rxForm.js
```diff
@@ -140,12 +140,19 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+```
+ disable-fn="disableOption(tableId, fieldId, rowId)"
+```
  */
 .directive('rxFormOptionTable', function ($interpolate) {
     return {
         restrict: 'E',
         templateUrl: 'templates/rxFormOptionTable.html',
         scope: {
+            id: '@',
```
parlarjb: This is also optional, and only used when `disableFn` is used, right? If so, put it down at the bottom of the list with `disableFn`, and add it to the docs

rogerza: Actually, it doesn't look like this is necessary. Even without it, the id is still applied, so I will remove it.

src/rxForm/rxForm.js
```diff
@@ -164,6 +171,11 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
                 return (val1 == val2);
             };
 
+            $scope.checkDisabled = function (row) {
+                var rowData = { tableId: $scope.id, fieldId: $scope.fieldId, rowId: row.id };
```
parlarjb: Now this isn't right, because `$scope.id` isn't correct. You shouldn't have `id` as a scope variable anyway (and I know you removed it) because `id` is a reserved attribute name for every HTML element.

Running this locally, the value getting passed in for `tableId` is always `undefined`, so I think we're just "lucky" that this is working right now, because your `disableOption` function wasn't actually checking the `tableId`.

You should instead have `$element` injected into the controller, and do:

```
var rowData = { tableId: $element.attr('id'), fieldId: $scope.fieldId, rowId: row.id };
```

(And not your fault, but this shouldn't even be in a controller, this whole directive should use a `link` function instead. Whoever put it in a controller was wrong. I'll file an issue to change that later)

rogerza: :sob: So it looks like I did have a reason for having it. I'll change it to use $element.

src/rxForm/rxForm.js
```diff
@@ -164,6 +171,11 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+                return $scope.disableFn(rowData);
```
CITguy: Do we need an object? Can we get away with just passing the values directly to the function as arguments (preferrably in increasing specificity -- see below)?

```javascript
$scope.checkDisabled = function (row) {
  var tableId = $element.attr('id');
  return $scope.disableFn(tableId, row.id, $scope.fieldId);
};
```

src/rxForm/rxForm.js
rogerza: Actually, the $element will be the <tr>, so i'll have to trace back to the parent <rx-form-option-table> again, probably with xpath for performance reasons. @parlarjb, how do you feel about that?

src/rxForm/rxForm.js
rogerza: yeah, no object would work, too.

src/rxForm/rxForm.js
parlarjb: No, the `$element` should be the `rx-form-option-table`. Were you seeing something different?

rogerza: yeah, that makes sense. I may have spoke to soon because the midways were failing. let me check again.

src/rxForm/rxForm.js
```diff
@@ -268,4 +280,52 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+        var result = document.evaluate(xpathToTable, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null);
+        if (result.singleNodeValue) {
+            var table = result.singleNodeValue;
+            var fieldId = table.getAttribute('field-id');
+            var rowId = self.getSelectedOptionForTable(table.id).rowId;
+            selectedOption = { tableId: table.id, fieldId: fieldId, rowId: rowId };
+        }
+        return selectedOption;
+    };
+
```
CITguy: Please use `.factory()` with an explicit `return` (this is the pattern used in `encore-ui-svcs`).

```javascript
.factory('rxFormUtils', function () {
  return {
    getSelectedOptionForTable: function (tableId) { ... }
    getSelectedOptionForTabSet: function (tabsetId) { ... }
  };
})
```

src/rxForm/rxForm.js
```diff
@@ -268,4 +280,52 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
...
+ * @ngdoc service
+ * @name encore.ui.rxModalForm:rxFormUtils
+ * @description
+ * Set of utility functions used by rxForm to access form data
+ *
+ * @example
+ * <pre>
+ * // Returns an array of selected options in all tables with id tableId
+ * // [{ tableId: 'tableId', fieldId: 'fieldId', rowId: 'rowId' }]
+ * rxFormUtils.rxFormOptionTableSelected(tableId)
```
CITguy: Where is `rxFormOptionTableSelected` defined? Please update ngdoc.

src/rxForm/rxForm.js
halkon: Just a note on this, both patterns have been used both in `encore-ui` ( @parlarjb / @CITguy call me out if I am wrong ).   However this is more of a utility belt, not something you need an instance of anything, so using service is an unnecessary extra step. 

The approach @CITguy would be fine, also, this next one I think would suffice (specially when documenting functions, might be more readable).

```javascript
.factory('rxFormUtils', function () {
  var getSelectedOptionForTable = function () {
     ...
  };
  var getSelectedOptionForTabSet: function () {
    ...
   };

   return {
    getSelectedOptionForTable: getSelectedOptionForTable,
    getSelectedOptionForTabSet: getSelectedOptionForTabSet
   };

});```

I just wanted to share that, I'll let @parlarjb & @CITguy provide the proper feedback, just wanted to explain more as to why using a service is not necessary mainly

Comment (parlarjb) [inline src/rxForm/rxForm.js]: Since we're already down the rabbit hole, my preferred way to do this is actually:

```
.factory('rxFormUtils', function() {
    var rxFormUtils = {};
    rxFormUtils.getSelectedOptionForTable = function() {
        ...
    };
    rxFormUtils.getSelectedOptionForTabSet: function() {
        ...
    };

    return rxFormUtils;
});
```

(I originally saw this pattern in http://toddmotto.com/opinionated-angular-js-styleguide-for-teams/ and I agree with it)

Comment (halkon) [inline src/rxForm/rxForm.js]: Oh hey! look at that... that's yet another pattern that's been used here, specifically in `encore-ui-svcs` Cloud Transforms.

We just need to stick with one and go with it, I actually haven't had enough coffee in my belly this morning so I forgot about this one.  This one would be better.

Comment (rogerza) [inline src/rxForm/rxForm.js]: I like this one, too, so I'll go with this. :)

Comment (CITguy) [inline src/rxForm/rxForm.js]: Yeah, all I care about is the explicit return. :smile: And we've used all three types of explicit returns in `encore-ui-svcs` (even in the [same file](https://github.com/rackerlabs/encore-ui-svcs/blob/master/app/scripts/transform.js)).

* [`return <OBJECT/FUNCTION>`](https://github.com/rackerlabs/encore-ui-svcs/blob/master/app/scripts/encore.js#L16)
* [`function foo () {...}; return { foo: foo };`](https://github.com/rackerlabs/encore-ui-svcs/blob/master/app/scripts/transform.js#L76)
* [`var svc = {}; svc.foo = function () {...}; return svc;`](https://github.com/rackerlabs/encore-ui-svcs/blob/master/app/scripts/encore.js#L119)

Comment (CITguy) [inline src/rxForm/rxForm.js]: Having said that, I do prefer the third way.

```javascript
.factory('rxFormUtils', function() {
    var rxFormUtils = {};
    rxFormUtils.getSelectedOptionForTable = function() {
        ...
    };
    rxFormUtils.getSelectedOptionForTabSet: function() {
        ...
    };

    return rxFormUtils;
});
```

CITguy:
DEV - LGTM

parlarjb:
LGTM

src/rxForm/rxForm.js
```diff
@@ -164,6 +171,10 @@ angular.module('encore.ui.rxForm', ['ngSanitize', 'encore.ui.rxMisc'])
                 return (val1 == val2);
             };
 
+            $scope.checkDisabled = function (row) {
+                return $scope.disableFn({ tableId: $element.attr('id'), fieldId: $scope.fieldId, rowId: row.id });
```
halkon: I know there's been a few iterations of this one but:

`$attrs` is an injectable. 

https://docs.angularjs.org/api/ng/service/$compile#-controller-

Might be better than `$element`

halkon: Awww Merged! it's alright, next time!

PR #849, opened 2015-03-18
rxPaginate support for paginated APIs

parlarjb:


src/rxPaginate/docs/rxPaginate.js
```diff
@@ -33,4 +29,60 @@ function rxPaginateCtrl ($scope, PageTracking) {
...
+    var serverInterface = {
+        getItems: function (pageNumber, itemsPerPage, params) {
+            var deferred = $q.defer();
+            var filterText = params.filterText;
+            var sortColumn = params.sortColumn;
+            var sortDirection = params.sortDirection;
+
+            if (sortColumn === 'name') {
+                sortColumn = 'id';
+            }
```
CITguy: Why are we overriding the sortColumn if it's `'name'`? What if I want to sort by name?

parlarjb: This is just the demo app JS. I'm sorting by `id` so "Server 45" will come before "Server 100"

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/849/gallery.html)

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+        <td colspan="2">
+            <rx-paginate page-tracking="pager"></rx-paginate>
+        </td>
+    </tr>
+</tfoot>
+```
+
+Here we are using the `<rx-paginate>` directive to draw the buttons, passing it the same `pager` instance described above.
+
+Because all of the `servers` get passed via `ng-repeat`, it means you don't need to take explicit action w.r.t. pagination if the set of data changes. You can change `$scope.servers` at any time, and `<rx-paginate>` will just re-process.
```
CITguy: I would expand `w.r.t.` to reduce any confusion for people who aren't familiar with the acronym.

CITguy:
For the README, I'm seeing some inconsistency with presentation of each type of pagination. 

The UI-Pagination starts with "here's the view", and continues with "let me explain what you're looking at" (explaining an example). However, the API-Pagination starts with "define this in the controller, and follows with "then do this in the view", etc. (follow these instructions).

Please edit README to use one form or the other.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+# UI-Based Pagination
+With UI-Based pagination, the entire set of data is looped over via an `ngRepeat` in the table's `<tbody>`, with the data passed into the `Paginate` filter. This filter does the work of paginating the set of data and communicating with the `<rx-paginate>` to draw the page selection buttons at the bottom of the table.
+
+As shown in the first example below, the `ngRepeat` will usually look like this:
+
+```
+ <tr ng-repeat="server in servers | orderBy: sorter.predicate:sorter.reverse | Paginate:pager ">
+```
+
+In this case, `servers` is a variable bound to your page `$scope`, and contains the full set of servers. This is then passed to `orderBy`, to perform column sorting with `rxSortableColumn`. The sorted results are then passed to `Paginate:pager`, where `Paginate` is a filter from the `rxPaginate` module, and `pager` is a variable on your scope created like `$scope.pager = PageTracking.createInstance();`. This `pager` is responsible for tracking pagination state (i.e. "which page are we on", "how many items per page", "total number of items tracked", etc.
```
CITguy: This might be better by implementing a numbered list.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+ 1. We now loop over a variable provided by the pager. It is the pager's responsibility to fill that variable with data. Its source for the data will be discussed below.
+ 2. We no longer pass the values through _any_ filters. Not a search text filter, not sorting filter, and not the `Paginate` filter.
+
+** BEGIN WARNING ** 
+
+You should _never_ access `lazyPager.items` from anywhere other than the `ng-repeat`. Do not touch it in your controller. It is a dynamic value that can
+change at anytime. It is only intended for use by `ng-repeat`.
+
+** END WARNING **
```
CITguy: Is there any possibility of doing a `server in lazyPager` (removing the need to directly access `lazyPager.items`)?

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+To use API-based pagination, you still start by creating a `PageTracking` instance on the scope.
+
+```
+$scope.lazyPager = PageTracking.createInstance();
+```
+
+Then in your `ngRepeat`, do the following:
+
+```
+<tr ng-repeat="server in lazyPager.items">
```
CITguy: IMO, I'd rename `lazyPager` to `pagedServers`.  This would be more understandable in code as `pagedServers.items`.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+** END WARNING ** 
+
+For the `<tfoot>`, it will look like this:
+
+```
+<tfoot>
+    <tr class="paginate-area">
+        <td colspan="2">
+            <rx-paginate page-tracking="lazyPager" server-interface="serverInterface" filter-text="searchText" sort-column="sort.predicate" sort-direction="sort.reverse"></rx-paginate>
```
CITguy: Please split attributes into multiple lines. User currently has to scroll horizontally to view example.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+        </td>
+    </tr>
+</tfoot>
+```
+
+`page-tracking` still receives the pager (`lazyPager` in this case) as an argument. What's new are the next four parameters.
+
+`server-interface` _must_ be present. It has to be passed an object with a `getItems()` method on it. This method is what `<rx-paginate>` will use to request data from the paginated API.
+
+`filter-text`, `sort-column` and `sort-direction` are all optional. If present, `<rx-paginate>` will watch the variables for changes, and will call `getItems()` for updates whenever the values change.
```
CITguy: The above three paragraphs may make more sense as an unordered list.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+## getItems()
```
CITguy: This section may make more sense as "Custom Pagination Interface" with "getItems()" as a subsection.  Jumping directly to this section, I have no context of where to define `getItems()` without first reading the explanation of the `<rx-paginate>` `server-interface` in the section above.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+where:
+
+ - `pageNumber`: the 0-based page of data being requested
+ - `itemsPerPage`: the value the user currently has selected for how many items per page they wish to see
+ - `filterText`: the filter search string entered by the user, if any
+ - `sortColumn`: the name of the selected sort column, if any
+ - `sortDirection`: either `'ASCENDING'` or `'DESCENDING'`
+
+When the framework calls `getItems()`, you _must_ return a promise. When this promise resolves, the resolved value must have the following properties on it:
```
CITguy: I'd make this bold like: `... you **_must_ return a promise**. When ...`

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+ - `items`: The actual items/rows of the table returned for the request. This should at least contain `itemsPerPage` items, if that many items exist on the given page
```
CITguy: I would indicate that this needs to be an array or iterable collection.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+ - `pageNumber`: The page number that these items belong to. Normally this should be the same as the `pageNumber` value passed to `getItems()`
```
CITguy: Is this 1 or 0 based? Where is it used?

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+ - `totalNumberOfItems`: The total number of items available, given the `filterText` parameter.
```
CITguy: Can we simplify this to `itemCount`?

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+Examples are below.
+
+## `totalNumberOfItems`
+
+As mentioned, `totalNumberOfItems` has to be the total items available given the search parameters. For example, say the following request was made:
```
CITguy: Maybe...

> If you could get all items from the API in *one call*, `totalNumberOfItems` would reflect the number of items returned (given necessary search parameters).

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+```
+var pageNumber = 0;
+var itemsPerPage = 50;
+getItems(pageNumber, itemsPerPage);
+```
+
+This is asking for all the items on page 0, with the data that the user is currently viewing 50 items per page. A valid response would return 50 items. However, the _total_ number of items available might be 1000 (i.e. 20 pages of results). Your response must then have `totalNumberOfItems: 1000`. This data is needed so we can display to the user "Showing 1-50 of 1000 items" in the footer of the table.
```
CITguy: > If you could retrieve *all values* from the server at once, and it returned 1000 items, `totalNumberOfItems` would be `1000`. However, `items` would only contain the first 50 items because that's what we actually requested (50 `itemsPerPage` starting on `pageNumber` 0).

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+var pageNumber = 0;
+var itemsPerPage = 50;
+var opts = {
+    filterText: "Ubuntu"
+};
+getItems(pageNumber, itemsPerPage, opts);
+```
+
+This means "Filter all your items by the search term 'Ubuntu', then return page 0". If the total number of items matching "Ubuntu" is 200, then your response would have `totalNumberOfItems: 200`.
+
```
CITguy: All of the above can be replaced with:

> This also holds true when you apply `filterText`, `sortColumn` and `sortDirection` options.

src/rxPaginate/README.md
CITguy: When would it *not* be the same as `pageNumber` passed to `getItems()`?

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+## Local Caching
+
```
CITguy: Maybe add

> If you plan on calling the API every time you navigate through paged data, this section will not apply to you.
> If you need/want to reduce the number of times you call your API, please read on.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+When a `getItems()` request is made, the framework passes in the user's `itemsPerPage` value. If it is 50, and there are 50 results available for the requested page, then you should return _at least_ 50 results. However, you may also return _more_ than 50 items.
```
CITguy: maybe set this section under a `### without caching (default)`

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+For example, say the following request is made when the page first loads:
+
+```
+var pageNumber = 0;
+var itemsPerPage = 50;
+getItems(pageNumber, itemsPerPage);
+```
+
+Because no data is available yet, `<rx-paginate>` will call `getItems()`, wait for the response, and then draw the items in the table. If you returned exactly 50 items, and the user then clicks "Next" or 2 (to go to the second page), then `getItems()` will have to be called again (`getItems(1, 50)`), and the user will have to wait for the results to come in.
```
CITguy: Maybe change to:

> Initially, `<rx-paginate>` will call `getItems()`, wait for a response, and then update items in the table.  If your custom interface `itemsPerPage` is equal to its `items.length`, and the user navigates to a different page of data, `getItems()` will be called again to fetch new information from the API.  The user will then need to wait before they see new data in the table. This remains true for every interaction with page data navigation.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+But if instead of returning 50 items on the initial `getItems()` call, you returned 200 items (i.e. four pages worth of results), then the framework will cache the extra values. Then, when the user clicks to go to page 2, they will immediately see the results, there will be no call to `getItems()` made. Once they click to go to page 5, a new request will be made through `getItems()`.
```
CITguy: Maybe

> However, if your custom interface were to pull more than `itemsPerPage` of data from the API, `<rx-paginate>` is smart enough to navigate through the saved data without needing to make an API request every time the page is changed.
> There are some caveats, though.
> 1. `items.length` must be a multiple of `itemsPerPage` (if `itemsPerPage = 50`, `items.length` must be 50, 100, 150, etc.)
> 2. You will need to calculate the page number sent to the API based on requested values in the UI.

The second item would be useful to have a table to show how the values change on the UI versus the Server/API.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+Internally, calling `refresh()` equates to `<rx-paginate>` doing a new `getItems()` call, with the current filter/sort criteria. But the point is that you can't just call `getItems()` yourself to cause an update. The framework has to call that method, so it knows to wait on the returned promise.
+
+This is shown in the demo below with a "Refresh" button above the API-paginated demo.
+
+
+## Error Handling
+`<rx-paginate>` includes a simple way to show error messages when `getItems()` rejects instead of resolves. By passing `error-message="Some error text!"` to `<rx-paginate>`, the string entered there will be shown in an rxNotification whenever `getItems()` fails. If `error-message` is not specified, then nothing will be shown on errors. In either case, on a failure, the table will stay on the page it was on before the request went out.
+
+If you wish to show more complicated error messages (and you should!), then you'll have to do that yourself. Either put error handling code directly into your `getItems()`, or have something else wait on the `getItems()` promise whenever it's called, and perform the handling there.
```
CITguy: `... more complicated error messages (and it is highly suggested that you do!), then ...`

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+        rxPromiseNotifications.add(promise, {
+            error: 'Error loading page ' + pageNumber
+        }
+
+        return promise;
+    }
+}
+
+Then, you'd pass the `getItemsWrapper` as `<rx-paginate server-interface="getItemsWrapper" ... >`. This keeps UI-based error display code out of your `PageRequest.getItems()`, and puts it into a separate object that knows about your UI, your desired error message, etc.
+```
```
CITguy: You might want to move the triple backticks (```)above this paragraph.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+`<rx-paginate>` includes a simple way to show error messages when `getItems()` rejects instead of resolves. By passing `error-message="Some error text!"` to `<rx-paginate>`, the string entered there will be shown in an rxNotification whenever `getItems()` fails. If `error-message` is not specified, then nothing will be shown on errors. In either case, on a failure, the table will stay on the page it was on before the request went out.
+
+If you wish to show more complicated error messages (and you should!), then you'll have to do that yourself. Either put error handling code directly into your `getItems()`, or have something else wait on the `getItems()` promise whenever it's called, and perform the handling there. 
+
+One way to do this is as so:
+
+Let's say that you had some `PageRequest` object, on which your `getItems()` was defined. You probably want your `getItems()` to be unaware of the UI, i.e. you don't want UI logic in there, just logic for dealing with the server. Then, instead of passing your `PageRequest` to `<rx-paginate>`, do something like this:
+
+```
+var getItemsWrapper = {
```
CITguy: If you mention a `PageRequest` object, you might want to use it in your example.

src/rxPaginate/README.md
```diff
@@ -1,6 +1,267 @@
...
+
+```
+var vals = rxPaginateUtils.calculateApiVals(pageNumber, itemsPerPage, serverItemsPerPage);
+items = requestToServer(vals.serverPageNumber, serverItemsPerPage);
+return items.slice(vals.offset);
+```
+
+If the user enters any search text, and you've passed the search field to `<rx-paginate>` via `search-text`, then the cache will be immediately flushed and a new request made. 
+
+## Forcing a Reload
```
CITguy: If the method is called `refresh()`, maybe change the section to "Forcing a Refresh".

src/rxPaginate/README.md
parlarjb: It might change if a bunch of things got deleted, and we have to send the user back to a different page than they expected.

src/rxPaginate/README.md
parlarjb: I originally had something like that. The reason I changed it was to make it as clear as possible that this is _not_ a count for the `items` property, but something else entirely.

src/rxPaginate/README.md
parlarjb: No, that's not correct. Given the caching, you can return more than 50 items in the `items` attribute.

src/rxPaginate/README.md
parlarjb: You want me to get rid of the example as well?

src/rxPaginate/README.md
parlarjb: Sorry, can you clarify what you mean here?

src/rxPaginate/README.md
parlarjb: oooooooppppppsssss

src/rxPaginate/README.md
CITguy: Well, since we qualify with `(given necessary search parameters)` on line 124, I don't see much point keeping this section pertaining to `filterText`. It behaves the same way, just with more search parameters.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/849/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/849/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/849/gallery.html)

src/rxPaginate/docs/rxPaginate.html
```diff
@@ -28,4 +29,67 @@
         <button ng-click="removeServers()" class="button sm negative">Remove first 2 servers</button>
         <button ng-click="addServers()" class="button sm positive">Add 2 servers</button>
     </p>
+    
+    <h2 class="title">API-Based Pagination</h2>
+    <p>The API used by this demo is returning 150 items at a time. This means with the default `itemsPerPage` of 25, five pages of results are coming back at a time. You should be able to click through pages 1-5 without a loading message, and then the loading message will appear for page 6.</p>
```
megalord: If 150 items are being loaded, and there are 25 items per page, shouldn't that populate six pages of results (not five)?

src/rxPaginate/docs/rxPaginate.js
```diff
@@ -33,4 +29,60 @@ function rxPaginateCtrl ($scope, PageTracking) {
...
+                sortColumn = '-' + sortColumn;
+            }
+
+            $timeout(function () {
+                var first = pageNumber * itemsPerPage;
+                var added = first + itemsPerPage;
+                var last = (added > allLazyServers.length) ? allLazyServers.length : added;
+
+                first = first;
+                last = last;
```
megalord: What is this for?

src/rxPaginate/docs/rxPaginate.js
```diff
@@ -33,4 +29,60 @@ function rxPaginateCtrl ($scope, PageTracking) {
...
+
+                var filteredServers = $filter('filter')(allLazyServers, filterText);
+                filteredServers = $filter('orderBy')(filteredServers, sortColumn);
+                var lazyServers = filteredServers.slice(first, last + 100);
```
megalord: Where is the `100` coming from? Should that be `itemsPerPage`, or perhaps a multiple of it?

src/rxPaginate/docs/rxPaginate.midway.js
```diff
@@ -1,14 +1,119 @@
 var exercise = require('../rxPaginate.exercise.js');
+var rxSortableColumn = require('../../rxSortableColumn/rxSortableColumn.page').rxSortableColumn;
+var Page = require('astrolabe').Page;
```
megalord: This is a red flag to me - mixing page object definitions with the test code.   I can see it has been done here because the page being created is specific to the demo.  I guess it would be nice if there was a way in the tasks to define pages for the demo that are not included in the build, but that is probably a larger conversation.

src/rxPaginate/docs/rxPaginate.midway.js
```diff
@@ -1,14 +1,119 @@
...
+        var osColumn = table.column('OS');
+
+        beforeEach(function () {
+            table.filter = '';
+            nameColumn.sortAscending();
+        });
+
+        it('should get new items when filter text is entered', function () {
+            table.filter = 'Ubuntu';
+            browser.driver.sleep(3000);
```
megalord: I assume these delays correspond to the `$timeout` in the demo (used for `getItems()`).  I am under the impression that protractor automatically waits for $timeouts, $intervals, and $http requests by default.  Were these tests failing before the sleep was added?

src/rxPaginate/docs/rxPaginate.midway.js
```diff
@@ -1,14 +1,119 @@
...
+            },
 
+            column: {
+                value: function (columnName) {
+                    var column = element(by.cssContainingText(tableSelector + ' rx-sortable-column', columnName));
+                    return rxSortableColumn.initialize(column, repeaterString);
+                }
+            },
+
+            columns: {
```
megalord: I don't see this being used anywhere.

src/rxPaginate/rxPaginate.js
```diff
@@ -6,24 +6,46 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage'])
...
+            numberOfPages: '@',
+            serverInterface: '=?',
+            filterText: '=?',
+            sortColumn: '=?',
+            sortDirection: '=?'
         },
-        link: function (scope, element) {
+        link: function (scope, element, attrs, rxLoadingOverlayCtrl) {
+
+            var errorMessage = attrs.errorMessage;
```
megalord: Will you add this attribute to the ngdoc definition above?

src/rxPaginate/rxPaginate.js
```diff
@@ -34,17 +56,161 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage'])
...
+
+                var changedVal = function (newVal, oldVal) {
+                    return newVal !== oldVal;
+                };
+
+                // Whenever the filter text changes (modulo a debounce), tell
+                // the PageTracker that it should go grab new items
+                if (!_.isUndefined(scope.filterText)) {
+                    scope.$watch('filterText', function (newVal, oldVal) {
+                        if (changedVal (newVal,  oldVal)) {
```
megalord: extra space

src/rxPaginate/rxPaginate.js
```diff
@@ -34,17 +56,161 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage'])
...
+                        }
+                    });
+                }
+
+                if (!_.isUndefined(scope.sortColumn)) {
+                    scope.$watch('sortColumn', function (newVal, oldVal) {
+                        if (changedVal(newVal, oldVal)) {
+                            columnOrDirectionChange();
+                        }
+                    });
```
megalord: If we were die hard about making this dry:
```javascript
function ifChanged (fn) {
  return function (newVal, oldVal) {
    if (newVal !== oldVal) {
      fn();
    }
  }
}

scope.$watch('filterText', ifChanged(textChange));
scope.$watch('sortColumn', ifChanged(columnOrDirection));
scope.$watch('sortDirection', ifChanged(columnOrDirection));
```
But I'm really just amusing myself at this point.

src/rxPaginate/rxPaginate.js
```diff
@@ -114,8 +285,167 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage'])
...
+        });
+        
+        function updateCache (pager, pageNumber, localItems) {
+            var numberOfPages = Math.floor(localItems.length / pager.itemsPerPage);
+            var cachedPages = numberOfPages ? _.range(pageNumber, pageNumber + numberOfPages) : [pageNumber];
+            pager.cachedPages = !_.isEmpty(cachedPages) ? cachedPages : [pageNumber];
+            pager.cacheOffset = pager.cachedPages[0] * pager.itemsPerPage;
+        }
+
+        updateCache(settings, 0, settings.localItems);
```
megalord: With all the changes to this factory, `settings` doesn't seem like the right name for the object.  It became pretty clear at this point, when the `settings` object is provided as the `pager` argument, but the code below also has that feel to it.  It may be as simple as `s/settings/pager/g`, but perhaps there is also a better way to organize this.

src/rxPaginate/rxPaginate.js
```diff
@@ -114,8 +285,167 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage'])
...
+        // This returns the slice of data for whatever current page the user is on.
+        // It is used for server-side pagination.
+        Object.defineProperty(settings, 'items', {
+            get: function () {
+                var info = rxPaginateUtils.firstAndLast(settings.pageNumber, settings.itemsPerPage, settings.total);
+                return settings.localItems.slice(info.first - settings.cacheOffset, info.last - settings.cacheOffset);
+            }
+        });
+    
+        Object.defineProperty(settings, 'totalPages', {
```
megalord: No love for `Object.defineProperties()`?

src/rxPaginate/rxPaginate.spec.js
```diff
@@ -180,11 +184,11 @@ describe('Pagination', function () {
             // clicking link should move to last page
             helpers.clickElement(link[0]);
 
-            expect(scope.pager.pageNumber).to.equal(scope.pager.totalPages - 1);
+            expect(scope.pager.isLastPage()).to.be.true;
         });
-
+
```
megalord: :astonished:

src/rxPaginate/rxPaginate.spec.js
```diff
@@ -233,6 +237,186 @@ describe('Pagination', function () {
...
+        beforeEach(function () {
+            // Load the directive's module
+            module('encore.ui.rxPaginate');
+            module('templates/rxPaginate.html');
+            module('encore.ui.rxMisc');
+            module('encore.ui.rxNotify');
+
+            // Inject in angular constructs
+            inject(function ($rootScope, $compile, $q, _$timeout_, PageTracking) {
+                $timeout = _$timeout_;
```
megalord: This isn't consistent with the pattern I have noticed (in this repo) for making services available to tests, as seen on line 288.
```javascript
$timeout = _$timeout_;
// vs
timeout = $timeout;
```

src/rxPaginate/rxPaginate.spec.js
```diff
@@ -320,7 +499,7 @@ describe('Pagination', function () {
 
         it('should switch page if item deletion causes us to be past last page', function () {
             // Go to last page
-            pager.pageNumber = pager.totalPages - 1;
+            pager.goToLastPage();
```
megalord: The previous spec could use the same treatment (`.pageNumber` -> `.goTo***`).

src/rxPaginate/docs/rxPaginate.midway.js
parlarjb: Could you clarify what you mean here?

parlarjb: I originally spoke with @Droogans about this. I _think_ the plan was that we'd eventually create a very generic table object like the one here, but useable elsewhere.

Droogans: It only exists for testing, so I usually put what I call "anonymous" page objects *at the top* of specs, before any `describe` blocks are defined.

So, @parlarjb I'd rather see the page object defined at the top of the file. @megalord, I am hesitant to put the anonymous demo page objects anywhere but the midway tests, because then I can't share it with other people as easily when they need an example of how to do it, as in http://rackerlabs.github.io/encore-ui/#/component/rxForm (see the "Protractor" tab).

megalord: Yeah, it definitely needs to be accessible in the online demo.  I think the "anonymous" idea is an improvement over the way it is now.

Droogans: After considering how to do it better, I think I could follow the example set in https://github.com/rackerlabs/encore-ui/blob/ff1917c6d642c052095376d6218e7412894a2cb7/grunt-tasks/modules.js#L48-L50, where I explicitly tacked on the exercises files onto the end of the "Protractor" tab.

It wouldn't be much work to do the same with any demo page objects as well. https://github.com/rackerlabs/encore-ui/issues/878

src/rxPaginate/docs/rxPaginate.midway.js
parlarjb: Good catch. Previously I was using the lodash `debounce` function, which is not digest-cycle aware, so this was required. Since then I moved to a `$timeout` based solution, so these aren't needed anymore. I'll take them out.

src/rxPaginate/docs/rxPaginate.midway.js
parlarjb: copy/paste artifact :)

src/rxPaginate/rxPaginate.js
parlarjb: Actually, I shouldn't even have the `changedVal`. That was there for a previous attempt at something, and now it's just silly to have around.

src/rxPaginate/rxPaginate.js
parlarjb: I agree 100%. I'll make the change.

src/rxPaginate/rxPaginate.js
parlarjb: I didn't know about `defineProperties`!

src/rxPaginate/rxPaginate.spec.js
parlarjb: http://en.wikipedia.org/wiki/Seppuku

src/rxPaginate/rxPaginate.spec.js
parlarjb: I'm trying to move things to the latter as I come across them

parlarjb: Or at least, whenever I touch one of them

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/849/gallery.html)

kobowley:
LGTM

megalord:
:+1:

CITguy:
LGTM

PR #852, opened 2015-03-19
feat(rxApp): show user name next to logout link.

### LINKED ISSUE #169: Add username next to Logout link in top left
Per the design, the username of the person logged in to encore should be available in the top left corner:
image

/cc @hussamd

megalord:
Fixes #169 

screen shot 2015-03-19 at 1 44 28 pm

### DIFF
--- a/src/rxApp/docs/rxApp.midway.js
+++ b/src/rxApp/docs/rxApp.midway.js
@@ -22,6 +22,22 @@ describe('rxApp', function () {
         expect(rxAppCustom.sectionTitle).to.eventually.equal('Example Menu');
     });
 
+    it('should have the user id in the logout link', function () {
+        browser.addMockModule('encore.ui.rxSession', function () {
+            angular.module('encore.ui.rxSession', []).value('Session', {
+                getUserId: function () { return 'rack0000'; }
+            });
+        });
+        browser.refresh()
+
+        rxAppStandard.userId.then(function (userId) {
+            expect(userId).to.equal('(rack0000)');
+
+            browser.removeMockModule('encore.ui.rxSession');
+            browser.refresh()
+        });
+    });
+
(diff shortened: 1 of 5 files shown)


src/rxApp/docs/rxApp.midway.js
```diff
@@ -22,6 +22,22 @@ describe('rxApp', function () {
         expect(rxAppCustom.sectionTitle).to.eventually.equal('Example Menu');
     });
 
+    it('should have the user id in the logout link', function () {
```
megalord: I admit this test is a little hacky, since it makes use of `.addMockModule()`, but I felt it was still helpful enough to include.  Another alternative is to just run
```
localStorage.setItem('encoreSessionToken', JSON.stringify({ access: { user: { id: 'rack0000' } } }));
localStorage.removeItem('encoreSessionToken');
```
but that seems more fragile to me.  Of course, I can remove this test if you all see fit.

glynnis:
:clap: :clap: :clap: :clap: :clap: 

Design: LGTM

parlarjb:
LGTM, but I wouldn't mind input from @Droogans on the `browser.addMockModule` usage

src/rxApp/docs/rxApp.midway.js
Droogans: I think it'd be worth it for this to abstracted away [via `rxLocalStorage`](https://github.com/rackerlabs/encore-ui/tree/master/src/rxLocalStorage) to reduce the number of `executeScript` calls to make, which IMO, is kind of a smell. It should be encapsulated better.

I'll mark it as TODO and remove this `addMockModule` call with direct manipulation of the browser's local storage. There's no need to mock something as small and fast a local storage, especially if valid defects can be caught through tests that do actually use local storage directly.

PR #854, opened 2015-03-20
fix(rxFloatingHeader): Remove bad example code

parlarjb:
This demo code for showing a message wasn't correct. A proper way
to do this will come into the framework when we move the bulk select
stuff in.


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/854/gallery.html)

megalord:
:+1:

PR #856, opened 2015-03-20
fix(rxCharacterCount): Clean up the generated elements

### LINKED ISSUE #847: rxCharacterCount doesn't clean up when bound element is removed
If the input or textarea is removed from the dom (say, via an ng-if) the character count stays visible.

parlarjb:
Closes #847

Properly clean up the character count and other generated elements,
created when rx-character-count is used on an element.

This also moves the character count into the `.counted-input-wrapper`
div. Previously it sat outside of it.

BREAKING CHANGE
We've removed the `.character-count-wrapper` class. Previously we
set widths on this class. It ended up being redundant, widths
should now be set w.r.t. `.counted-input-wrapper`

### DIFF
--- a/src/rxCharacterCount/rxCharacterCount.js
+++ b/src/rxCharacterCount/rxCharacterCount.js
@@ -1,11 +1,31 @@
 angular.module('encore.ui.rxCharacterCount', [])
 .directive('rxCharacterCount', function ($compile) {
-    var counter = '<div class="character-countdown" ' +
-                  'ng-class="{ \'near-limit\': nearLimit, \'over-limit\': overLimit }"' +
+    var counterStart = '<div class="character-countdown" ';
+    var counterEnd =   'ng-class="{ \'near-limit\': nearLimit, \'over-limit\': overLimit }"' +
                   '>{{ remaining }}</div>';
 
-    var background = '<div class="input-highlighting"><span>{{ underLimitText }}</span>' +
+    var backgroundStart = '<div class="input-highlighting" ';
+    var backgroundEnd = '><span>{{ underLimitText }}</span>' +
                      '<span class="over-limit-text">{{ overLimitText }}</span></div>';
+
+    var extraDirectives = function (attrs) {
+        var extra = '';
+        if (_.has(attrs, 'ngShow')) {
+            extra += 'ng-show="' + attrs.ngShow + '" ';
+        }
+        if (_.has(attrs, 'ngHide')) {
+            extra += 'ng-hide="' + attrs.ngHide + '" ';
+        }
+        return extra;
+    };
+
+    var buildCounter = function (attrs) {
+        return counterStart + extraDirectives(attrs) + counterEnd;
+    };
+
+    var buildBackground = function (attrs) {
+        return backgroundStart + extraDirectives(attrs) + backgroundEnd;
+    };
 
     return {
         restrict: 'A',
@@ -19,13 +39,13 @@ angular.module('encore.ui.rxCharacterCount', [])
             var wrapper = angular.element('<div class="counted-input-wrapper" />');
             element.after(wrapper);
 
-            $compile(background)(scope, function (clone) {
+            $compile(buildBackground(attrs))(scope, function (clone) {
                 wrapper.append(clone);
                 wrapper.append(element);
             });
 
-            $compile(counter)(scope, function (clone) {
-                wrapper.after(clone);
+            $compile(buildCounter(attrs))(scope, function (clone) {
+                wrapper.append(clone);
             });
 
             var maxCharacters = _.parseInt(attrs.maxCharacters) || 254;
@@ -74,6 +94,7 @@ angular.module('encore.ui.rxCharacterCount', [])
 
             scope.$on('$destroy', function () {
                 element.off('input', writeLimitText);
+                wrapper.remove();
             });
         }
     };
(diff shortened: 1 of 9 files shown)


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/856/gallery.html)

src/rxCharacterCount/rxCharacterCount.js
```diff
@@ -1,12 +1,32 @@
 angular.module('encore.ui.rxCharacterCount', [])
 .directive('rxCharacterCount', function ($compile) {
-    var counter = '<div class="character-countdown" ' +
-                  'ng-class="{ \'near-limit\': nearLimit, \'over-limit\': overLimit }"' +
+    var counterStart = '<div class="character-countdown" ' ;
```
megalord: Looks like there is an extra space right before the `;` at the end of the line. :grin:

src/rxCharacterCount/rxCharacterCount.spec.js
```diff
@@ -120,6 +122,44 @@ describe('rxCharacterCount', function () {
...
+        });
+        
+        it('should hide the character count if the element hides (ngHide)', function () {
+            parentDiv = helpers.createDirective(ngHideTemplate, compile, originalScope);
+            el = parentDiv.find('textarea');
+            var characterCount = parentDiv.find('.character-countdown');
+            originalScope.hideTextArea = false;
+            originalScope.$apply();
+            expect(el.is(':visible'), 'textarea').to.be.false;
+            expect(characterCount.is(':visible'), 'character countdown').to.be.false;
```
megalord: When you set `hideTextArea` to `false`, I would think the the test should expect the element to be visible.  Am I missing something here?

parlarjb: hmmm.... wtf.... I'll look into it

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/856/gallery.html)

src/rxCharacterCount/rxCharacterCount.spec.js
```diff
@@ -120,6 +122,44 @@ describe('rxCharacterCount', function () {
...
+            parentDiv = helpers.createDirective(ngShowTemplate, compile, originalScope);
+            el = parentDiv.find('textarea');
+            var characterCount = parentDiv.find('.character-countdown');
+            originalScope.showTextArea = false;
+            originalScope.$apply();
+            expect(el.hasClass('ng-hide'), 'textarea').to.be.true;
+            expect(characterCount.hasClass('ng-hide'), 'character countdown').to.be.true;
+        });
+        
+        it('should hide the character count if the element hides (ngHide)', function () {
```
megalord: After the changes made in the last commit, this description is not quite accurate.  Maybe something like,
`'should show the character count if the element becomes visible (ngHide)'`

megalord:
LGTM

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/856/gallery.html)

src/rxCharacterCount/rxCharacterCount.js
```diff
@@ -1,12 +1,32 @@
...
+    var extraDirectives = function (attrs) {
+        var extra = '';
+        if (_.has(attrs, 'ngShow')) {
+            extra += 'ng-show="' + attrs.ngShow + '" ';
+        }
+        if (_.has(attrs, 'ngHide')) {
+            extra += 'ng-hide="' + attrs.ngHide + '" ';
+        }
+        return extra;
+    };
```
CITguy: How does this scale with additional angular attribute directives? What if I used `ng-when` to show/hide the element/directive?

CITguy: We can do our due diligence to catch the common use cases, but I feel like we need to suggest in the documentation that if developers wish to show/hide the anchored element, they'll need to place that logic outside of the element itself.

So, instead of doing this:
```html
<textarea rx-character-count ng-show="isShown">{{someValue}}</textarea>
```

To do something like this:
```html
<span ng-show="isShown">
  <textarea rx-character-count>{{someValue}}</textarea>
</span>
```
In its current state, this directive is highly unstable and prone to little bugs like what this PR is attempting to fix.  Until it stabilizes, I suggest we document workarounds.

CITguy: After doing some preliminary research, we need to handle:
* `ng-if`
* `ng-show`
* `ng-hide`
* `ng-switch-when`
* `ng-switch-default`

If you handle each case in this logic, I should be good to go for this PR.

parlarjb: `ng-if` is handled by the `$destroy`. I'll look into the `switch` directives

parlarjb: The `switch` directives seem to "just work" because they appear to be using `ng-if` internally (or, at least, they're completely removing elements from the DOM, not just showing/hiding them).

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/856/gallery.html)

CITguy:
DEV - LGTM

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/856/gallery.html)

PR #858, opened 2015-03-20
style(progressbar): update styles and docs/demo for progress bar

glynnis:
@ElementE and I did some style work to make progress bars a little more Encore-y in appearance. In updating CSS we realized that we also needed to add progress bars to the style guide. This PR does both.

### DIFF
--- a/src/progressbar/progressbar.less
+++ b/src/progressbar/progressbar.less
@@ -1,83 +1,74 @@
 @import 'vars';
-//
-// Progress bars
-// --------------------------------------------------
-
-.progress-bar-variant(@color) {
-  background-color: @color;
-
-  // Deprecated parent class requirement as of v3.2.0
-  .progress-striped & {
-    #gradient > .striped();
-  }
-}
 
 //** Background color of the whole progress component
-@progress-bg:                 #f5f5f5;
-//** Progress bar text color
-@progress-bar-color:          #fff;
-//** Variable for setting rounded corners on progress bar.
-@progress-border-radius:      4px;
+@progress-bg: #f5f5f5;
 
 //** Default progress bar color
-@progress-bar-bg:             @siteBrandingBg;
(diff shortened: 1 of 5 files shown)


glynnis:
## Before:
screen shot 2015-03-20 at 3 04 35 pm
screen shot 2015-03-20 at 3 04 22 pm

## After:
screen shot 2015-03-20 at 3 03 50 pm
screen shot 2015-03-20 at 3 04 03 pm
screen shot 2015-03-20 at 3 04 10 pm

glynnis:
retest this please

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/858/gallery.html)

parlarjb:
LGTM

src/progressbar/progressbar.less
```diff
@@ -97,92 +88,86 @@
...
+    .animation(progress-bar-stripes 2s linear infinite);
 }
 
-
 // Variations
 // -------------------------
 
-.progress-bar-success {
-  .progress-bar-variant(@progress-bar-success-bg);
+.progress-bar-secondary {
```
megalord: It seems that `.progress-bar-secondary` and `.progress-bar-tertiary` can be used to change the background color, but the styleguide implies that all progress bars should have the same blue background.  If these classes are still relevant, should they be documented?

I think they may be relevant in the context of the `type` attribute of the directive, but that is not being used in the demo.

glynnis: @megalord Good catch. I left them in because they'll be relevant for stacked progress bars, which I'm planning to document as part of a separate PR (it involves some design overhead to pick colors, and some technical implementation to get tooltips working for progress bars).

megalord:
LGTM

PR #871, opened 2015-03-30
fix(metadata): correct/update styling

CITguy:
* remove `float` from `<dd>`
* move padding from `<dd>` to `<dt>` margin
* update styling based on [EOD-643](https://jira.rax.io/browse/EOD-643)

**POTENTIAL BREAKING CHANGE:**
* removed `.list-definition` class and consolidated to using just `.metadata`

Closes #870

PR #872, opened 2015-04-01
feat(layout): Update metadata style to left align dt elements

julianbonilla:
Before

before

After

after

/cc @glynnis @ElementE

### DIFF
--- a/src/metadata/metadata.less
+++ b/src/metadata/metadata.less
@@ -6,7 +6,6 @@
     dt {
         line-height: 1.6em;
         float: left;
-        min-width: 9em;
         font-weight: 700;
         text-align: left;
         clear: left;
--- a/src/metadata/metadata.less
+++ b/src/metadata/metadata.less
@@ -8,7 +8,7 @@
         float: left;
         min-width: 9em;
         font-weight: 700;
-        text-align: right;
+        text-align: left;
         clear: left;
     }
     dd {


src/metadata/metadata.less
```diff
@@ -8,7 +8,7 @@
         float: left;
         min-width: 9em;
```
glynnis: I think we should remove the min-width line -- it causes weird spacing and wrapping especially at smaller browser widths.

glynnis:
FWIW, I'm going to submit a separate PR to remove that vertical rule from this layout -- that's not a style we're really using anymore and it looks broken now that we've changed up metadata styles. Not part of this work, though.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/872/gallery.html)

CITguy:
If we want to keep the aligned values as shown in the screenshots, we might want some visual cue to aid the user's eyes across to the value in the case of short labels.  

Maybe something like this:
screen shot 2015-04-02 at 10 15 07 am

@glynnis, thoughts?

CITguy:
* The `min-width` property is how we determine the actual width of the `<dt>` (for reasons unknown). 
  * @julianbonilla: `min-width` is still being set on [Line 34](https://github.com/rackerlabs/encore-ui/pull/872/files#diff-efed70219550db03635e3049f53a8a4eR34), btw.
* Do we know why we're splitting the CSS definition among two classes (`.list-definition` and `.metadata`)? @glynnis, @parlarjb 
  * Can we just consolidate the styles into `.metadata`?

glynnis:
@CITguy Turns out the way Billing implemented this change didn't correspond exactly with what we designed :grin: 

This is what we actually want:
screen shot 2015-04-02 at 3 08 30 pm

As you indicated, having too much space between key and value makes them hard for the eye to match them with each other. Having each column left-aligned (as you have above) also means that wrapping lines will get weird for responsive treatment.

@ElementE and I conferred on this, and we agree that just having each pair left aligned (rather than keys left aligned in one column and values left aligned in another) would be best, with a small space between each key and value pair (like the screenshot I attached above).

@CITguy I'm not sure that there's a reason the definition is split, unless we use the `.list-definition` class somewhere else for non-metadata styles. Possibly it's just an artifact of old styles that didn't get refactored very well.

CITguy:
We can probably close this PR. I've updated #871 to include the left-aligned definitions.

**Chrome**
screen shot 2015-04-03 at 10 25 16 am

**Firefox**
screen shot 2015-04-03 at 10 25 42 am

julianbonilla:
These style updates were pulled into #871.

PR #879, opened 2015-04-03
feat(rxCharacterCount): turn highlighing off by default

megalord:
This is a quick fix to add the `highlight` attribute to `rx-character-count`.  It is not intended to address any of the known issues with the highlighting, but rather to get the general component to a better state.

### DIFF
--- a/src/rxCharacterCount/README.md
+++ b/src/rxCharacterCount/README.md
@@ -12,7 +12,7 @@ The 254 and 10 values are both configurable. To change the maximum number of cha
 By default, any text field using `ng-model` has `ng-trim="true"` applied to it. This means that any leading and trailing spaces/blanks in your text field will be ignored. They will not count towards the remaining character count. If you want it to count leading/trailing spaces, then just add `ng-trim="false"` to your `<textarea>`.
 
 ### Styling ###
-When specifying a width other than the default, you should style some built-in classes in addition to the text field itself. As in the demo, the `.input-highlighting` class should have the same width as the text field, and the `.character-count-wrapper` should be used to correctly position the counter.
+When specifying a width other than the default, you should style some built-in classes in addition to the text field itself. As in the demo, the `.input-highlighting` class should have the same width as the text field (if highlighting is used), and the `.counted-input-wrapper` should be used to correctly position the counter.
 
 ### ngShow/ngHide/ngIf/ngSwitch/etc. ###
 If you wish to show/hide your `textarea` element, we recommend placing the element inside of a `<div>` or `<span>`, and doing the `ng-show` / `ng-hide` /etc. on that `div` / `span`. For example, 
@@ -30,3 +30,10 @@ We _do_ have preliminary support for putting these directives directly inside th
 ```
 
 But this support should be considered experimental. If you choose to take advantage of it, please ensure you've extensively tested that it performs correctly for your uses.
+
+### Highlighting ###
+Characters that are over the limit will be highlighted in red if the `highlight="true"` attribute is on the directive's element. Because this functionality is currently unstable, it has been left off by default. Please test your use case heavily before shipping with this feature enabled.
+
+Known failure cases:
+* Content that causes a scrollbar in the textarea
+* Initial text (coming from the model) that is over the limit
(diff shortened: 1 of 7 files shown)


src/rxCharacterCount/rxCharacterCount.exercise.js
```diff
@@ -69,9 +70,11 @@ exports.rxCharacterCount = function (options) {
             expect(component.isOverLimit()).to.eventually.be.false;
         });
 
-        it('should not show any highlights on an empty text box', function () {
-            expect(component.overLimitText).to.eventually.equal('');
-        });
+        if (options.highlight) {
```
megalord: This conditional is a bit repetitive, but since the tests are ordered, I am not aware of a better way to optionally run some of the tests.  Any ideas?

src/rxCharacterCount/README.md
```diff
@@ -12,7 +12,7 @@ The 254 and 10 values are both configurable. To change the maximum number of cha
 By default, any text field using `ng-model` has `ng-trim="true"` applied to it. This means that any leading and trailing spaces/blanks in your text field will be ignored. They will not count towards the remaining character count. If you want it to count leading/trailing spaces, then just add `ng-trim="false"` to your `<textarea>`.
 
 ### Styling ###
-When specifying a width other than the default, you should style some built-in classes in addition to the text field itself. As in the demo, the `.input-highlighting` class should have the same width as the text field, and the `.character-count-wrapper` should be used to correctly position the counter.
+When specifying a width other than the default, you should style some built-in classes in addition to the text field itself. As in the demo, the `.input-highlighting` class should have the same width as the text field (if highlighting is used), and the `.character-count-wrapper` should be used to correctly position the counter.
```
parlarjb: This one is totally my fault, but in the last release we got rid of `.character-count-wrapper`, and are using `.counted-input-wrapper`. Would you mind updating this here?

src/rxCharacterCount/README.md
```diff
@@ -30,3 +30,6 @@ We _do_ have preliminary support for putting these directives directly inside th
 ```
 
 But this support should be considered experimental. If you choose to take advantage of it, please ensure you've extensively tested that it performs correctly for your uses.
+
+### Highlighting ###
+Characters that are over the limit will be highlighted in red if the `highlight="true"` attribute is on the directive's element. Because this functionality is currently unstable, it has been left off by default.
```
parlarjb: I think for now, we should point out the known failure cases, and emphasize that devs should heavily test before shipping with it turned on

src/rxCharacterCount/rxCharacterCount.exercise.js
parlarjb: I can't think of anything either. @Droogans any thoughts?

Droogans: I think I can infer what `options.highlight` means, but it would be nice if it were [added to the exercise docs above](https://github.com/megalord/encore-ui/blob/fa47438d548663375edd23b6cd04be3f2bf0425d/src/rxCharacterCount/rxCharacterCount.exercise.js#L5-L12).

As far as ordering the tests go, you may have options. I tend to order them so that performance is a top priority, seeing how this may be called many times in many places. But if there is very little set up (as it may be the case here with simple text addition/deletion calls), it may be worth breaking it out.

```js
if (options.highlight) {
    describe('highlighting', function () {
        // ...
    });
}
```

src/rxCharacterCount/README.md
megalord: will do

src/rxCharacterCount/rxCharacterCount.exercise.js
megalord: My bad, I'll update the docs.  By breaking it out, we will just trade one duplication for another, because now the parts of the test that generate the text string and write it will be in the highlighting part too.  However, adding further tests down the road favors your approach, so I think that is a better solution.

CITguy: You should be able to do something like the following to prevent the duplicated setup.

```javascript
describe('rxCharacterCount', function () {
  // common setup logic

  describe('default functionality', function () {
    // all tests not pertaining to highlighting
  });
  describe('with highlighting', function () {
    // all tests pertaining to highlighting
  });
});


```

parlarjb:
LGTM

src/rxCharacterCount/rxCharacterCount.js
```diff
@@ -90,7 +86,13 @@ angular.module('encore.ui.rxCharacterCount', [])
                 scope.$apply();
             }
 
-            element.on('input', writeLimitText);
+            if (attrs.highlight === 'true') {
```
parlarjb: It's my fault they're missing, but since we're adding a new attribute, could you add `jsdocs` parameters comments for this directive?

megalord: :+1:

PR #880, opened 2015-04-03
fix(rxBreadcrumbs): Fix rxBreadcrumbs.visit();

jj4th:
Note that there were no midways or specs that test visit.  The click function does not bubble, so we need to actually click the link itself.

### DIFF
--- a/src/rxBreadcrumbs/docs/rxBreadcrumbs.midway.js
+++ b/src/rxBreadcrumbs/docs/rxBreadcrumbs.midway.js
@@ -115,6 +115,14 @@ describe('rxBreadcrumbs', function () {
             expect(middle.href).to.eventually.equal(browser.baseUrl + '/');
         });
 
+        it('should visit the correct page when clicking on the breadcrumb', function () {
+            var homeHref = browser.baseUrl + '/#/overview';
+
+            middle.visit();
+            expect(browser.getCurrentUrl()).to.eventually.equal(homeHref);
+        });
+        // Note that after this test, we are now at the /#/overview page
+
     });
 
     describe('default breadcrumbs', function () {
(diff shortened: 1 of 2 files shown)


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/880/gallery.html)

CITguy:
LGTM

parlarjb:
LGTM

src/rxBreadcrumbs/docs/rxBreadcrumbs.midway.js
```diff
@@ -115,6 +115,14 @@ describe('rxBreadcrumbs', function () {
...
         });
 
+        it('should visit the correct page when clicking on the breadcrumb', function () {
+            var homeHref = browser.baseUrl + '/#/overview';
+
+            middle.visit();
+            expect(browser.getCurrentUrl()).to.eventually.equal(homeHref);
+        });
+        // Note that after this test, we are now at the /#/overview page
+
```
Droogans: Please add an `after` function call to put the url back. Although the next describe block returns back using `before`, I'd rather see tests clean up after themselves between each describe block, rather than relying on the ordering of the tests to work correctly.

jj4th: Ah, so, this was somewhat intentional, as the philosophy in Cloud UI is that every describe block must be able to execute in isolation.  An additional page transition at the end just adds an unnecessary delay of approximately 10-15 seconds.  As this is not Cloud UI, I'll go ahead and add the after block.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/880/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/880/gallery.html)

PR #881, opened 2015-04-03
Add rx-toggle-switch to framework.

### LINKED ISSUE #860: Bring rxSwitch into the framework, add to docs
Currently the directive is in Billing.

screen shot 2015-03-23 at 11 13 23 am

megalord:
Fixes #860 

screen shot 2015-04-03 at 10 19 35 am

### DIFF
--- a/src/rxToggleSwitch/docs/rxToggleSwitch.html
+++ b/src/rxToggleSwitch/docs/rxToggleSwitch.html
@@ -1,0 +1,29 @@
+<!-- Sample HTML goes here as a live example of how the component can be used -->
+<div ng-controller="rxToggleSwitchCtrl">
+    <p>
+        <h3>Simple use</h3>
+        <rx-toggle-switch class="demo-default-values" ng-model="encoreSuperpowers"></rx-toggle-switch>
+        <span>Do you have powers? {{ encoreSuperpowers }}</span>
+    </p>
+    <p>
+        <h3>Specific model values</h3>
+        <rx-toggle-switch class="demo-model-values"
+            ng-model="toggle2"
+            true-value="active"
+            false-value="disabled">
+        </rx-toggle-switch>
+        <span>Model value: {{ toggle2 }}</span>
+    </p>
+    <p>
+        <h3>Performing an asynchronous operation</h3>
+        <rx-toggle-switch class="demo-post-hook"
+            ng-model="toggle3"
+            post-hook="attemptChange(value)"
+            disabled="loading">
+        </rx-toggle-switch>
+    </p>
+    <p>
+        <h3>Disabled</h3>
+        <rx-toggle-switch class="demo-disabled" ng-model="toggle4" disabled="true"></rx-toggle-switch>
+    </p>
+</div>
--- a/src/rxToggleSwitch/docs/rxToggleSwitch.js
+++ b/src/rxToggleSwitch/docs/rxToggleSwitch.js
@@ -1,0 +1,23 @@
+/*jshint unused:false*/
+
+// This file is used to help build the 'demo' documentation page and should be updated with example code
+angular.module('demoApp')
+.controller('rxToggleSwitchCtrl', function ($scope, $timeout, rxNotify) {
+    $scope.toggle3 = true;
+
+    $scope.attemptChange = function (value) {
+        $scope.loading = true;
+        rxNotify.add('Saving...', {
+            loading: true
+        });
+
+        // Simulate an api request
+        $timeout(function () {
+            $scope.loading = false;
+            rxNotify.clear('page');
+            rxNotify.add('Change saved', {
+                type: 'success'
+            });
+        }, 1000);
+    };
+});
--- a/src/rxToggleSwitch/rxToggleSwitch.js
+++ b/src/rxToggleSwitch/rxToggleSwitch.js
@@ -22,8 +22,9 @@ angular.module('encore.ui.rxToggleSwitch', [])
     return {
         restrict: 'E',
         templateUrl: 'templates/rxToggleSwitch.html',
-        require: '^ngModel',
+        require: 'ngModel',
         scope: {
+            model: '=ngModel',
             disabled: '=?',
             postHook: '&',
             trueValue: '@',
@@ -32,6 +33,10 @@ angular.module('encore.ui.rxToggleSwitch', [])
         link: function (scope, element, attrs, ngModelCtrl) {
             var trueValue = _.isUndefined(scope.trueValue) ? true : scope.trueValue;
             var falseValue = _.isUndefined(scope.falseValue) ? false : scope.falseValue;
+
+            if (_.isUndefined(scope.model) || scope.model !== trueValue) {
+                scope.model = falseValue;
+            }
 
             ngModelCtrl.$formatters.push(function (value) {
                 return value === trueValue;
--- a/src/rxToggleSwitch/rxToggleSwitch.js
+++ b/src/rxToggleSwitch/rxToggleSwitch.js
@@ -1,0 +1,59 @@
+angular.module('encore.ui.rxToggleSwitch', [])
+/**
+ *
+ * @ngdoc directive
+ * @name encore.ui.rxToggleSwitch:rxToggleSwitch
+ * @restrict E
+ * @description
+ * Displays an on/off switch toggle
+ *
+ * @paran {string} [ng-model] The scope property to bind to
+ * @param {boolean} [disabled] Indicates if the input is disabled
+ * @param {function} [postHook] A function to run when the switch is toggled
+ * @param {expression} [trueValue=true] The value of the scope property when the switch is on
+ * @param {expression} [falseValue=false] The value of the scope property when the switch is off
+ *
+ * @example
+ * <pre>
+ *     <rx-toggle-switch ng-model="foo"></rx-toggle-switch>
+ * </pre>
+ */
+.directive('rxToggleSwitch', function () {
+    return {
+        restrict: 'E',
+        templateUrl: 'templates/rxToggleSwitch.html',
+        require: '^ngModel',
+        scope: {
+            disabled: '=?',
+            postHook: '&',
+            trueValue: '@',
+            falseValue: '@'
+        },
+        link: function (scope, element, attrs, ngModelCtrl) {
+            var trueValue = _.isUndefined(scope.trueValue) ? true : scope.trueValue;
+            var falseValue = _.isUndefined(scope.falseValue) ? false : scope.falseValue;
+
+            ngModelCtrl.$formatters.push(function (value) {
+                return value === trueValue;
+            });
+
+            ngModelCtrl.$parsers.push(function (value) {
+                return value ? trueValue : falseValue;
+            });
+
+            ngModelCtrl.$render = function () {
+                scope.state = ngModelCtrl.$viewValue ? 'ON' : 'OFF';
+            };
+
+            scope.update = function () {
+                if (scope.disabled) {
+                    return;
+                }
+
+                ngModelCtrl.$setViewValue(!ngModelCtrl.$viewValue);
+                ngModelCtrl.$render();
+                scope.postHook({ value: ngModelCtrl.$modelValue });
+            };
+        }
+    };
+});
--- a/src/rxToggleSwitch/rxToggleSwitch.less
+++ b/src/rxToggleSwitch/rxToggleSwitch.less
@@ -55,11 +55,12 @@
         .knob {
             left: 100%;
             -webkit-transform: translate(-100%, 0);
-                    transform: translate(-100%, 0);
+            -moz-transform: translate(-100%, 0);
+            transform: translate(-100%, 0);
         }
     }
     .knob {
-        transition: all 0.1s ease-out;
+        .transition-value(all 0.1s ease-out);
 
         position: absolute;
         display: inline-block;
--- a/src/rxToggleSwitch/rxToggleSwitch.less
+++ b/src/rxToggleSwitch/rxToggleSwitch.less
@@ -1,0 +1,79 @@
+/*
+ * rxToggleSwitch
+ */
+.rx-toggle-switch {
+
+    @textColor: #fff;
+    @offStateColor: #b3b3b3;
+    @borderColor: darken(@offStateColor, 10%);
+    @onStateColor: #00ac31;
+    @knobColor: #fff;
+    @knobBorderColor: darken(@knobColor, 10%);
+    @knobSize: 1.2em;
+    @edgeSpacing: @knobSize * 0.9;
+    @lineHeight: @knobSize * 1.25;
+    @blurDist: @edgeSpacing;
+    @blurSpread: (@edgeSpacing/3) * -1;
+
+    font-size: 0.8em;
+    line-height: @lineHeight;
+
+    display: inline-block;
+
+    border: 1px @borderColor solid;
+    border-radius: 9px;
+    background-color: @offStateColor;
+    color: @textColor;
+
+    padding: 0 @edgeSpacing 0 @edgeSpacing;
+
+    width: 5em;
+
+    cursor: pointer;
+    position: relative;
+    text-align: right;
+
+    .box-shadow(inset 1px 2px @blurDist @blurSpread rgba(0,0,0,0.50));
+    .user-select(none);
+    .box-sizing(border-box);
+
+    &[disabled] {
+        background-color: desaturate(lighten(@offStateColor, 7%), 80%);
+        color: #666666;
+        &.on {
+            background-color: desaturate(lighten(@onStateColor, 33%), 80%);
+        }
+        cursor: not-allowed;
+        .knob {
+            background-color: #adadad;
+            border: 1px #999999 solid;
+        }
+    }
+    &.on {
+        background-color: @onStateColor;
+        text-align: left;
+        .knob {
+            left: 100%;
+            -webkit-transform: translate(-100%, 0);
+                    transform: translate(-100%, 0);
+        }
+    }
+    .knob {
+        transition: all 0.1s ease-out;
+
+        position: absolute;
+        display: inline-block;
+        top: 0;
+        left: 0;
+
+        width: @knobSize;
+        height: @knobSize;
+        border-radius: (@knobSize/2)+1;
+        background-color: @knobColor;
+        border: 1px @knobBorderColor solid;
+
+        .box-shadow(0 0 @blurDist @blurSpread rgba(0,0,0,0.55));
+        overflow: hidden;
+    }
+
+}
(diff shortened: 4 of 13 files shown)


CITguy:
I feel like it should it be called `rx-toggle-switch`.  My first reaction to seeing `rx-switch` was that it is an angular implementation of a `switch...case` structure. This is obviously not the case, but `rx-toggle-switch` seems more semantically correct.

glynnis:
@megalord Can you also add a demo to the style guide in the form section? (http://rackerlabs.github.io/encore-ui/#/styleguide/forms)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

halkon:
I totally :+1: `rx-toggle-switch` the `rx-switch` name was never really liked and I think `rx-toggle-switch` was preferred for when importing into the framework.

megalord:
I agree, `rx-toggle-switch` is a better name; I'll make the change.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

src/rxToggleSwitch/rxToggleSwitch.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+.directive('rxToggleSwitch', function ($timeout) {
+    return {
+        restrict: 'E',
+        templateUrl: 'templates/rxToggleSwitch.html',
+        scope: {
+            model: '=',
+            loading: '=?',
+            disabled: '=?',
+            postHook: '&',
```
parlarjb: This is optional, so let's add the `?`

src/rxToggleSwitch/rxToggleSwitch.js
```diff
@@ -0,0 +1,68 @@
...
+                return value === getTrueValue();
+            };
+
+            scope.update = function (value) {
+                if (scope.disabled) {
+                    return;
+                }
+                var newValue = switchValue(value);
+
+                scope.model = newValue;
```
parlarjb: Instead of having a `.model` attribute, I'd love to just use `ng-model` and the `ngModelController`. With `$setViewValue` and `$render()`, we should be able to get what we'd need, and you'd be able to remove the `model` from the template.

src/rxToggleSwitch/rxToggleSwitch.js
halkon: Optional I believe only matters for 2 way binding.  
For expressions or attributes, it will just ignore the attribute if not passed (instead of crashing like in the case of a 2-way binding)

https://docs.angularjs.org/api/ng/service/$compile#-scope-

src/rxToggleSwitch/rxToggleSwitch.js
```diff
@@ -0,0 +1,68 @@
...
+        templateUrl: 'templates/rxToggleSwitch.html',
+        scope: {
+            model: '=',
+            loading: '=?',
+            disabled: '=?',
+            postHook: '&',
+            trueValue: '@?',
+            falseValue: '@?',
+            trueText: '@?',
+            falseText: '@?'
```
halkon: Given the comment above... We could take off the `?` off of these

parlarjb:
@glynnis are you ok with the spacing between the right of the knob and the left of the text, when it's in an "off" position?

glynnis:
@parlarjb I think there needs to be more spacing there.

Also from a design perspective, I think having a switch that allows for text other than "on/off" seems ripe for misuse (the one that says "Disabled").

The point of a switch is: it's on or off, or cannot be changed. Disabling is communicated by its appearance/color only, not by text in the switch. If the visual treatment for a disabled switch doesn't communicate that clearly enough, I'd rather work on the visual design than allow for text other than "On" and "Off". Opening it up so that devs can put any text there opens up the possibility that someone would use a switch in an incorrect context, like a scenario where a set of radio buttons is more appropriate.

So, basically: can we remove the example and code that allows for the switch that says "Disabled"? 

@megalord let me know if you want me to tweak CSS for spacing and/or work on the visual cues for when a switch is disabled. Happy to let you handle it and post a new screen shot, but I can jump in if you need me.

glynnis:
To frame the problem of a switch with text via a silly example:
If you have a light switch in your house, you expect it to control one thing, and to have a state of "on" or "off". Having text on a switch in the UI is kind of like having a light switch that in an "on" position turns on your kitchen light, and in an "off" position turns your dryer on "tumble". 

By allowing text in a switch, we leave ourselves open to the possibility of switches that are doing complex actions that are not true on/off situations, which dilutes the meaning and use of a switch, and would be confusing to the user.

Hope that makes sense.

src/rxToggleSwitch/rxToggleSwitch.js
megalord: agreed

src/rxToggleSwitch/rxToggleSwitch.js
megalord: Good idea! I'll give it a shot.

megalord:
Yeah, removing the ability to set the text makes sense.  I also updated the style of the component and added it to the styleguide as suggested.
screen shot 2015-04-07 at 3 28 20 pm
Let me know what you think. :smile:

halkon:
That's actually a very valid point, and it may have been over-engineered from get go (last year).  The only purpose of being able to customize the text was so that you could change between:

Yes/No
On/Off
True/False

And have the flexibility of doing so (Enable/Disable would fit this obviously).  

But it can get confusing if on every page every switch changes it's true/false visual representation.   Thanks for the update @megalord

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

halkon:
I R DONE with DIS PR

LGTM

LGTM

CITguy:
I'll defer to @glynnis on this, but there's not enough contrast between the gray text and gray background for the disabled switch.

glynnis:
@megalord and I are gonna sync up about some of the design concerns for this. Adding a "do not merge" for now until we get things rolling on that front.

glynnis:
Oops! Didn't mean to close this with my errant click. 

Played around with colors to try to better distinguish disabled from active switches. Also improved the text contrast:

screen shot 2015-04-08 at 1 17 04 pm

Jordan has the color codes for these changes to implement. Still working on design improvements for the "spinner" switch, so leaving this as "do not merge".

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

CITguy:
@glynnis what would a disabled "On" look like?

megalord:
@CITguy this is what we settled on:
screen shot 2015-04-08 at 3 36 55 pm

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

megalord:
I removed the `loading` attribute thanks to @glynnis 's feedback.  I modified the demo to show how a toggle that fires an api call would work.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

glynnis:
Design: LGTM.

@halkon @CITguy Can you guys look this over again now that Jordan has completed changes?

halkon:
LGTM

:+1: :+1: :+1: :+1: :+1:

src/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -0,0 +1,92 @@
...
+
+    width: 5em;
+
+    cursor: pointer;
+    position: relative;
+    text-align: right;
+
+    -webkit-box-shadow: inset 1px 2px @blurDist @blurSpread rgba(0,0,0,0.50);
+       -moz-box-shadow: inset 1px 2px @blurDist @blurSpread rgba(0,0,0,0.50);
+            box-shadow: inset 1px 2px @blurDist @blurSpread rgba(0,0,0,0.50);
```
CITguy: You can use the `.box-shadow()` mixin for this.

src/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -0,0 +1,92 @@
...
+
+        width: @knobSize;
+        height: @knobSize;
+        border-radius: (@knobSize/2)+1;
+        background-color: @knobColor;
+        border: 1px @knobBorderColor solid;
+
+        -webkit-box-shadow: 0 0 @blurDist @blurSpread rgba(0,0,0,0.55);
+           -moz-box-shadow: 0 0 @blurDist @blurSpread rgba(0,0,0,0.55);
+                box-shadow: 0 0 @blurDist @blurSpread rgba(0,0,0,0.55);
```
CITguy: You can use the `.box-shadow()` mixin for this.

src/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -0,0 +1,92 @@
...
+ * rxSwitch
+ */
+.user-select(@select) {
+    -webkit-touch-callout: @select;
+      -webkit-user-select: @select;
+       -khtml-user-select: @select;
+         -moz-user-select: @select;
+          -ms-user-select: @select;
+              user-select: @select;
+}
```
CITguy: This should be in the `mixins.less` file. Additionally, you can define it simpler as:

```less
.user-select(...) {
  -prefixed-property: @arguments;
  ...
}
```

src/rxToggleSwitch/docs/rxToggleSwitch.js
```diff
@@ -0,0 +1,22 @@
+/*jshint unused:false*/
+
+// This file is used to help build the 'demo' documentation page and should be updated with example code
+function rxToggleSwitchCtrl ($scope, $timeout, rxNotify) {
```
CITguy: The component template should probably be updated to do this, but we should define the demo controllers using the `.controller('ctrlName', function () {...})` syntax for angular 1.3 compatibility.

src/rxToggleSwitch/rxToggleSwitch.less
megalord: Oooh good to know

src/rxToggleSwitch/rxToggleSwitch.less
megalord: Well, it only takes one argument, so I don't feel like a splat is necessary.  To me, a splat tells the reader that multiple arguments are expected.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

src/rxToggleSwitch/rxToggleSwitch.less
CITguy: I tend to use a splat `(...)` when I write a wrapper for vendor-prefixed properties to denote that it accepts the standard arguments, but the functionality is pretty much the same.  I just like keeping things consistent.

src/styles/mixins.less
```diff
@@ -74,3 +74,12 @@
...
 }
+
+.user-select(@select) {
+    -webkit-touch-callout: @select;
+    -webkit-user-select: @select;
+    -khtml-user-select: @select;
+    -moz-user-select: @select;
+    -ms-user-select: @select;
+    user-select: @select;
+}
```
CITguy: I tend to use a splat `(...)` when I write a wrapper for vendor-prefixed properties to denote that it accepts the standard arguments. However, to LESS, the functionality should be the same.

CITguy:
Have you rebased?

halkon:
Does he particularly need to? The PR is merge-able

CITguy:
@halkon: Yes.  I'm seeing an old version of the [`.border-radius()`](https://github.com/rackerlabs/encore-ui/blob/rx-switch/src/styles/mixins.less#L24) mixin that was changed in #888, so I wouldn't want to introduce any unforeseen regressions.

halkon:
I think git will handle it just fine, that being said, `.border-radius(...)` is definitely not being used in this PR

halkon:
@glynnis https://github.com/rackerlabs/encore-tq-ui/pull/275 is being referenced as using `Yes`/ `No`, mind you while implemented last year, the idea was to cover for these cases. I'll find out (given due time) the reasoning for using `Yes`/ `No` versus `On`/ `Off`, however not to disagree with the decision taken to eliminate the ability to define `On`/ `Off` values, just trying to, rather than cover all cases, make sure we're not missing any

halkon:
DOH! posted before some searching in a bigger PR: 

https://github.com/rackerlabs/encore-tq-ui/pull/266/files#diff-dea126b201803ec1dab674699f938bdbR19

Linguistically it would seem more proper to say that "A Feature" is "personal" as a `Yes`/ `No` versus a `On`/ `Off`

May be minimal, and it could probably just be a setting (much like the other PR mentioned) to just change from `On`/ `Off` to a more personal `Yes`/ `No`.

halkon:
@glynnis talked with @solim8r today and more or less it came up with, if we're not gonna allow custom text, why even show any text at all?  We could go full mobile mode in the fact that "Green" means on and "Gray" means off.

Throwing that comment out there, I think it would get rid of any linguistics/grammar issues and would likely just make this whole thing simpler.

glynnis:
@halkon I think it would be fine to offer an alternate "yes / no" version of the switch, as that keeps with the function and purpose of a switch, as long as the "yes / no" version is hard-coded and not the result of allowing folks to fill in whatever they want.

My problem with alternate text is with the idea of allowing consumers of the switch to add whatever text they want -- that's where it becomes problematic due to potential for long text breaking the design and/or folks filling in text that isn't true to the switch's function and purpose.

I do think it's important that we keep text present on the switch, as there's some general user research that suggests that not all users find switch design as intuitive as others, and the "on/off" or "yes/no" helps communicate its function more clearly.

halkon:
Oh I don't disagree with the custom text comment :grin:, I can definitely see the disadvantage on it, and unless we implement a dictionary in it to validate opposites... better to not.

That being said, offering a few other alternatives would be a good thing.  I think Yes/No would be the first, however, it can come in a later PR if need be.  

Interesting about the user research, it makes sense, I kind of was thinking about that this morning on my way in to the office.  "Is it really intuitive that `Green vs Gray` means on/off?".

I think even in essence `Green vs Red` might confuse people even if we deal with it on a day to day basis with stop lights.

@CITguy are we still waiting on the rebase?

megalord:
rebased and squashed

halkon:
:astonished: :scream: Merge at will!

src/rxToggleSwitch/docs/rxToggleSwitch.html
```diff
@@ -0,0 +1,29 @@
+<!-- Sample HTML goes here as a live example of how the component can be used -->
+<div ng-controller="rxToggleSwitchCtrl">
+    <p>
+        <h3>Simple use</h3>
+        <rx-toggle-switch class="demo-default-values" ng-model="encoreSuperpowers"></rx-toggle-switch>
+        <span>Do you have powers? {{ encoreSuperpowers }}</span>
```
parlarjb: Because `encoreSuperpowers` is not initially set on the scope, it goes to `undefined`, and nothing _initially_ shows up here.
screen shot 2015-04-13 at 5 17 30 pm

Add `$scope.encoreSuperpowers = false` to your `rxToggleSwitchCtrl`, and you'll get:
screen shot 2015-04-13 at 5 18 03 pm

src/rxToggleSwitch/docs/rxToggleSwitch.html
```diff
@@ -0,0 +1,29 @@
...
+    </p>
+    <p>
+        <h3>Specific model values</h3>
+        <rx-toggle-switch class="demo-model-values"
+            ng-model="toggle2"
+            true-value="active"
+            false-value="disabled">
+        </rx-toggle-switch>
+        <span>Model value: {{ toggle2 }}</span>
```
parlarjb: Same problem as above, see the screenshots there.

src/rxToggleSwitch/docs/rxToggleSwitch.html
megalord: Oh, good point, I guess model needs to be initialized with `false-value` (if not yet set).  This could be a source of confusion, though perhaps effectively beatable with good docs.

parlarjb: Yeah, that's probably the right way to do it.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

parlarjb:
LGTM

src/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -0,0 +1,79 @@
...
+            background-color: #adadad;
+            border: 1px #999999 solid;
+        }
+    }
+    &.on {
+        background-color: @onStateColor;
+        text-align: left;
+        .knob {
+            left: 100%;
+            -webkit-transform: translate(-100%, 0);
```
TheNando: -webkit- but no -moz-?

src/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -0,0 +1,79 @@
...
+                    transform: translate(-100%, 0);
+        }
+    }
+    .knob {
+        transition: all 0.1s ease-out;
```
TheNando: Likewise, if you include support for older browsers with the -browser- selectors, you should include it here as well.

TheNando:
I also assess the looks favorably.

src/rxToggleSwitch/rxToggleSwitch.less
megalord: Ah, good catch.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/881/gallery.html)

PR #882, opened 2015-04-03
fix(spelling): Managed was mispelled

jj4th:


CITguy:
@jj4th, this PR is unnecessary. I've addressed the spelling error in #871

jj4th:
@CITguy You only changed it in one of the two files.

CITguy:
Ah, I was unaware of the other file.

CITguy:
but now I'm curious as to why we have the same demo code in two places...

jj4th:
Oh nice, dependency failure.

jj4th:
retest this please

parlarjb:
@CITguy The duplication is a side-effect of how are docs are currently split up. We previously _only_ had the component docs, not the style guide. Now we have the question of "what should go in the style guide, what should be with the components?" Eventually they'll be combined into something more cohesive.

parlarjb:
LGTM

PR #886, opened 2015-04-06
feat(rxApp): cache nav in local storage

### LINKED ISSUE #515: Store menu data in LocalStorage for quicker load times
After the first download of the menu json, store the data so that it can be instantly retrieved on page load. The normal CDN retrieval happens behind the scenes and the nav will be updated after load. 

Benefits:
- Can dev locally w/o waiting on CDN
- Avoid blank menu while data is loading

megalord:
Fixes #515 

With this change, the nav is initially populated with a cached copy (if available).  If the request to the CDN succeeds, it will overwrite the cached version (both on the page and in the cache).  If the request fails, it will do so silently, and the cached version will be used.

### DIFF
--- a/src/rxApp/rxApp.js
+++ b/src/rxApp/rxApp.js
@@ -1,5 +1,5 @@
 angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnvironment', 'ngSanitize',
-    'ngRoute', 'cfp.hotkeys', 'encore.ui.rxSession'])
+    'ngRoute', 'cfp.hotkeys', 'encore.ui.rxSession', 'encore.ui.rxLocalStorage'])
 /**
 * @ngdoc service
 * @name encore.ui.rxApp:encoreRoutes
@@ -10,7 +10,8 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
 * @returns {object} Instance of rxAppRoutes with `fetchRoutes` method added
 */
 .factory('encoreRoutes', function (rxAppRoutes, routesCdnPath, rxNotify, $q, $http,
-                                     rxVisibilityPathParams, rxVisibility, Environment, rxHideIfUkAccount) {
+                                   rxVisibilityPathParams, rxVisibility, Environment,
+                                   rxHideIfUkAccount, LocalStorage) {
 
     // We use rxVisibility in the nav menu at routesCdnPath, so ensure it's ready
     // before loading from the CDN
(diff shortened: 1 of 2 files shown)


halkon:
That's all I got

src/rxApp/rxApp.js
```diff
@@ -33,9 +34,23 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
     }
 
     encoreRoutes.fetchRoutes = function () {
-        return $http.get(url)
-            .success(encoreRoutes.setAll)
-            .error(setFailureMessage);
+        var routesKey = 'encoreRoutes-' + Environment.get().name;
```
parlarjb: We have to avoid `Environment.get().name`. It's irreparably broken. It'd be better for you to explicitly check against the environments we're interested in, `Environment.isLocal()`, etc. You might need to add an `Environment.isStaging()` to do this properly

megalord: Oh, good to know.  I shouldn't need an `isStaging()`, because I can make staging as the default (as is the case when determining the url).

parlarjb: There's a note about it at the bottom of the rxEnvironment docs, in case you're curious as to the reasons it's so bad http://rackerlabs.github.io/encore-ui/#/component/rxEnvironment

src/rxApp/rxApp.js
```diff
@@ -25,17 +26,47 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
-        return $http.get(url)
-            .success(encoreRoutes.setAll)
-            .error(setFailureMessage);
+        var routesKey = 'encoreRoutes-' + suffix;
+        var cachedRoutes = LocalStorage.getObject(routesKey);
+
+        $http.get(url)
+            .success(function (routes) {
+                encoreRoutes.setAll(routes);
+                LocalStorage.setObject(routesKey, routes);
```
parlarjb: I think you forgot to do `cachedRoutes = routes` here. Because of the return statement of this function, you'll probably need this.

megalord: Actually, I'm not sure that is necessary, since the `success` handler will run after `fetchRoutes()` has already returned.  The only way the handler would run first is if the promise is already resolved, but that should not happen for http requests.

parlarjb: What's the goal of the `scope.routes = appRoutes.fetchRoutes();` assignment below then? It is just to say "If there was a cached value, here it is, otherwise an empty array means nothing was cached"?

My thought was that either way, it would represent the set of routes that `fetchRoutes()` ends up using.

megalord: Yes, you are exactly right; it is a *temporary* value for the routes.  The routes become fixed by `appRoutes.setAll()` just a few lines below.

It's impossible for `fetchRoutes()` to return the final value of the routes, since it does not know what that is.  As was the case before my changes, the function kicks off the api request and uses `setAll()` to resolve the internal promise indicating the routes are loaded.

parlarjb: d'oh, I should have expanded the old code to remind myself how the rest of it worked. Carry on :)

src/rxApp/rxApp.js
```diff
@@ -25,17 +26,47 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+            })
+            .error(function () {
+                if (cachedRoutes) {
+                    encoreRoutes.setAll(cachedRoutes);
```
parlarjb: @glynnis This code here says "If we fail to grab the nav from the CDN, and we have a cached version, then use the cached version". Should we be notifying the user that the CDN failed, and that they're using something from cache? Or are you ok with silently loading up the cached values?

parlarjb:
LGTM

src/rxApp/rxApp.js
glynnis: I feel like silent loading is probably best, since for some uses an error message may cause confusion rather than provide helpful information.

That said, how often do we expect this will happen? And is there some kind of "check" we can do for users that may be experiencing this and report a bug? If loading a cached nav would happen with some frequency or lead to false bug reports of some kind, then showing some kind if error may be worth any confusion it might cause.

parlarjb: It should be very rare. I'll file an issue that maybe we (Encore devs) should be silently notified whenever this happens, or it's something we should look for in New Relic

glynnis: Ok, if we have notifications/tracking, then I stand by my initial statement: don't show an error, just silently load cache.

parlarjb: Maybe we should show the error for now, and remove it once we actually are tracking?

glynnis: That's fine too. Something relatively general like, "There was a problem loading navigation, so a cached version has been loaded for display." ??

Seems like it should get the orange/warning treatment.

halkon: So I'm gonna ask something and don't... you know... burn my house down but... And perhaps I should look more and see if it's been discussed or implemented.  But...

Can we immediately set/load the cached version of the menu when starting up while the real menu loads (and then replace?)

I know this may possibly at times cause some jumpiness or differences in the menu.  However.  it would provide a more instant experience rather than seeing an empty side bar (if it's taking a while).

halkon: Nevermind I tried to do a bit of reading and I think that's what @parlarjb and @megalord already spoke about above.

PR #887, opened 2015-04-07
feat(tabs): Update styles for tab component

julianbonilla:
* Add blue border around tab content area to provide visual context for tab
content areas (helpful especially for long screen that involve scrolling).
* Add small margin between tabs themselves so they are more visually distinct
from one another.

Before:

before

After: 

after

### DIFF
--- a/src/tabs/tabs.less
+++ b/src/tabs/tabs.less
@@ -41,6 +41,7 @@ Provides styling for the Angular-UI tab HTML
     li {
         .flex(0 0 auto);
         font-size: 14px;
+        margin-right: 1px;
     }
 
     a {
@@ -80,6 +81,9 @@ Provides styling for the Angular-UI tab HTML
     display: none;
     background: @white;
     padding: 10px;
+    border-left: 2px solid @tabBorder;
+    border-right: 2px solid @tabBorder;
+    border-bottom: 2px solid @tabBorder;
 
     &.active {
         display: block;


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/887/gallery.html)

src/tabs/tabs.less
```diff
@@ -79,7 +80,10 @@ Provides styling for the Angular-UI tab HTML
     clear: both;
     display: none;
     background: @white;
-    padding: 10px;
+    padding: 20px 20px 10px 20px;
```
CITguy: Is there a reason for the extra padding on three sides?

julianbonilla: To be consistent with Billing.

glynnis:
Design: LGTM

parlarjb:
LGTM

src/tabs/tabs.less
CITguy: I'm only concerned because Encore UI is a framework shared among multiple applications, not just Billing. It may only be an extra 10px on three sides, but for data-dense applications that are expecting smaller padding, this may wreak havoc on existing layouts. This is especially true for the left/right padding.

If Billing needs the extra padding, I feel like those style overrides should remain in Billing.

That's my 2 cents. @parlarjb, @glynnis any comments/concerns?

parlarjb: That's a good point, but I'll wait on feedback from @glynnis

glynnis: @julianbonilla Talked over this with @ElementE, and we agree with @CITguy -- keep the original padding. In cases where it makes sense to use more we can add on a project level, but that probably won't be necessary.

julianbonilla:
@parlarjb @glynnis @CITguy Removed the extra padding

glynnis:
Design: LGTM :balloon:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/887/gallery.html)

CITguy:
LGTM

PR #888, opened 2015-04-07
fix(styles): normalize .border-radius()

CITguy:
* correct implmentation of .border-radius() mixin
* replaced `border-radius` with mixin version


glynnis:
:clap: :clap: :clap: :clap: :clap: :clap: :clap: 

LGTM

PR #891, opened 2015-04-09
fix(exercises): Resolve require-loop

Droogans:
Since exercises need their underlying components to run, the
`exports.exercise = require('./exercise');` line needs to be below all
index.js components in order to prevent an infinite require loop from
occurring.


parlarjb:
LGTM

PR #892, opened 2015-04-09
test(rxPaginate): Handle one-page exercises

Droogans:
@parlarjb (or anyone for that matter) if you can think of a good way to catch more of these exercise edge cases without making the demo site unreadable with dozens of components all over the place, I'd appreciate it.

This one only shows up when you have a pagination element with one page of data on it being exercised.


parlarjb:
LGTM

PR #897, opened 2015-04-15
feat(rxMisc): Added rxAutoSave service

### LINKED ISSUE #679: Store in-progress form/input data
If a user is entering data in a text area, or other `input` elements, we should be able to temporarily save that data. This would allow them to navigate away and come back, and still have the content there.

Github and Stackoverflow both do this for their comment boxes, but they both send all data to their servers, and repopulate it on a return to the page.

I'd rather use a localstorage/sessionstorage solution, as described here http://stackoverflow.com/questions/14420624/how-to-keep-changed-form-content-when-leaving-and-going-back-to-https-page-wor

parlarjb:
Closes #679

This provides a way to have form data automatically saved to and
loaded from LocalStorage. This is to improve the user experience
if they're typing in a form and accidentally navigate away.

### DIFF
--- a/src/rxMisc/README.md
+++ b/src/rxMisc/README.md
@@ -2,3 +2,196 @@ A module for shared functionality across framework components.
 
 ## rxDOMHelper
 A small set of functions to provide some functionality that isn't present in Angular's jQuery-lite, and other useful DOM-related functions.
+
+
+## rxAutoSave
+`rxAutoSave` provides a way to store values in a form for later. For instance, if a user is entering values into a form, then accidentally navigate to a new page, we likely want the values to be present again when they click the "Back" button in their browser. By correctly setting up an `rxAutoSave` instance for the form, this can happen automatically. By default, all saved values will be cleared after two days.
+
+`rxAutoSave` is a service intended to be used in controllers. No directives are provided. The intent is that the HTML forms themselves will have no knowledge that their values are being saved. `rxAutoSave` operates by doing a `$watch` on the model values for a given form, storing those model values whenever they change, and loading them on instantation.
+
+The stored data is keyed on the page URL. This means you can track the form state for multiple pages simultaneously. For example, say you have an "Edit" form. The user has gone to edit some values for "Server1", at `"/servers/server1/edit"`, and for "Server2" at `"/servers/server2/edit"`. The edit progress for both servers will be saved independently of each other. `rxAutoSave` will also let you independently store values for multiple forms appearing on the same page.
+
+By default, all values are stored in the browser's `LocalStorage`. This means that if a user logs into a different computer, their stored values will not be present. Use of `SessionStorage` is also supported out-of-the-box. If you wish to save form states elsewhere (for instance, to an API), see the "Storage Location" section below. 
+
+### Setting up your template
+
+Nothing explicit needs to be done in your templates to add support for `rxAutoSave`. The only requirement is that all the `ng-model` values in a given form are stored within one object (`formData` below). For example, say you have the following form in your template:
+
+```
+    <form name="demoForm">
+        <rx-form-item label="A checkbox field!">
+            <input type="checkbox" ng-model="formData.checkbox" />
+        </rx-form-item>
+
+        <rx-form-item label="Name">
+            <input type="text" ng-model="formData.name"/>
+        </rx-form-item>
+
+        <rx-form-item label="Description">
+            <textarea rows="10" cols="50" ng-model="formData.description"></textarea>
+        </rx-form-item>
+
+    </form>
+```
+
+Note that all the models for the form are attributes of the `formData` scope variable.
+
+### Setting up your controller
+
+In your controller, you would have something like this in your initialization:
+
+```
+    $scope.formData = {
+        checkbox: false,
+        name: '',
+        description: ''
+    };
+```
+
+By default, every time this page was loaded, the form would be initialized with an unchecked checkbox, a blank `Name` field and a blank `Description`.
+
+To have `rxAutoSave` automatically save values, first inject `rxAutoSave` into your controller, and modify initialization as follows:
+
+```
+    $scope.formData = {
+        checkbox: false,
+        name: '',
+        description: ''
+    };
+
+    var autosave = rxAutoSave($scope, 'formData');
+```
+
+And that's it! Your `rxAutoSave` instance will watch for any change to `$scope.formData`, and will automatically write those changes to `LocalStorage`. 
+
+A third argument can be passed to `rxAutoSave`, specifying usage options. The default values for these options are:
+
+```
+    var autosave = rxAutoSave($scope, 'formData', {
+      clearOnSuccess: null,        // Promise
+      ttl: 172800,                 // Integer (seconds) - two days default
+      load: true,                  // Boolean or Promise that will resolve with a Boolean
+      save: true,                  // Boolean or Promise that will resolve with a Boolean
+      exclude: [],                 // Array<String>
+      storageBackend: LocalStorage // Object
+    });
+```
+
+All of these options will be described below.
+
+
+### Multiple Forms on one page
+
+`rxAutoSave` supports independently saving multiple forms on one page. To do this, have each form's model in its own object, and create individual `rxAutoSave` instances for each. i.e.:
+
+```
+    $scope.form1Data = {
+        checkbox: false,
+        name: '',
+        description: ''
+    };
+
+    $scope.form2Data = {
+        customerName: '',
+        birthday: ''
+    };
+
+    var autosave1 = rxAutoSave($scope, 'form1Data');
+    var autosave2 = rxAutoSave($scope, 'form2Data');
+```
+
+
+### Clearing values
+If you need to clear the stored values, you can call `autosave.clear()`. This will clear the values from `LocalStorage`, but won't affect your `$scope.formData` values.
+
+More likely, rather than manually calling `autosave.clear()`, you'd like the values to be cleared on a "successful submit". For example, if your user is editing the form described above, and they click a "Submit" button to send the values to a server, `LocalStorage` should be cleared for this form if the server call is a success.
+
+To do this, pass an "options" parameter as the third argument to `rxAutoSave`, setting a promise on the `clearOnSuccess` attribute, i.e.
+
+```
+    var autosave = rxAutoSave($scope, 'formData', { clearOnSuccess: serverSubmitPromise });
+```
+
+If the `serverSubmitPromise` resolves, then `rxAutoSave` will automatically clear the stored values for `formData` on this page.
+
+When instantiating your controller, there's a good chance that the `clearOnSuccess` promise you're interested in doesn't actually exist yet, i.e. if you want to clear on a successfull submit, you need the submit `promise`. Instances of `rxAutoSave` provide a `clearOnSuccess()` method to accept this promise after instantiation:
+
+```
+    var autosave = rxAutoSave($scope, 'formData');
+    
+    // Take some other actions
+    ...
+
+    $scope.onSubmit = function () {
+        // Server.save() is some $resource that returns a promise
+        var promise = Server.save($scope.formData);
+        autosave.clearOnSuccess(promise);
+    }
+
+```
+
+### Automatic expiry
+Another way to automatically clear values is to set an explict Time-To-Live (TTL) when instantiating your `rxAutoSave` instance. This is done with the `ttl` property of the `opts` object,
+
+```
+    // Automatically expire after 24 hours
+    var autosave = rxAutoSave($scope, 'formData', { ttl: 86400 });
+```
+
+By default, a `ttl` of `172800` (two days) is used.
+
+The `ttl` property takes a length of time in seconds. Whenever something in `formData` changes, the expiry time will be freshly set. With the example above, whenever `formData` is changed, the new expiry time will be set to 24 hours from the time of the change. In addition, we freshly set the expiry time whenever the data is loaded. If `formData` is 12 hours away from expiring, and the user visits the page again, then the expiry will be freshly set to a new 24 hours, whether or not the user makes a change.
+
+If a user visits a page after the data has expired, the data will be cleared from storage and not automatically loaded. (i.e. we're not running a continuous background process to look for expired data, we only check for expiration the next time `rxAutoSave` tries to load the data).
+
+To turn off automatic expiry for a given form, pass a value of `{ ttl: 0 }`. In this case, the data will never expire. You will have to clear it at an appropriate time by using one of the methods mentioned above.
+
+### Preventing automatic loading
+
+If you need to prevent `rxAutoSave` from automatically loading stored values, you can again use the optional third parameter, this time setting `load: false`, i.e.
+
+```
+    var autosave = rxAutoSave($scope, 'formData', { load: false });
+```
+
+`load:` will accept a boolean, or it can accept a promise that eventually resolves to a boolean. Accepting a promise will let you delay your decision on whether or not to load (for example, asking a user if they want values loaded). Note that if you use a promise, `rxAutoSave` will look at its resolved value. If the resolved value is `true`, then the data will be loaded. If the resolved value is `false`, or the promise fails/rejects, then the data will not be loaded.
+
+### Excluding some values from loading/saving
+
+By default, `rxAutoSave` automatically loads and saves all the stored values for a form. If you want to prevent it from loading/saving _some_ values, you can do:
+
+```
+    var autosave = rxAutoSave($scope, 'formData', { exclude: ['description'] });
+```
+
+This will tell `rxAutoSave` not to load from or save to the stored `description` value, but everything else in `formData` will be loaded/saved.
+
+### Manual saving
+
+It might be that you don't want your `rxAutoSave` instance to automatically save to the storage backend automatically. In some cases, you might want to disable automatic saving and instead manually tell your instance when it should save. To turn off automatic saving, set up your instance as follows:
+
+```
+    var manualsave = rxAutoSave($scope, 'formData', { save: false });
+```
+
+Then, whenever you want your `autosave` instance to commit the current model values to storage, do
+
+```
+    manualsave.save();
+```
+
+As with the `load` parameter, you can pass either a boolean or a promise to `save`.
+
+### Storage location
+All values for `rxAutoSave` are by default stored in the browser's `LocalStorage`, and keyed on the URL of the page, with a `rxAutoSave::` prefix. For example, if the above form were present at the URL `'users/JonnyRocket/edit'`, then the form data would be saved into `LocalStorage` at location `'rxAutoSave::users/JonnyRocket/edit'` 
+
+If you wish to use a different storage backend (`SessionStorage`, for instance), use the `storageBackend` parameter:
+
+
+```
+    var autosave = rxAutoSave($scope, 'formData', { storageBackend: SessionStorage });
+```
+
+`storageBackend` requires that you pass it an object which has `getObject(key)` and `setObject(key, val)` methods. `LocalStorage` and `SessionStorage` are both provided by EncoreUI, and support this interface.
+
+You can use your own custom backends as well, as long as it supports `getObject(key)` and `setObject(key, val)`.
--- a/src/rxMisc/docs/rxMisc.js
+++ b/src/rxMisc/docs/rxMisc.js
@@ -1,0 +1,48 @@
+function rxMiscCtrl ($scope, $timeout, $q, rxNotify, rxAutoSave) {
+    $scope.formData = {
+        checkbox: false,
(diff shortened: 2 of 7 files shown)


src/rxMisc/docs/rxMisc.js
```diff
@@ -0,0 +1,48 @@
...
+    // Clear with an explicit autosave.clear() call
+    $scope.clearStorage = function () {
+        $scope.status.loading = true;
+        $timeout(function () {
+            $scope.status.loading = false;
+            autosave.clear();
+            rxNotify.add("rxAutoSave data has been cleared! Navigate away and return, and the form won't be auto-populated", {
+                type: 'success'
+            });
+        }, 1000);
```
Droogans: Can you make this an interval instead? I know it's only making me wait 1000 msecs, but still.

```js
$interval(function () {
    // ...same code...
}, 1000, 1)
```

parlarjb: Making you wait is the whole point. I want the `<rx-button>` to have a chance to toggle itself, to have some visual indication that something actually happened

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,228 @@ angular.module('encore.ui.rxMisc', [])
...
+ * A factory that controllers can use to help automatically save and load
+ * (from LocalStorage) forms on any given page.
+ */
+.factory('rxAutoSave', function ($location, debounce, LocalStorage) {
+
+    /* 
+     * We'll version the schema for the stored data, so if we need to change
+     * the schema in the future, we can do automatic migrations. Never 
+     * delete any of these documented schemas. If you have to add a new version,
+     * then add it on top, but keep the documentatino for the old one around.
```
Droogans: *documentation

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,228 @@ angular.module('encore.ui.rxMisc', [])
...
+     *                      // Serialized form data
+     *                  }
+     *          }
+     *      }
+    */
+    var version = 1;
+
+    var key = function () {
+        return 'rxAutoSave::' + $location.url();
+    };
```
Droogans: Shouldn't functions have active names?

`getKey()` seems more accurate. Although I'm more a fan of properties, I'm ok with either.

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,228 @@ angular.module('encore.ui.rxMisc', [])
...
+
+    // This will be used by the rxAutoSave instance to interact with
+    // LocalStorage. It takes the key and watchVar strings as arguments.
+    // This is not publically exposed, it can only be used and accessed
+    // by the rxAutoSave instance
+    var StorageAPI = function (key, watchVar) {
+        this.key = key;
```
Droogans: This is technically a `keyFn`.

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,228 @@ angular.module('encore.ui.rxMisc', [])
...
+        var form = this.getForm();
+        form.data = val;
+        this.setForm(form);
+    };
+
+    // This is what we return from rxAutoSave, and calling this
+    // function will return an instance
+    return function (scope, watchVar, opts) {
+
+        var api = new StorageAPI(key(), watchVar);
```
Droogans: Ah, you are passing in the result of `key()`.

Is there any reason to pass it in? Should it be configurable if it's shielded from end-user execution like this?

Perhaps `key()` could be the default, or better, just drop the `key()` call and definition entirely.

```js
var StorageApi = function (watchVar) {
    this.key = 'rxAutoSave::' + $location.url();
    this.watchVar = watchVar;
};
```

`watchVar` should be assigned since it's passed in on line 255. The key? I don't think it should.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

parlarjb:
Right now, unless you explicitly set a `ttl` when creating your `rxAutoSave` instance, the form data will live in LocalStorage indefinitely. I'm now leaning towards a default `ttl` of 1 week

Droogans:
I think the important thing is that it shouldn't be forever. Suppose a page gets auto-saved, and then later is replaced with a newer version. This leaves an auto-saved form in localStorage until the user manually clears everything.

:+1: for one week.

parlarjb:
@halkon mentioned that maybe a week is too long. It might be surprising to return to a form five or six days later and see your data there. Two days?

CITguy:
Two days :+1:

src/rxMisc/README.md
```diff
@@ -2,3 +2,137 @@ A module for shared functionality across framework components.
...
+
+### Excluding some values from saving
+
+By default, `rxAutoSave` automatically saves all values in `formData`. But in many cases you'll want to exclude _certain_ values (maybe passwords, maybe auto-generated values, etc.). You can tell it to never save certain properties for a form with:
+
+```
+    var autosave = rxAutoSave($scope, 'formData', { excludeFromSave: ['description'] );
+```
+
+With this, this instance of `rxAutoSave` will never save the `description` value to `LocalStorage`.
```
CITguy: I'd keep all excluded keys in one option (`exclude`) rather than two (`excludeFromLoad`, `excludeFromSave`). Unless we have usecases to do so, I feel like splitting it up is more work than needed.

CITguy: Essentially, the interface would look like this (with default values):
```javascript
AutoSave($scope, 'formData', {
  clearOnSuccess: null, // Promise
  ttl: 172800000,       // Integer (two days default)
  load: true,           // Boolean
  exclude: []           // Array<String>
});
```

parlarjb: My use case for `excludeFromLoad` was future-looking. Have the option of excluding some field from being loaded, and then asking the user if they want that field loaded. 

But I'm not firmly tied to that and be convinced to just move to `exclude`

parlarjb:
Another API question: It feels kind of ridiculous to specify `ttl` in terms of milliseconds. Any objections to making it seconds?

halkon:
:+1: for seconds, milliseconds is a bit too much granularity that's unnecessary

src/rxMisc/README.md
halkon: So I totally don't disagree with having just an exclude function, most likely, `99.99999999%` of the use cases, somebody will want to exclude `save` and `load`, at the same time.

But in the weird cases where someone will want to `save` and not `load`, or do some weird trickery to confuse the users because it's fun, this obviously becomes hard (that's the only thing that came to mind, and being able to `load`/ `save` manual values from an API or something to that effect)

Sadly the one thing I will do is share the link from `$compile.link` in which it's either one function or broken down into an object of `pre` and `post` (in this case we could do `save` and `load` lists in an object)

That being said, it's overloading a parameter, so I'll slowly back the idea away.  (Or perhaps we have `exclude`, `excludeFromSave`, `excludeFromLoad` if we want to be explicit).

src/rxMisc/README.md
```diff
@@ -2,3 +2,137 @@ A module for shared functionality across framework components.
...
+
+### Storage location
+All values for `rxAutoSave` are stored in the browser's `LocalStorage`, and keyed on the URL of the page, with a `rxAutoSave::` prefix. For example, if the above form were present at the URL `'users/JonnyRocket/edit'`, then the form data would be saved into `LocalStorage` at location `'rxAutoSave::users/JonnyRocket/edit'`
```
halkon: For Ticketing I will be adding a `SessionStorage` service a la the `LocalStorage` one we already have in the framework, I might get in touch with you in regards to this, but I can see the case where we just use `SessionStorage` (as we don't really care how often the forms get lost, only if the immediate data loss is recovered).  Just wondering if that's crossed your mind?

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+    var StorageAPI = function (watchVar) {
+        this.key = 'rxAutoSave::' + $location.url();
+        this.watchVar = watchVar;
+    };
+
+    // Get all the saved data for this page. If none
+    // exists, then create an empty object that matches
+    // the current schema.
+    StorageAPI.prototype.getAll = function () {
+        return LocalStorage.getObject(this.key) || {
```
halkon: Mentioned in a previous comment, perhaps it might be beneficial to just add:

```
StorageAPI.prototype.getDataStorage = function () {
     return this.session ? SessionStorage : LocalStorage;
};
```

Also could start adding the ground work for one day in the future using an API (or letting an application define an API for this)

src/rxMisc/README.md
parlarjb: It has. I tried to keep everything pretty generic, in case we want to move to a storage API in the future. As it stands, it wouldn't take much to have an extra option `storageBackend` that defaults to `localstorage`, but can receive `sessionstorage` as an option. Then it'd be a matter of extending the private `StorageAPI` class

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+            }
+        };
+    };
+
+    // Given a `watchVar`, return the corresponding
+    // `form` object from LocalStorage. This form object should include
+    // both `.data` and `.config` properties.
+    // If no form currently exists for `watchVar`, then an empty
+    // object will be created that matches the current schema
+    StorageAPI.prototype.getForm = function () {
```
halkon: so right now, it's a 1:1 relationship between `Form` and `Instance` and `Instance` takes care of saving `Form` in `Storage` devices.  

More of a specific clarification than anything as I think through:

```
Form -> Instance
Instance -> Storage
Storage <-> Multiple Instances
```

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+    */
+    var version = 1;
+
+    // This will be used by the rxAutoSave instance to interact with
+    // LocalStorage. Its `watchVar` argument is the string name of the
+    // object that's being watched, representing the model for the form.
+    // StorageAPI is not publically exposed, it can only be used and accessed
+    // by the rxAutoSave instance
+    var StorageAPI = function (watchVar) {
+        this.key = 'rxAutoSave::' + $location.url();
```
halkon: What if there's 2 forms in one location, independent of each other?  Could we add an optional name/identifier parameter to the instance creation that can also be an option in `rxAutoSave`?

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+    };
+
+    // Return the current saved data for a given watchVar
+    StorageAPI.prototype.getDataObject = function () {
+        return this.getForm().data;
+    };
+
+    // For a given watchVar, store `val` as its saved
+    // data, into LocalStorage
+    StorageAPI.prototype.setDataObject = function (val) {
```
halkon: So, talking about `exclusions`, might as well have a function to add/remove *exclusions*.  There could be a use case in which a form value controls whether or not another value is set/created/edited, and may need to be saved/not saved

src/rxMisc/rxMisc.js
parlarjb: I guess I didn't make it explicit in the docs, but this already works. If you have a `formA` and a `formB`, holding their models in `$scope.formA` and `$scope.formB`, then you do:

```
    var autoSaveFormA = rxAutoSave($scope, 'formA');
    var autoSaveFormA = rxAutoSave($scope, 'formB');
```

and it "just works". I'll update the docs to point this out.

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+
+            updateExpiryTime();
+
+            // Write all the storedObject values into scope[watchVar], except
+            // for any specified in opts.excludeFromLoad
+            var storedObject = api.getDataObject();
+            _.assign(scope[watchVar], _.omit(storedObject, opts.excludeFromLoad));
+        };
+
+        // This is the "instance" that is returned when someone
```
halkon: Oh... I thought part of the `StorageApi` got returned/extended here.  Or at least some of the functionality would be implemented here

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,225 @@ angular.module('encore.ui.rxMisc', [])
...
+        // calls rxAutoSave($scope, 'someWatchVar')
+        var autoSaveInstance = {
+            clear: function () {
+                api.expire();
+            },
+
+            clearOnSuccess: function (promise) {
+                promise.then(function () {
+                    autoSaveInstance.clear();
```
halkon: I know we have discussed this, I just, tend to always hate creating an anonymous function just because of javascript instances and the nightmare they are. 

:crying_cat_face: 

I wish there was a more readable way than: `autoSaveInstance.clear.bind(autoSaveInstance)` or `_.bind(autoSaveInstance.clear, autoSaveInstance)`

src/rxMisc/README.md
halkon: :+1: awesome, I think there's a few areas in all the application where "storage" is an "issue" if you will, it may be fruitful (yet could be over-engineering) if we come up with a storage Service that encompasses `local`, `session` and `apiPromises`.

src/rxMisc/rxMisc.js
halkon: WOO HOO!

owl-baby-bath

src/rxMisc/README.md
parlarjb: The thing to keep in mind is that all of this works by using `$watch` on a model. That means if you want to load a custom value into one field of the model, then all you have to do is set the value yourself. 

i.e.
```
$scope.form = { foo: '', bar: ''}
rxAutoSave($scope, 'form', { excludeFromLoad: ['bar'] });

// manually set bar via an API
$scope.form.bar = SomeAPIValue()
```

halkon: Well, I completely didn't flush out my thought, so I'll just respond:
* Link: https://docs.angularjs.org/api/ng/service/$compile#-link-
* The use case of the API was more something that popped into my head, but you're right.
* I did forget that this is implemented by using `$watch` therefore any interaction outside of `saving` and `loading`  is no different than what we have today

src/rxMisc/rxMisc.js
parlarjb: You got it

src/rxMisc/rxMisc.js
CITguy: @halkon, do you have any use cases that we can test against? If not, I'd rather keep this feature as simple as possible for now and add that functionality later on if/when we need it.  I don't want to fall into the trap of over-engineering this service unless they're features we actually need.

halkon: Not at the moment no.  I'm quite tired and memory/history/creativity is very low today. 

Keep it simple it is! :+1:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

src/rxMisc/docs/rxMisc.js
```diff
@@ -0,0 +1,48 @@
+function rxMiscCtrl ($scope, $timeout, $q, rxNotify, rxAutoSave) {
+    $scope.formData = {
+        checkbox: false,
+        name: '',
+        description: '',
+        sensitive: ''
+    };
+
+    var autosave = rxAutoSave($scope, 'formData', {
+        excludeFromSave: ['sensitive'],
```
CITguy: change rom `excludeFromSave` to `exclude`

parlarjb: Good catch!

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,246 @@ angular.module('encore.ui.rxMisc', [])
...
+        
+        // We don't want to write to LocalStorage every time the model changes,
+        // because that would turn typing into a textarea into an expensive operation.
+        // We'll instead debounce the the writes for 1 second
+        var debounced = debounce(update, 1000);
+
+        $q.when(opts.save).then(function (shouldSave) {
+            if (shouldSave) {
+                // The `true` third argument tells $watch to do a deep comparison
+                scope.$watch(watchVar, debounced, true);
```
halkon: Sadly I don't think there's any way to capture this but just as an FYI, `$watch` returns a deregistration function for the listener, so that you can take it off of scope.

However I don't think there's any easy way of doing that from here as you probably don't know when you should stop `$watch`ing.  But if you get an :bulb: idea, even better

https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$watch

src/rxMisc/rxMisc.js
```diff
@@ -114,4 +114,248 @@ angular.module('encore.ui.rxMisc', [])
...
+     *                  }
+     *              }
+     *              "form2": {
+     *                  config: {
+     *                      expires: 33421234322,
+     *                  }
+     *                  data: {
+     *                      // Serialized form data
+     *                  }
+     *          }
```
megalord: I know this isn't intended to be a valid object, but it looks like a `}` is missing. :smile:

src/rxMisc/README.md
```diff
@@ -2,3 +2,196 @@ A module for shared functionality across framework components.
...
+```
+
+And that's it! Your `rxAutoSave` instance will watch for any change to `$scope.formData`, and will automatically write those changes to `LocalStorage`. 
+
+A third argument can be passed to `rxAutoSave`, specifying usage options. The default values for these options are:
+
+```
+    var autosave = rxAutoSave($scope, 'formData', {
+      clearOnSuccess: null,        // Promise
+      ttl: 172800000,              // Integer (two days default)
```
CITguy: I think there's too many zeros here. Also, note in the comment that it is in seconds `// Integer (seconds) - two days default` (or similar)

CITguy:
LGTM

halkon:
LGTM

Droogans:
Squash and LGTM

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/897/gallery.html)

PR #901, opened 2015-04-16
feat(LESS): Bundle our main LESS files with bower

### LINKED ISSUE #635: Export encore-ui's LESS file and have other products bring it in
Right now there is some duplicate code in the framework and in Cloud that concerns styling for Flexbox. See https://github.com/rackerlabs/encore-ui/blob/master/src/styles/flexbox.less and https://github.com/rackerlabs/encore-cloud-ui/blob/master/app/styles/cloud-mixins.less#L38-95

While responding to PR feedback here: https://github.com/rackerlabs/encore-cloud-ui/pull/277#issuecomment-63699234 I discovered that the need for that duplication exists, since variables get compiled with LESS and Cloud can't read/rely on variables defined in the Framework. 

We should look into including the LESS file from the Framework in each product so that LESS variables for flexbox can be compiled (and de-duped) in each codebase.

parlarjb:
Closes #635

This will include our flexbox.less, mixins.less and vars.less
with bower, so projects can build against our variables.

### DIFF
--- a/grunt-tasks/options/copy.js
+++ b/grunt-tasks/options/copy.js
@@ -109,6 +109,11 @@ module.exports = {
         }, {
             src: 'bower.json',
             dest: '<%= config.bower %>/bower.json'
+        }, {
+            expand: true,
+            cwd: '<%= config.exportableStyles %>/',
+            src: '*.less',
+            dest: '<%= config.bower %>/less/'
         }]
     }
 };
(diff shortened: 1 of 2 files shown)


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/901/gallery.html)

CITguy:
LGTM

halkon:
LGTM
...




..

LGTM

PR #903, opened 2015-04-16
fix(rxPaginate): Fix infinite loop on one page tables

parlarjb:
After deploying the new pagination code, we noticed some infinite
digest problems in Cloud. It turns out we were trying to go
to the last page over and over. Now we check if we're already
on the last page, and only go if we're not.


comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/903/gallery.html)

CITguy:
LGTM

PR #906, opened 2015-04-17
feat(rxStatusColumn): add disabled status

### LINKED ISSUE #766: Add "disabled" style for rxStatusColumn
This is something utilized heavily by the new dedicated Load Balancer tool coming into Encore.

megalord:
Fixes #766 

screen shot 2015-04-17 at 1 32 41 pm

### DIFF
--- a/src/rxStatusColumn/rxStatusColumn.less
+++ b/src/rxStatusColumn/rxStatusColumn.less
@@ -44,6 +44,9 @@ td.rx-status-column {
             color: #287BBD;
         }
     }
+    &.status-DISABLED {
+        background: #dededd;
+    }
 
     .fa-exclamation-circle {
         color: transparent;
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -400,7 +400,7 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
     $scope.pager = PageTracking.createInstance();
     $scope.people = [
         { name: 'Patrick Deuley', occupation: 'Design Chaplain', number: 1, status: 'ACTIVE' },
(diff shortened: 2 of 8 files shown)


src/rxStatusColumn/rxStatusColumn.less
```diff
@@ -44,6 +44,9 @@ td.rx-status-column {
             color: #287BBD;
         }
     }
+    &.status-DISABLED {
+        background: #cfd2d4;
```
CITguy: Where did this color come from? Should we use the documented #dededd as noted in the styleguide?

megalord: #dededd is lighter and (to me) does not have enough contrast with the grey table rows.  This color was agreed upon in the issue discussion, but unfortunately I do not recall where it was originally determined.

megalord: Here's what #dededd looks like.  @glynnis mind weighing in?
screen shot 2015-04-17 at 3 26 50 pm

glynnis:
Conferred with @ElementE. Let's go with #dededd since that's more in line with our docs re: disabled color. 

Once again this falls under the "we plan to normalize colors soon" plan, which we're closer and closer to doing. There are lots of conversations in the works on the design side about what tools we need in place before we can do that work safely, so I have hope that we'll be able to tackle it in the near future (finally).

Thanks for bringing this to our attention! :cake: And sorry for the confusion.

glynnis:
Design: LGTM

CITguy:
LGTM

PR #914, opened 2015-04-21
test(rxForm): Simple generation of form elements

### LINKED ISSUE #904: rxForm.create
I hate writing out forms....they look so ugly....

```js
txtName: {
    get: function () {
        return element(by.model('formData.name'));
    }
},

name: {
    get: function () {
        return this.txtName.getAttribute('value');
    },
    set: function (input) {
        this.txtName.clear();
        this.txtName.sendKeys(input);
    }
}
```

All of that boilerplate enables the use of `rxForm.fill`, a really useful function for auto-form filling via POJOs, but the amount of boilerplate needed to generate boring form getters and setters sort of prohibits its use for most people.

That would all change if instead I could replace the above code with

```js
name: rxForm.create.textField(element(by.model('formData.name')));
```

Whether or not this would include a side effect of adding a `txt{{PROPERTY_NAME}}` web element property to the page object is still up for debate.

The following web elements need to have their boilerplate supported:

text fields (`txt`-prefixed web elements)
checkbox fields (`chk`-prefixed web elements)
radio fields (`rad`-prefixed web elements)
dropdown fields (`sel`-prefixed web elements)

rxOptionFormTable would not be included.

Droogans:
Closes #904

BREAKING CHANGE: `rxForm.form.fill` is now `rxForm.fill`

### DIFF
--- a/src/rxForm/docs/rxForm.midway.js
+++ b/src/rxForm/docs/rxForm.midway.js
@@ -14,99 +14,37 @@ var formPageObject = Page.create({
 
     form: {
         set: function (formData) {
-            rxForm.form.fill(this, formData);
+            rxForm.fill(this, formData);
         }
     },
 
-    txtPlainTextbox: {
-        get: function () {
-            return elementByLabel('Plain textbox').$('input');
-        }
-    },
+    plainTextbox: rxForm.textField.generateAccessor(elementByLabel('Plain textbox').$('input')),
 
-    plainTextbox: {
-        get: function () {
-            return this.txtPlainTextbox.getAttribute('value');
-        },
-        set: function (input) {
-            this.txtPlainTextbox.clear();
-            this.txtPlainTextbox.sendKeys(input);
-        }
-    },
-
-    chkRequireName: {
-        get: function () {
-            return elementByLabel('Require Name?').$('input');
-        }
-    },
-
-    requireName: {
-        get: function () {
-            return this.chkRequireName.isSelected();
-        },
-        set: function (enable) {
-            var checkbox = rxForm.checkbox.initialize(this.chkRequireName);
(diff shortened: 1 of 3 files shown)


megalord:
:+1: This is a big help, thanks!

parlarjb:
Could you add a note about this feature in `utils/rx-page-objects/README.md`?

Droogans:
Yes. That's a great point. I need to highlight the biggest time-savers right up front, which has been an issue with onboarding.

Droogans:
@parlarjb there it is!

parlarjb:
Rather than introduce a breaking change, could you have `rxForm.form.fill` just call to `rxForm.fill`?

Droogans:
I say don't bump the framework for page objects breaking API changes. This is a minor version change since it's just tests, and seeing how it's a single `find`/`sed` call from fixing in any project, it's not a lot of work.

parlarjb:
The issue is that we want the version numbers to indicate when a breaking change has happened. If someone waits awhile to upgrade, they're probably not going to go through all the release notes in between upgrades and look for breaking changes. They'll just rely on the version number. 

It's easy to find with `find / sed`, but people won't know to do that, they'll just have tests start to break.

Droogans:
Let's just leave this open until the next breaking change comes along, and I'll remember to include this with it.

parlarjb:
Did you mean to leave the PR open, and you'll wait for a breaking change release? I'm not sure when we'll be doing that.

Droogans:
@parlarjb I updated the commit message to be non-breaking!

Also, I went with using an alias. No sense waiting around for a breaking change.

parlarjb:
Thanks!

Issue #824, opened 2015-02-27
Create UX/visual design for collapsible divs


glynnis:
This has come up specifically for filters, but we should design a pattern that allows any div to be collapsed (not just filters).

Look, feel, and interaction design for this is specified in:
https://projects.invisionapp.com/d/main#/console/2965787/66281327/comments/31724169
https://projects.invisionapp.com/d/main#/console/2965787/66342618/comments/31724257

When this gets added to the Framework @velveteer, let's make sure we update the filter box in Transaction Details (Billing Tabs design)

glynnis:
See https://github.com/rackerlabs/encore-ui/issues/218

CITguy:
@glynnis is this issue still valid considering the inclusion of `rxCollapse`?

glynnis:
@CITguy ah, should've been closed on merge. Closing.

PR #915, opened 2015-04-22
feat(rxCollapse): Add collapsible component

julianbonilla:
Expanded

expanded

Collapsed

collapsed

### DIFF
--- a/src/rxCollapse/README.md
+++ b/src/rxCollapse/README.md
@@ -1,0 +1,8 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+The rxCollapse element can be used to display and hide content. It can be configured to show as either expanded or collapsed on page load. A double chevron is used as a toggle to show/hide the inner contents, while keeping the header and border visible.
+
+### Attributes
+
+- `title`: The title to display next to the toggle button.
+- `expanded`: Initially expanded or collapsed.  Default is expanded.
--- a/src/rxCollapse/rxCollapse.js
+++ b/src/rxCollapse/rxCollapse.js
@@ -1,0 +1,30 @@
+angular.module('encore.ui.rxCollapse', [])
+/**
+ * @ngdoc directive
+ * @name encore.ui.rxCollapse:rxCollapse
+ * @restrict E
+ * @scope
+ * @description
+ * Hide and show an element with a transition.
+ *
+ * @param {string} [title] The title to display next to the toggle button.
+ * @param {string} [expanded] Initially expanded or collapsed.  Default is expanded.
+ *
+ * @example
+ * <pre>
+ *     <rx-collapse title="Filter results" expanded="true"></rx-collapse>
+ * </pre>
+ */
+.directive('rxCollapse', function () {
+    return {
+        restrict: 'E',
+        templateUrl: 'templates/rxCollapse.html',
+        transclude: true,
+        scope: {
+            title: '@'
+        },
+        link: function (scope, element, attrs) {
+            scope.isExpanded = (attrs.expanded === 'false') ? false : true;
+        }
+    };
+});
--- a/src/rxCollapse/rxCollapse.less
+++ b/src/rxCollapse/rxCollapse.less
@@ -1,0 +1,49 @@
+/*
+ * rxCollapse
+ */
+rx-collapse {
+  display: block;
+
+  .collapse-container {
+
+    .double-chevron-cell {
+      display: inline-block;
+      .box-sizing(border-box);
+      width: 30px;
+      padding: 0 10px;
+      background-color: @borderGrey;
+      vertical-align: middle;
+
+      .double-chevron {
+        cursor: pointer;
+        &:after {
+          display: inline-block;
+          color: @white;
+          font-size: 28px;
+          .chevron-mixin(-1);
+        }
+      }
+
+      &.expanded {
+        .double-chevron {
+          &:after {
+            .chevron-mixin;
+          }
+        }
+      }
+    }
+
+    border: 2px solid @borderGrey;
+
+    .rx-collapse-title {
+      display: inline-block;
+      margin: 0 10px;
+      vertical-align: middle;
+    }
+
+    .collapse-body {
+      margin: 0;
+      padding: 10px;
+    }
+  }
+}
--- a/src/rxCollapse/templates/rxCollapse.html
+++ b/src/rxCollapse/templates/rxCollapse.html
@@ -1,0 +1,8 @@
+<div class="collapse-container">
+    <div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+        <a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+    </div>
+    <h3 class="rx-collapse-title title">{{title}}</h3>
+    <div class="collapse-body" ng-show="isExpanded" ng-transclude>
+    </div>
+</div>
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -182,6 +182,10 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                             linkText: 'Lists'
                         },
                         {
+                            href: '#/styleguide/basics#collapsible',
+                            linkText: 'Collapsible Element'
+                        },
+                        {
                             href: '#/styleguide/basics#wells',
                             linkText: 'Wells'
                         },
@@ -424,4 +428,24 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
 })
 .controller('commentBoxStyleguideCtrl', function ($scope) {
     $scope.comment = '';
+})
+
+// Used by the Filters styleguide page to show a collapsible region
+.controller('collapseFilterExampleCtrl', function ($scope) {
+    $scope.filter = { region: '' };
+
+    $scope.regions = [
+        { name: 'DFW', city: 'Dallas-Fort Worth' }, { name: 'ORD', city: 'Chicago' },
+        { name: 'IAD', city: 'Northern Virginia' }, { name: 'LON', city: 'London' },
+        { name: 'HKG', city: 'Hong Kong' }, { name: 'SYD', city: 'Sydney' }
+    ];
+
+    $scope.servers = [
+        { name: 'General1-1', ram: '1 GB', cpu: 1, disk: '20GB SSD', region: 'DFW'},
+        { name: 'General1-2', ram: '2 GB', cpu: 2, disk: '40GB SSD', region: 'ORD'},
+        { name: 'General1-4', ram: '4 GB', cpu: 4, disk: '80GB SSD', region: 'IAD'},
+        { name: 'General1-8', ram: '8 GB', cpu: 8, disk: '160GB SSD', region: 'LON'},
+        { name: 'I/O1-15', ram: '15 GB', cpu: 4, disk: '40GB SSD', region: 'HKG'},
+        { name: 'I/O1-30', ram: '30 GB', cpu: 8, disk: '40GB SSD', region: 'SYD'}
+    ];
 });
--- a/demo/styleguide/basics.html
+++ b/demo/styleguide/basics.html
@@ -137,6 +137,16 @@
     </div>
     <rx-styleguide code-url="styleguide/basics/lists.html"></rx-styleguide>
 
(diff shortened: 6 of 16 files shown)


parlarjb:
Did @glynnis have a hand in this? And how does it relate to the "Nested Content" demos here http://rackerlabs.github.io/encore-ui/#/styleguide/tables ?

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,94 @@
...
+  .box-sizing(border-box);
+  position: relative;
+  padding: 6px;
+  border: 2px solid @borderGrey;
+
+  .title {
+    margin: 0 2em;
+  }
+
+  .collapse-body {
```
julianbonilla: Need some feedback on `.collapse-body` style.  The current use case is for showing/hiding a collection of `rx-form-item` filters.  Since we don't have a horizontal style for that component, these styles are opinionated about the layout of form components.  The other option is not providing any styles for this element, which might lead to some inconsistency across products.

julianbonilla:
@parlarjb This is based on work done in Billing by @glynnis and @velveteer.

parlarjb:
Great, thanks.

We'll need all the tests filled in, and an addition to the style guide. I'd talk to @glynnis to see where in the style guide she'd like it to appear.

And add an exercise for collapsing/expanding the box. 

You should also remove the `src/rxCollapse/docs/rxCollapse.js` file if it's not being used.

src/rxCollapse/README.md
```diff
@@ -0,0 +1,8 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+The rxCollapse component provides a toggle for expanding and collapsing an element. It provides a common layout and style for showing and hiding a collection of filters for a table or list.
+
+### Attributes
+
+- `type`: title: The title to display next to the toggle button.
```
megalord: Should `type:` be removed (leaving only `title:`)?

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,94 @@
...
+.double-chevron-cell.expanded {
+  .double-chevron {
+    cursor: pointer;
+    &:after {
+      display: inline-block;
+      color: #bbbbbb;
+      font-size: 28px;
+      content: "\00bb";
+      transform: rotate(90deg) scaleX(1);
+      transition: all .2s;
```
megalord: Can you use the [chevron mixin](https://github.com/rackerlabs/encore-ui/blob/master/src/styles/mixins.less) instead?

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+.double-chevron-cell.expanded {
+  .double-chevron {
+    cursor: pointer;
+    &:after {
+      display: inline-block;
+      color: #bbbbbb;
+      font-size: 28px;
```
megalord: Correct me if I'm wrong, but I don't believe these styles need to be repeated (from above).  Any rules applied to `.double-chevron-cell .double-chevron` will still be applied if the `.double-chevron-cell` element also has the `.expanded` class.

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,90 @@
...
+  .box-sizing(border-box);
+  width: 30px;
+  padding: 0px 10px;
+  .double-chevron {
+    cursor: pointer;
+    &:after {
+      display: inline-block;
+      color: #bbbbbb;
+      font-size: 28px;
+      .chevron-mixin(-1)
```
CITguy: You need a semicolon at the end of this line.

src/rxCollapse/rxCollapse.less
CITguy: I believe you are correct. This can be included as such:

```less
.double-chevron-cell {
  .box-sizing(border-box);
  width: 30px;
  padding: 0px 10px;

  .double-chevron {
    cursor: pointer;
    &:after {
      display: inline-block;
      color: #bbbbbb;
      font-size: 28px;
      .chevron-mixin(-1);
    }
  }

  &.expanded {
    .double-chevron {
      &:after {
        .chevron-mixin;
      }
    }
  }
}
```

kobowley:
@glynnis see jay's request that you weigh in on where this should be in the documentation.

CITguy:
The LESS can be organized a bit better (you're defining styles for `.double-chevron` multiple times and it doesn't seem necessary.  Considering that `rx-collapse` is an element, you can style directly to the element selector `rx-collapse`. Additionally, these styles need to be contained within the `rx-collapse` scope, so they don't bleed into the global styles. As such I'd recommend using a LESS structure like this:

```less
rx-collapse {
  display: block; // override the browser default

  .collapse-container {
    .double-chevron-cell {
      .double-chevron { ... }
    }
    .title { ... }
    .collapse-body {
      margin: 0px;
      padding: 10px; // or other reasonable value
    }
  }
}//rx-collapse
```

IMO, it's not the responsibility of `rx-collapse` to style `rx-form-item` or other directives within its body.  `.collapse-body` only needs to provide space for contents to sit, so it really only needs to set a reasonable `padding` value. @glynnis, thoughts on this?

glynnis:
Closes #824

glynnis:
I agree with @CITguy's most recent comments about refactoring/cleaning up the LESS -- let's just make sure we double check the styles for expandable table rows that may be affected, just to ensure the refactor didn't miss anything :+1:

I think documentation in the Styleguide should probably go here: http://rackerlabs.github.io/encore-ui/#/styleguide/basics before or after "Wells".

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/915/gallery.html)

julianbonilla:
@glynnis Added documentation and an example to the Styleguide.

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,50 @@
...
+            .chevron-mixin;
+          }
+        }
+      }
+    }
+
+    border: 2px solid @borderGrey;
+
+    h3 {
+      display: inline-block;
```
CITguy: This would make more sense on the `.title` selector. Otherwise, it will affect **all** H3 tags within the directive (regardless if they're in the template).

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/915/gallery.html)

src/rxCollapse/rxCollapse.less
glynnis: :+1:

src/rxCollapse/rxCollapse.less
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    border: 2px solid @borderGrey;
+
+    .rx-collapse-title {
+      display: inline-block;
+    }
+
+    .title {
+      margin: 0 1em;
```
CITguy: I'd put this definition with `.rx-collapse-title`, as well.

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -0,0 +1,8 @@
+<div class="collapse-container">
+    <div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+        <a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+    </div>
+    <h3 class="rx-collapse-title title">{{title}}</h3>
```
CITguy: :+1:

src/rxCollapse/rxCollapse.js
```diff
@@ -0,0 +1,28 @@
...
+ *     <rx-collapse title="Filter results" expanded="true"></rx-collapse>
+ * </pre>
+ */
+.directive('rxCollapse', function () {
+    return {
+        restrict: 'E',
+        templateUrl: 'templates/rxCollapse.html',
+        transclude: true,
+        link: function (scope, element, attrs) {
+            scope.title = attrs.title || 'Filter';
```
parlarjb: One issue I've had in the past with this pattern is that if the title changes, it won't get picked up. For instance, maybe the title isn't ready on page load, because it's coming in from an API. Then the title will get set to `'Filter'`, and there will be no way to update it.

You could have `title` defined in `scope: ...`, and put a `$watch` on it (or maybe you can watch `attrs`, I'm not sure).

glynnis:
So I think if we're removing the styling for filters from this element (styles like those in Billing), then we should not be using filters as an example in the style guide or demo.

Here's my reasoning:
1) If people do use filters within the collapsible component, we would indeed want them to look like what Billing has implemented. However, I understand @CITguy's point about separating concerns.
2) The filter styles of radio buttons in the docs is not how we would prefer they be displayed from a design standpoint, but if we use it as an example that style will proliferate/become dominant.
3) The collapsible element can be utilized for things other than filters.

@CITguy @julianbonilla re: # 1, do you think there is a good way to offer/demonstrate the filter styling as Billing has done in a way that satisfies separation of concerns? If so, let's do that. The demo Julian currently has on http://localhost:9001/#/styleguide/basics ("Filters") would be better placed on the tables page and retitled "Collapsible component for table filters".

In light of # 2 and # 3 (whether or not we continue with filter styles and move them to the Tables page), we should add a demo to the style guide on the Basics page that just shows a collapsible element with a title + some "lorem ipsum" body text. Since it's really a collapsible element, not a filter, I'd change the text to be something more like:

```
Collapsible Element

- The collapsible element can be used to display and hide content.
- It can be configured to show as either expanded or collapsed on page load.
- A double chevron is used as a toggle to show/hide the inner contents, while keeping the header and border visible.
```

Let me know if any of this doesn't make sense -- happy to sync up on a call if that's helpful.

glynnis:
Other small visual thing: looks like the padding on the Framework version could be evened out a little to look more like what's been implemented in Billing.

Billing (what we want it to look like):
screen shot 2015-04-27 at 1 44 41 pm

This PR:
screen shot 2015-04-27 at 1 49 42 pm

The goal is for the title to be roughly vertically centered when the component is collapsed, with a similar amount of padding for the left side.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/915/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/915/gallery.html)

julianbonilla:
@glynnis @CITguy Any more feedback on Styleguide changes?

glynnis:
LGTM

parlarjb:
Could you fix the failing test? After that, LGTM 

https://travis-ci.org/rackerlabs/encore-ui/builds/60434082#L2501

CITguy:
I'm not sure about the collapsible table filtering pattern. There's no feedback that the table is filtered when it's collapsed.

*Filtered by ORD, but there's no indication of applied filters when collapsed.*
screen shot 2015-04-30 at 11 06 13 am

I'd feel more comfortable excluding this pattern for now and capture a story to revisit once the component has been released. I like the idea, but I don't think the pattern has matured yet. @glynnis, thoughts?

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/915/gallery.html)

kobowley:
@CITguy given that this component is already part of billing UI and is in production. I specifically asked for it in the framework. I can see you point about the user's needing feedback when the filter is collapsed, this can be dealt with separately. Please add an issue to the waffle board. LGTM on design and product sign off.

CITguy:
@kobowley: I know we're keeping the rxCollapse, but are we keeping the collapsible table filter pattern in the styleguide or removing it?

glynnis:
@CITguy I think we should keep the collapsible table filter pattern, yes -- that's what is in billing, and I think it's an important example to have present for other products.

That said, I think @kobowley is right -- let's open an issue for improving the feedback given to users in the case of the filter, so that it's more obvious when the box is collapsed that the table has been filtered.

CITguy:
ok, I wasn't sure whether the issue would be to update the example in this PR or to define it once it was removed. I'll get that issue in place.

Issue #884, opened 2015-04-03
rxForm (and ilk) needs refactor/overhaul


CITguy:
While attempting to create an `rxSelect` directive based on http://jsbin.com/berose, I've run into multiple issues.

### 1. rxFormItem Implementation
Implementation of `rx-form-item` (in its current form) requires selective transclusion (rearranging content to fit the directive template).  Angular 1.x does not have functionality to perform selective transclusion (that I know of). See #748 

### 2. Flexbox Implementation
*The blue boxes are to show the boundaries of the label and content.*

**Incorrect implementation (current)** - Current implementation requires JS logic to dynamically add css class to style the wrapper div. The screenshot shows that it is unsuccessful in Firefox for the text area. 
![screen shot 2015-04-03 at 3 01 07 pm](https://cloud.githubusercontent.com/assets/545605/6988102/83ebbf34-da13-11e4-80d4-75b21d1c0c6a.png)

**Correct implementation** - No need for additional CSS class for edge case styling (i.e. textarea)
![screen shot 2015-04-03 at 3 07 39 pm](https://cloud.githubusercontent.com/assets/545605/6988117/ad2d89f4-da13-11e4-999a-6714e91ec983.png)

### 3. Advanced Form Element Styles
As per proof-of-concept at http://jsbin.com/wusiki, we can do some advanced styling of native form elements (and we're beginning to see the need for it - per [this comment](https://github.com/rackerlabs/encore-ui/issues/758#issuecomment-84167612)), but current directives either hinder implementation or require a great deal of logic overhaul to add support.

For example, the `rx-form-item` directive attempts to apply an `id` attribute to the transcluded form element for `<label for="id_here">` functionality.  Unfortunately, because of its implementation and hacky logic for selective transclusion, when applied to an `rxSelect` directive, the ID is placed on the wrapper div rather than the select, which is incorrect.  The design of `rx-form-item` would require knowledge of the structure of `rx-select` in order to apply the ID correctly, and this problem would only get worse with the addition of more custom element directives.

These three items alone should show that we need to rethink the way we implement form element styling and directives to support them.

halkon:
![YESSSS](https://cloud.githubusercontent.com/assets/1437154/7004907/c0151872-dc37-11e4-834e-9083241e34f1.gif)

halkon:
[...]

PR #922, opened 2015-04-27
feat(rxSelectFilter): Add table filtering component

### LINKED ISSUE #912: Multiple select table filtering
There is a need for a multiple select dropdown that can be used to filter table data.  Ticketing is already using this style, but this is not yet part of the framework.

screen_shot_2015-04-17_at_4 01 17_pm

megalord:
Fixes #912 

select-filter

### DIFF
--- a/src/rxPaginate/docs/rxPaginate.html
+++ b/src/rxPaginate/docs/rxPaginate.html
@@ -31,7 +31,7 @@
     </p>
     
     <h2 class="title">API-Based Pagination</h2>
-    <p>The API used by this demo is returning 100 items more than the user's selected `itemsPerPage`. If the user's `itemsPerPage` is 25, then the API will return 125 itmems on each request. This means with the default `itemsPerPage` of 25, five pages of results are coming back at a time. You should be able to click through pages 1-5 without a loading message, and then the loading message will appear for page 6.</p>
+    <p>The API used by this demo is returning 100 items more than the user's selected `itemsPerPage`. If the user's `itemsPerPage` is 25, then the API will return 125 items on each request. This means with the default `itemsPerPage` of 25, five pages of results are coming back at a time. You should be able to click through pages 1-5 without a loading message, and then the loading message will appear for page 6.</p>
 
     <p>Click the "Refresh" button to see how the current page can be reloaded without the user interacting with the <code>&lt;rx-paginate&gt;</code> buttons.</p>
 
@@ -41,11 +41,9 @@
     <table rx-floating-header rx-loading-overlay class="table-striped demo-api-pagination">
         <thead>
             <tr>
-                <th colspan="2">
-                    <input placeholder="Filter by any..." style="font-style: italic" ng-model="data.searchText">
-                    <button class="pull-right button" ng-click="clearFilter()">
-                        Clear Filter
-                    </button>
+                <th colspan="2" class="filter-header">
+                    <rx-search-box rx-placeholder="'Filter by any...'" ng-model="data.searchText"></rx-search-box>
+                    <rx-select-filter filter="osFilter"></rx-select-filter>
                 </th>
             </tr>
             <tr>
@@ -85,6 +83,7 @@
                         server-interface="serverInterface"
                         error-message="Error loading page!"
                         filter-text="data.searchText"
+                        selections="osFilter.selected"
                         sort-column="sort.predicate"
                         sort-direction="sort.reverse">
                     </rx-paginate>
@@ -93,3 +92,14 @@
         </tfoot>
     </table>
 </div>
+<style>
+    rx-search-box {
+        font-style: italic;
+    }
+    rx-select-filter {
+        margin-left: 20px;
+    }
+    .os-filter .rx-multi-select {
+        width: 200px;
+    }
+</style>
--- a/src/rxSelectFilter/README.md
+++ b/src/rxSelectFilter/README.md
@@ -1,0 +1,168 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+A multi-select dropdown for table filtering.
+
+# Services
+
+## SelectFilter
+
+This service exposes an object with single method, `create()`, used to create instances of a `SelectFilter`.  It is configurable via three options:
+- `properties`: A list of the properties to create a filter control for.  Assuming the source data is an array of objects, a property is equivalent to an object's key.
+  ```
+  SelectFilter.create({
+      properties: ['year']
+  });
+  ```
+- `available` (optional): An object that tracks which options are available for a property. It has the form:
+  ```
+  SelectFilter.create({
+      // other options...
+      available: {
+          year: [2013, 2014, 2015],
+      }
+  });
+  ```
+Note that the key of the object matches a value in the `properties` array.
+- `selected` (optional): An object that tracks which options are selected for a property. It has the same form as the `available` object, but the arrays indicate which options are selected, and as such are strict subsets of their `available` counterparts.
+  ```
+  SelectFilter.create({
+      // other options...
+      selected: {
+          year: [2014],
+      }
+  });
+  ```
+
+### Option Defaults
+
+Every property that is listed in `properties` but not provided as a key to `available` will be automatically populated the first time `applyTo()` (see below) is called.
+```
+var filter = SelectFilter.create({
+    properties: ['year']
+});
+
+filter.applyTo([{
+    eventId: 1,
+    year: 2013
+}, {
+    eventId: 2,
+    year: 2014
+}, {
+    eventId: 3,
+    year: 2013
+}]);
+// filter.available is { year: [2013, 2014] }
+```
+**Note:** There is an implied requirement that, when relying on the auto-populated filter, the input array will have at least one item for every available option.  For example, this may not be the case when used with server-side pagination.
+
+
+Every property that is listed in `properties` but not provided as a key to `selected` is initialized to have all options selected (by looking them up in `available`).  If property is also not provided to `available`, its initialization is delayed until the first call of `applyTo()`.
+
+```
+var filter = SelectFilter.create({
+    properties: ['year'],
+    available: {
+        year: [2013, 2014, 2015]
+    }
+});
+// filter.selected is { year: [2013, 2014, 2015] }
+```
+
+
+### Instances
+
+Instances of `SelectFilter` have an `applyTo()` method, which applies the filter's internal state of selected options to the array. This will not often be called directly, instead used by the `Apply` filter (see below).  As stated previously, the first call of `applyTo()` will initialize any `properties` that have not been defined in `available` or `selected`.
+
+```js
+var filter = SelectFilter.create({
+    properties: ['year'],
+    selected: {
+        year: [2014]
+    }
+});
+
+var filteredArray = filter.applyTo([{
+    eventId: 1,
+    year: 2013
+}, {
+    eventId: 2,
+    year: 2014
+}, {
+    eventId: 3,
+    year: 2013
+}]);
+// filteredArray is [{ eventId: 2, year: 2014 }]
+```
+
+The instance will also have all of the constructor options as public properties, so that they can be watched or changed.
+
+
+# Directives
+
+## rxSelectOption
+
+`<rx-select-option>` is to `<rx-multi-select>` as `<option>` is to `<select>`.
+
+Just like `<option>`, it has a `value` attribute and uses the element's content for the label. If the label is not provided, it defaults to a titleized version of `value`.
+
+```
+<rx-select-option value="DISABLED">Disabled</rx-select-option>
+```
+
+
+## rxMultiSelect
+
+A replacement for `<select multiple>` when space is an issue, such as in the header of a table.
+
+The options for the control can be specified by passing an array of strings (corresponding to the options' values) to the `options` attribute of the directive or using `<rx-select-option>`s. An 'All' option is automatically set as the first option for the dropdown, which allows all options to be toggled at once.
+
+The following two dropdowns are equivalent.
+```
+<!-- $scope.available = [2014, 2015] -->
+<rx-multi-select ng-model="selected" options="available"></rx-multi-select>
+
+<rx-multi-select ng-model="selected">
+    <rx-select-option value="2014"></rx-select-option>
+    <rx-select-option value="2015"></rx-select-option>
+</rx-multi-select>
+```
+
+This component requires the `ng-model` attribute and binds the model to an array of the selected options.
+
+The preview text (what is shown when the element is not active) follows the following rules:
+* If no items are selected, show "None".
+* If only one item is selected from the dropdown, its label will display.
+* If > 1 but < n-1 items are selected, show "[#] Selected".
+* If all but one item is selected, show "All except [x]"
+* If all items are selected, show "All Selected".
+
+
+## rxSelectFilter
+
+Uses an instance of `SelectFilter` to create a set of `<rx-multi-select>`s that modify the instance object.
+```
+// In the controller
+$scope.filter = SelectFilter.create({
+    // options...
+});
+
+// In the template
+<rx-select-filter filter="filter"></rx-select-filter>
+```
+
+# Filters
+
+## Apply
+
+Merely calls the `applyTo()` method of a `SelectFilter` instance to an input array.
+
+```html
+<tr ng-repeat="item in list | Apply:filter">
+```
+
+
+# Styling
+
+Remember to adhere to the Null Pattern for [tables](#/styleguide/tables) in Encore UI.  When there are no results to populate the table, the table content should say "No [x] were found."  However, if the table is empty because of the applied filters, the message should instead say "No results match those criteria."  This has been included in the demo below for an example implemtation.
+
+The developer is reponsible for ensuring that the multi-select input is wide enough to show the preview text. Especially for the "All except [x]" case, the text could wrap if the proper width is not set.  To facilitate this, a class is provided that adds a '-filter' suffix to the filtered property.  See the demo markup for an example.
--- a/src/rxSelectFilter/docs/rxSelectFilter.html
+++ b/src/rxSelectFilter/docs/rxSelectFilter.html
@@ -1,0 +1,45 @@
+<!-- Sample HTML goes here as a live example of how the component can be used -->
+<div ng-controller="rxSelectFilterCtrl">
+    <h3 class="title">rxMultiSelect Input</h3>
+    <rx-form-item label="Classification">
+        <rx-multi-select ng-model="data.classification" id="classification">
+            <rx-select-option value="A">Type A</rx-select-option>
+            <rx-select-option value="B">Type B</rx-select-option>
+            <rx-select-option value="C">Type C</rx-select-option>
+            <rx-select-option value="D">Type D</rx-select-option>
+        </rx-multi-select>
+    </rx-form-item>
+    <span>Model value: [{{ data.classification.join(', ') }}]</span>
+
+    <h3 class="title">Filtered Table</h3>
+    <table class="table-striped">
+        <thead>
+            <th colspan="6" class="filter-header">
+                <rx-select-filter filter="filter"></rx-select-filter>
+            </th>
+            <tr>
+                <th>Account</th>
+                <th>Status</th>
+                <th>Description</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr ng-repeat="ticket in (filteredTickets = (tickets | Apply:filter))">
+                <td>{{ticket.account}}</td>
+                <td>{{ticket.status}}</td>
+                <td>{{ticket.description}}</td>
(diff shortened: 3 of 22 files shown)


parlarjb:
Before I look into this, I'll note that it'd be nice if it also could work natively with server-side pagination and rxPaginate, like we do with filter boxes, sorting and column order. See https://github.com/rackerlabs/encore-ui/blob/master/src/rxPaginate/rxPaginate.js#L126 for the start of that stuff. I can talk to you about this if you'd like more details.

(Thanks for this work, btw!)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+A select dropdown for table filtering.
```
glynnis: "A select dropdown" -> "A multi-select dropdown"

glynnis: Does this only work in the context of filtering? Or could this type of input be used in a different context?

megalord: I purposefully wrote the code so that the input could be used independently of the filtering part, however, I'm not sure we should encourage devs to do so - it depends on how you feel about `<select multiple>`.  The way I see it, this component should only be used when `<select multiple>` takes up too much space, such as in the header of a table, but I'm certainly open to suggestions.

glynnis:
So I'd like to mark this as "blocked" until we get https://github.com/rackerlabs/encore-ui/issues/884 merged and released, that way we can use new form styles and check boxes here. @megalord would that work for you? @CITguy says it'll likely be ~1-2wks.

megalord:
The application work that this pertains to cannot wait that long, but I can just copy this code into that project.  That being said, it would be nice if the code gets reviewed soon to a point where we can say, "ready for release except new form styles".  That way, I can copy the code with good confidence that the interfaces will not change, and can be easily swapped out later.

Perhaps it would be possible to move forward with this PR, and then just make sure that this code gets updated for the same release as the new styles.  Either way, what's really important to me is that we nail down how the component should work.

glynnis:
@megalord Ok, sounds good. After speaking with @CITguy it sounds like the checkboxes can be easily inserted later on and incorporated in a future release without any pain downstream.

I want to tweak how the design looks for the multi-select that reveals what's selected to the side -- I think the `[]` present before a user has made a selection is a little confusing. If it sounds good to you @megalord, I'm just gonna hop on the branch and make the style changes directly, rather than passing you a mock-up and asking you to figure out how to implement :smile: 

Sound good? I'll jump on this branch today to wrap up those minor design concerns so we can fit this in with a release sooner rather than later.

megalord:
Oh, so the `[]` is actually just part of the demo, not the component itself, but I'll move it and put some explanatory text.  The text of the multi-select itself (rather than the options) indicates which options are selected, but it's a bit primitive - it only says "None", "Multiple", "All", or the label of the selected option if only one is selected.

glynnis:
@megalord Ohhhhh. *facepalm*. Eric and I were thinking it was an intentional part of the control to indicate to the user what's selected. It's possible that others looking at the demo would make that assumption too and incorporate that into designs.

If the `[]` is clearly indicated as just for demo purposes and not the design pattern, that's fine. Could you left-align the "Account" label so it's not indented in the "Filtered Table" demo?

With those two things addressed, LGTM

megalord:
With updates...
screen shot 2015-04-29 at 9 27 34 pm

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

src/rxSelectFilter/rxSelectFilter.js
```diff
@@ -0,0 +1,243 @@
...
+                    return this.options.length === $scope.selected.length;
+                } else {
+                    return _.contains($scope.selected, option);
+                }
+            };
+        },
+        link: function (scope, element, attrs, controllers) {
+            var selectElement = rxDOMHelper.find(element, '.rx-multi-select')[0];
+
+            var documentClickHandler = function () {
```
parlarjb: Did you forget to set `event` as an arg here?

megalord: Sure did... now I'm wondering why it still works.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

src/rxPaginate/docs/rxPaginate.html
```diff
@@ -31,7 +31,7 @@ <h2 class="title">UI-Based Pagination</h2>
     </p>
     
     <h2 class="title">API-Based Pagination</h2>
-    <p>The API used by this demo is returning 100 items more than the user's selected `itemsPerPage`. If the user's `itemsPerPage` is 25, then the API will return 125 itmems on each request. This means with the default `itemsPerPage` of 25, five pages of results are coming back at a time. You should be able to click through pages 1-5 without a loading message, and then the loading message will appear for page 6.</p>
+    <p>The API used by this demo is returning 100 items more than the user's selected `itemsPerPage`. If the user's `itemsPerPage` is 25, then the API will return 125 items on each request. This means with the default `itemsPerPage` of 25, five pages of results are coming back at a time. You should be able to click through pages 1-5 without a loading message, and then the loading message will appear for page 6.</p>
```
parlarjb: :+1:

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+A multi-select dropdown for table filtering.
+
+# Services
+
+## SelectFilter
+
+This service exposes an object with single method, `create`, used to create instances of a `SelectFilter`.  It is configurable via three attributes:
+- `properties`: A list of the properties to create a filter control for.  Assuming the source data is an array of objects, a property is equivalent to an object's key.
```
parlarjb: Could you give a "for example" here? I'm having a hard time parsing the meaning

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+- `available`: An object that tracks which options are available for a property. It has the form:
+```
+{
+      property1: ['value1-1', 'value1-2'],
+      property2: ['value2-1', 'value2-2', 'value2-3']
+}
+```
+If not provided, this object will be automatically populated the first time `apply` is called.
+- `selected`: An object that tracks which options are selected for a property. It has the same form as the `available` object, but the array values indicate which options are selected, and as such are strict subsets of their `available` counterparts.  If not provided, this object defaults to a clone of `available`, meaning that all options are selected by default.
```
parlarjb: Same for `selected` and `available`. I think a concrete example here would really help.

parlarjb: I think you need to be more explicit about being able to provide values to `selected`. It's not obvious from the description that you can choose the default selected values. I had to look at the example in `/docs` to see that you can do

```
 selected: {
            account: ['A']
        }
    };
```

The way it's written makes it sound more like it's a read-only private property that the directive uses to track stuff

megalord: Good point, I'll add an example in the README that explains how the three options work together (and clarify how the defaults work).

parlarjb:
The dropdown arrows feel too low to me. @glynnis do you have an opinion? 
screen shot 2015-04-30 at 4 22 15 pm

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+    properties: ['property1', 'property2']
+});
+var filteredArray = filter.applyTo(inputArray);
+```
+
+# Directives
+
+## rxSelectFilter
+
+Uses an instance of `SelectFilter` to create a set of `<rx-multi-select>`s that modify the instance object.
```
parlarjb: If this is generating a set of `<rx-multi-select>` elements, then the documentation for `rxSelectFilter` should come after that of `rxMultiSelect`. In this order, I don't understand what `rxSelectFilter` does at this point

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## rxMultiSelect
+
+A replacement for `<select multiple>` when space is an issue, such as in the header of a table.
+
+The options for the control can be specified by passing an array of strings to the `options` attribute (corresponding to the options' values) or using `<rx-select-option>`s. An 'All' option is automatically set as the first option for the dropdown, which allows all options to be toggled at once.
```
parlarjb: `can be specified by passing an array of strings to the `options` attribute` -> The "options attribute" of what?

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+If not provided, this object will be automatically populated the first time `apply` is called.
+- `selected`: An object that tracks which options are selected for a property. It has the same form as the `available` object, but the array values indicate which options are selected, and as such are strict subsets of their `available` counterparts.  If not provided, this object defaults to a clone of `available`, meaning that all options are selected by default.
+
+Instances of `SelectFilter` have an `applyTo` method, which applies the filter's internal state of selected options to the array. This will not often be called directly, instead used by the `Apply` filter (see below).
+
+```js
+var filter = SelectFilter.create({
+    properties: ['property1', 'property2']
+});
+var filteredArray = filter.applyTo(inputArray);
```
parlarjb: Could you show a before/after here, so we can see the result of applying the filter to the array?

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```html
+<rx-select-option value="DISABLED">Disabled</rx-select-option>
+```
+
+# Filters
+
+## Apply
+
+Merely applies an instance of `SelectFilter` to an input array.
```
parlarjb: Make note here to remind the reader that this is calling `applyTo` on the `SelectFilter` instance. That way, if they understand what `applyTo` does, they'll understand what the filter is doing.

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```html
+<tr ng-repeat="item in list | Apply:filter">
+```
+
+## titleize
+
+Converts a string to a title case, stripping out underscores and capitalizing words. This is used to generate the text for an `<rx-select-option>` with no content.
```
parlarjb: What do you think about putting this in `rxMisc`? There might be some value in having it there. It'll likely be "better exposed" if we do that. @CITguy thoughts?

@glynnis is this something you could see getting a mention in the style guide?

src/rxSelectFilter/docs/rxSelectFilter.html
```diff
@@ -0,0 +1,42 @@
+<!-- Sample HTML goes here as a live example of how the component can be used -->
+<div ng-controller="rxSelectFilterCtrl">
+    <h3 class="title">Multi-Select Input</h3>
```
parlarjb: Could you change this to "`rxMultiSelect` Input", or something similar, to point out that it's using `rxMultiSelect`? Maybe even "Multi-Select Input via rxMultiSelect"

src/rxSelectFilter/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+A multi-select dropdown for table filtering.
+
+# Services
+
+## SelectFilter
+
+This service exposes an object with single method, `create`, used to create instances of a `SelectFilter`.  It is configurable via three attributes:
+- `properties`: A list of the properties to create a filter control for.  Assuming the source data is an array of objects, a property is equivalent to an object's key.
+- `available`: An object that tracks which options are available for a property. It has the form:
```
parlarjb: Could you update the demo so it includes the use of `available`?

parlarjb:
UX question (@glynnis please jump in)

screen shot 2015-04-30 at 4 53 03 pm

^ This is the default view when the "Status" filter has nothing selected. It shows the word `None`, and the table is empty.

screen shot 2015-04-30 at 4 53 22 pm

screen shot 2015-04-30 at 4 53 16 pm

The two screenshots above show what it's like when all the dropdown options are selected for "Status". The "All" option in the dropdown gets selected, and the "Status" box shows "All" when you make the dropdown disappear.

In the two examples I've shown so far, "All" and "None" are special words, inserted by the directives. The directive also automatically adds "All" as the first item in the list, giving a quick way to select everything.

screen shot 2015-04-30 at 4 53 39 pm

screen shot 2015-04-30 at 4 57 29 pm

These two ^ screenshots show what happens when we want to filter on the value `IN_PROGRESS`. `IN_PROGRESS` is not a special term coming from the directive, instead it is automatically grabbed when the directive looks for all the possible `status` values for each row in the table. The directive has taken the text `IN_PROGRESS`, and turned it into the pretty "In Progress", for display in the dropdown options, and to display as the final selected. Because "In Progress" is the only value selected, it is what gets shown for "Status".

screen shot 2015-04-30 at 4 58 25 pm
screen shot 2015-04-30 at 4 58 47 pm

If we select "In Progress" _and_ "Transferred", then the directive displays the final "special" value, "Multiple". 

My UX concern is that the three special "Status" values, "All", "None", and "Multiple", are displayed in the same way as an actual status value from the table ("In Progress", "Transferred", etc.)

Can we separate that somehow? Even something as simple as putting quotation marks around the displayed status if it's one of "In Progress", "Transferred", etc., and not putting quotations if it's instead "All", "None" or "Multiple"

The place it could really cause issues is if some items in the table have an actual status of "All", "None" or "Multiple"

src/rxSelectFilter/README.md
megalord: It refers to the options attribute of the directive, e.g. `<rx-multi-select options="addDatacenters">`.  I'll make this more clear.

megalord:
Addressing the UX concern, do you think it would be better to have "In Progress, Transf..." instead of "Multiple"?  It would be pretty easy to cut off the text with an ellipsis, and that gets rid of one of the special values.

Another idea would be to indicate the number of items selected, such that
"None", "In Progress", "Multiple", "All" becomes "0 Selected", "In Progress", "X Selected", "All Selected"
The verbosity would prevent confusion between an actual value and a special value.

CITguy:
Why not just italicize the "special" values?

megalord:
The arrows were too low because the font size was smaller than that of a normal select input.  After fix:
screen shot 2015-05-01 at 11 10 10 am

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

parlarjb:
I like @CITguy's suggestion of doing italics for the "special" values

CITguy:
Based on the screenshot below, the "special" values are visually separated from the actual values? Can we do something similar (maybe just an `<hr />` between the special and actual values)?

*Screenshot from autofilter in LibreOffice Calc*
screen shot 2015-05-05 at 1 57 55 pm

megalord:
Good ideas, I'll put the special values in italics but leave the wording as is, and give the All option a visual distinction from the others. Thanks!

glynnis:
Here's what I think we should do:

1. Filter selections should always default to "All" on page load, so that any filtering is done by the user rather than by default (right now the "Account" is defaulting to "A"). @megalord Can you correct this in the demo?

2. If filter selections always default to "All", then any filtering done means that an action was taken by the user, and they thus are aware that the table has been filtered.

3. If no results match the filter settings the user has set, then a row should appear with a message, "No results match those criteria." This treatment should be like those we currently use for empty tables (see "Null patterns" here: http://rackerlabs.github.io/encore-ui/#/styleguide/tables). However, if the table is *actually* empty and there are no results available, we should instead display the standard "No [x] was found."

megalord:
@glynnis Thanks for taking a look.

1. I am hesitant to make this change because the app I wrote this for will not default to All - that's why I made sure it was possible to set the initial selection as a subset of the available options.  This will probably only be relevant for paginated or query-based apis, where changing the selection initiates a new request based on the change.  I can certainly make that distinction in the documentation if you find the use case acceptable.

2. Does this sentiment relate to the "special values" we discussed above? I'm not sure what you are trying to imply.

3. This I definitely agree with. I'll update the demo and put a reminder in the readme for this component.

megalord:
I just noticed that native select elements have a "pop out" effect when clicked, where the menu is displayed on top of the element and has a shadow.  I think that would give a better feel to this control (not as flat).  @glynnis do you have any thoughts?
screen shot 2015-05-06 at 4 13 42 pm
vs
screen shot 2015-05-06 at 4 19 39 pm

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

glynnis:
@parlarjb: @megalord and I met to speak about this directly. Here's what was decided:

- If only one item is selected from the dropdown, its label will display.
- If all but one item is selected, show "All except [x]" (like in Ticketing).
- If  > 1 but < n-1 items are selected, show "[#] selected".
- Change "All" to "Select All" with line dividing it from other options.
- We will move forward with some "pop-out"/shading effects that Jordan identified. These will blend in better with browser defaults that we use for other select elements.

@megalord has a use case where select boxes may not default to "all" on load, which the copy changes above will address as far as providing informative feedback to the user about what's selected and what isn't. I think all of these will address concerns I had about making this general enough to be used at the framework level + good UX for communicating what's selected.

Jordan will update, and @parlarjb and I can give a final LGTM before ready for release.

Thanks!

src/rxSelectFilter/rxSelectFilter.js
```diff
@@ -0,0 +1,231 @@
...
+                if (event.target === selectElement) {
+                    scope.listDisplayed = !scope.listDisplayed;
+                } else {
+                    event.stopPropagation();
+                }
+            };
+
+            var selectCtrl = controllers[0];
+            var ngModelCtrl = controllers[1];
+            selectCtrl.ngModelCtrl = ngModelCtrl;
```
parlarjb: Could you do this assignment after you create the `$render` method? `rxSelectOption` checks for the presence of this (`if (selectCtrl.ngModelCtrl) {`), and calls `$render`. This is technically a race condition (except for the fact that there are no threads in JavaScript, but, you know)

src/rxSelectFilter/rxSelectFilter.js
```diff
@@ -0,0 +1,231 @@
...
+            }, function (isSelected) {
+                scope.isSelected = isSelected;
+            });
+
+            selectCtrl.addOption(scope.value);
+
+            // ngModelCtrl will not be defined yet for the 'all' option
+            // when the options attribute is used for rxMultiSelect.
+            if (selectCtrl.ngModelCtrl) {
+                selectCtrl.ngModelCtrl.$render();
```
parlarjb: I don't want to hold up the PR for this, but in general, I would rather this logic be a bit different.

I'd prefer that instead of this `if` block, you just do a single call,

`selectCtrl.render()`

And then in the `rxMultiSelect` controller, do something like:

```
this.render = function () {
    if (this.ngModelCtrl) {
        this.ngModelCtrl.$render()
    }
}
```

i.e. `rxSelectOption` shouldn't haven't to concern itself with internal state of the `selectCtrl`, it should just have a `render()` method it can call, and let the `selectCtrl` worry about everything else.

parlarjb:
I mentioned some ways we could slightly restructure the `rxMultiSelect` controller to better hide its implementation details, but I'm cool with what you've got. Thanks for the work on this.

src/rxSelectFilter/rxSelectFilter.js
megalord: Yeah, that separation should exist. It looks like what you've suggested should work.  I'll see what I can do.

megalord:
screen shot 2015-05-07 at 3 30 01 pm
screen shot 2015-05-07 at 3 30 11 pm
screen shot 2015-05-07 at 3 30 20 pm
screen shot 2015-05-07 at 3 30 27 pm
screen shot 2015-05-07 at 3 30 37 pm
screen shot 2015-05-07 at 3 30 55 pm

glynnis:
@megalord @parlarjb From screenshots above, LGTM :+1:

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/922/gallery.html)

PR #923, opened 2015-04-28
docs(tables): add demo for table rows with a single action

### LINKED ISSUE #908: Design treatment for single-action rows in tables
An emerging need in Encore applications is the ability to associate a single action with a table row. In these instances, an action cog doesn't make sense since there's only one action per row (cog menu is an unnecessary extra click). However, we need to make a design decision about visual appearance for this.

Is it an action link? A button? Any other style guidelines or rules around it? Decide and document.

cc @megalord @ElementE

megalord:
Fixes #908 
screen shot 2015-04-28 at 10 27 24 am

The markup is a little more verbose than I would like, but pointer events aren't allowed on disabled elements, hence the need to wrap the button (and to use a button instead of a link).

### DIFF
--- a/demo/styleguide/tables.html
+++ b/demo/styleguide/tables.html
@@ -60,6 +60,15 @@
     </div>
     <rx-styleguide code-url="styleguide/tables/delete.html"></rx-styleguide>
 
+    <h3 class="title">Other single action</h3>
+    <div class="component-description clear">
+        <ul class="list">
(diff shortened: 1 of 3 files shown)


glynnis:
Design: LGTM

Thanks very much @megalord. Rockin' it as usual. :metal:

parlarjb:
LGTM

PR #926, opened 2015-04-29
Document design pattern for manual saving

### LINKED ISSUE #907: Create a "save" feature
An emerging need in Encore applications is the ability to manually or auto-save progress while completing forms.

We need a visual/interactive pattern for this, including decisions around:
- visual treatment
- placement
- interactive design
- whether manual vs. auto is an option

Also needs to be added as a component or reusable style in the framework. @megalord has some work already put in on this in an application, but not at the Framework level.

cc @megalord @ElementE @parlarjb

megalord:
Fixes #907 
screen shot 2015-04-29 at 10 42 34 am
screen shot 2015-04-29 at 10 42 43 am

### DIFF
--- a/demo/styleguide/forms/manual-saving.html
+++ b/demo/styleguide/forms/manual-saving.html
@@ -1,0 +1,12 @@
+<form name="form">
+    <rx-notification ng-if="form.$dirty" type="info">Changes made are not automatically saved. 
+        <button class="button submit xs" ng-disabled="form.$invalid">Save Now</button>
+    </rx-notification>
+    <span class="field-input">
+        <input type="text"
+            ng-model="text"
+            placeholder="no symbols allowed..."
+            ng-pattern="/^\w+$/"
+            ng-required>
+    </span>
+</form>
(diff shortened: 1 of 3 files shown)


demo/styleguide/forms/manual-saving.html
```diff
@@ -0,0 +1,12 @@
+<form name="form">
+    <rx-notification ng-if="form.$dirty" type="info">Changes made are not automatically saved. 
+        <button class="button submit xs" ng-disabled="form.$invalid">Save Now</button>
```
glynnis: It looks like from the screen shot that the button is getting the bold styles from `rx-notification`. Could we add an override so it matches text weight on normal buttons? Seems like the best way to do that cleanly would be to add an explicit font weight on the `button` element as part of this PR.

glynnis:
Other than font weight issue, LGTM

demo/styleguide/forms/manual-saving.html
megalord: Oh, good find.

megalord:
After change to font weight:
screen shot 2015-05-01 at 9 29 43 am
screen shot 2015-05-01 at 9 29 51 am

glynnis:
LGTM :+1:

CITguy:
FYI, light gray on light blue = no bueno for colorblindness

Issue #927, opened 2015-04-30
Add a callback for when a notification has been dismissed


halkon:
Allow to add a callback for when a notification has been dismissed (and even perhaps removed from the stack)

This will allow to create triggers on the application when certain notifications are dismissed, without having to watch the stack:

https://github.com/rackerlabs/encore-tq-ui/pull/304/files#r29440711

parlarjb:
Completed with https://github.com/rackerlabs/encore-ui/pull/933

PR #933, opened 2015-05-04
feat(rxNotify): Add ondismiss method to rxNotify

nickburns2006:


halkon:
Closes: https://github.com/rackerlabs/encore-ui/issues/927

nickburns2006:
@halkon, had to allow the `$interval` to invoke the `$apply` otherwise it wasn't firing the `ondismiss`.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/933/gallery.html)

parlarjb:
I like this idea. Did you contemplate doing it with a promise, instead of a callback? `.add` returns a `message` object right now, we could put a `$promise` attribute onto that object.

That's just a thought though. I'm fine with going the callback route if you think it'll get us everything we need.

nickburns2006:
@parlarjb @halkon and I discussed this, and while I understand the thoughts of doing it as a promise, it doesn't make sense in a "do this on dismissal" architecture. I felt like setting things up to be $promise-based altered the architecture/behavior too drastically.

parlarjb:
LGTM

CITguy:
LGTM

PR #936, opened 2015-05-05
feat(rxCheckbox): add rxCheckbox component

CITguy:
Per necessity of #884, this directive can stand alone for use elsewhere in the framework.

### DIFF
--- a/src/rxCheckbox/README.md
+++ b/src/rxCheckbox/README.md
@@ -1,0 +1,27 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+rxCheckbox is an **attribute directive** that wraps a native checkbox element in markup required for styling purposes.
+To use the directive, you can replace `type="checkbox"` with `rx-checkbox`. The directive is smart enough to set the correct input type.
+
+
+## Styling
+
+* Directive results in an **inline-block element**
+* You can style the output against decendents of the **`.rxCheckbox`** CSS class.
+
+
+## Show/Hide
+
+If you wish to show/hide your `rxCheckbox` element (and its label), we recommend placing the element (and its label) inside of a
+`<div>` or `<span>` wrapper, and performing the show/hide logic on the wrapper.
+
+```html
+<span ng-show="isShown">
+  <input rx-checkbox id="chkDemo" ng-model="chkDemo" />
+  <label for="chkDemo">Label for Demo Checkbox</label>
+</span>
+```
+
+It is highly recommended that you use `ng-show` and `ng-hide` for purposes of display logic. Because of the way that `ng-if` and `ng-switch`
+directives behave with scope, they may introduce unnecessary complexity in your code.
+
--- a/src/rxCheckbox/rxCheckbox.js
+++ b/src/rxCheckbox/rxCheckbox.js
@@ -1,0 +1,45 @@
+angular.module('encore.ui.rxCheckbox', [])
+.directive('rxCheckbox', function () {
+    return {
+        restrict: 'A',
+        scope: {
+            ngDisabled: '=?'
+        },
+        compile: function (tElement, tAttrs) {
+            // automatically set input type
+            tElement.attr('type', 'checkbox');
+            tAttrs.type = 'checkbox';
+
+            return function (scope, element, attrs) {
+                var disabledClass = 'rx-disabled';
+                var wrapper = angular.element('<div class="rxCheckbox"></div>');
+                var fakeCheckbox = '<div class="fake-checkbox">' +
+                        '<div class="tick fa fa-check"></div>' +
+                    '</div>';
+
+                element.wrap(wrapper);
+                element.after(fakeCheckbox);
+
+                // apply/remove disabled attribute so we can
+                // apply a CSS selector to style sibling elements
+                if (attrs.disabled) {
+                    wrapper.addClass(disabledClass);
+                }
+                if (_.has(attrs, 'ngDisabled')) {
+                    scope.$watch('ngDisabled', function (newVal) {
+                        if (newVal === true) {
+                            wrapper.addClass(disabledClass);
+                        } else {
+                            wrapper.removeClass(disabledClass);
+                        }
+                    });
+                }
+
+                // remove stylistic markup when element is destroyed
+                element.on('$destroy', function () {
+                    wrapper[0].remove();
+                });
+            };
+        }//compile
+    };
+});//rxCheckbox
--- a/src/rxCheckbox/rxCheckbox.less
+++ b/src/rxCheckbox/rxCheckbox.less
@@ -1,0 +1,97 @@
+/*
+ * rxCheckbox
+ */
+.rxCheckbox {
+  display: inline-block;
+  width: @rxCheckbox-size;
+  height: @rxCheckbox-size;
+  color: @rxCheckbox-color;
+  background-color: @rxCheckbox-background-color;
+  position: relative;
+
+  input[type="checkbox"],
+  .fake-checkbox {
+    position: absolute;
+    top: 0;
+    left: 0;
+    display: block;
+    margin: 0px;
+    padding: 0px;
+    border: 0px;
+    outline: 0px;
+  }
+
+  .fake-checkbox {
+    box-sizing: border-box;
+    z-index: 0;
+    width: 100%;
+    height: 100%;
+
+    // enabled unchecked
+    background-color: @rxCheckbox-background-color;
+    border: 1px solid @rxCheckbox-border-color;
+
+    // perfect center alignment for tick
+    display: flex;
+    flex-flow: row nowrap;
+    align-items: center;
+    justify-content: center;
+    color: transparent;
+
+    .tick {
+      flex: 0 0 auto;
+      font-size: @rxCheckbox-font-size;
+    }
+  }//.fake-checkbox
+
+  &.rx-disabled {
+    // disabled, unchecked
+    .fake-checkbox {
+      background: @inputBackground;
+    }
+
+    input[type="checkbox"] {
+      cursor: not-allowed;
+
+      // disabled, checked
+      &:checked ~ .fake-checkbox {
+        background-color: @inputBackgroundDisabled;
+        border: none;
+      }
+    }
+
+    & + label {
+      color: @inputColorDisabled;
+    }
+  }//.rx-disabled
+
+  input[type="checkbox"] {
+    z-index: 1;
+    height: 110%;
+    width: 110%;
+    opacity: 0;
+    cursor: pointer;
+
+    // enabled checked
+    &:checked ~ .fake-checkbox {
+      background-color: @rxCheckbox-background-color-selected;
+      color: @rxCheckbox-color;
+      border: none;
+    }
+
+    &.ng-invalid {
+      // enabled unchecked error
+      & ~ .fake-checkbox {
+        background-color: @rxCheckbox-background-color;
+        border: 2px solid @inputBackgroundError;
+      }
+
+      // enabled checked error
+      &:checked {
+        & ~ .fake-checkbox {
+          background-color: @inputBackgroundError;
+        }
+      }
+    }
+  }
+}//.rxCheckbox
--- a/src/rxCheckbox/rxCheckbox.page.js
+++ b/src/rxCheckbox/rxCheckbox.page.js
@@ -1,0 +1,143 @@
+/*jshint node:true*/
+var _ = require('lodash');
+var Page = require('astrolabe').Page;
+
+/**
+   @namespace
+ */
+var rxCheckbox = {
+    eleWrapper: {
+        get: function () {
+            return this.rootElement.element(by.xpath('..'));
+        }
+    },
+
+    eleFakeCheckbox: {
+        get: function () {
+            return this.eleWrapper.$('.fake-checkbox');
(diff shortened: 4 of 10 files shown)


parlarjb:
Could you include some screenshots?

CITguy:
screen shot 2015-05-06 at 9 56 33 am

CITguy:
Rebased

megalord:
I'm curious, why are the attributes being documented in the demo markup as opposed to via ngdocs?  That seems to be the standard from what I've seen.

src/rxCheckbox/rxCheckbox.less
```diff
@@ -0,0 +1,99 @@
+/*
+ * rxCheckbox
+ */
+.rxCheckbox {
+  display: inline-block;
+  width: @rxCheckbox-size; // OK
```
megalord: Can you remove all of the `// OK` comments?

CITguy: Ah, good call.

src/rxCheckbox/rxCheckbox.page.js
```diff
@@ -0,0 +1,103 @@
...
+                }
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @description make sure checkbox is deselected/unchecked
+     */
+    deselect: {
```
megalord: The `rxForm` version of a checkbox uses "unselect" instead of "deselect".  While I'm not partial to either, it may be confusing for anyone migrating to have to change their tests.  It would be easier if all they had to do is change
```javascript
rxForm.checkbox.initialize(...)
```
to
```javascript
rxCheckbox.initialize(...)
```

CITguy:
@megalord, I need to apply a styling to the first cell in the row so that the value doesn't wrap.  I cannot do this via markdown.  I addition, I wanted to keep non-demo markup to bhe bottom so that when a consumer looks at the "Markup" tab, they see the demo code first.

src/rxCheckbox/rxCheckbox.page.js
CITguy: Keep in mind, that this PR is to support upcoming changes to rxForm.

CITguy: Also, http://english.stackexchange.com/questions/18465/unselect-or-deselect

src/rxCheckbox/rxCheckbox.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        inject(function ($location, $rootScope, $compile) {
+            rootScope = $rootScope;
+            scope = $rootScope.$new();
+            compile = $compile;
+        });
+
+        el = helpers.createDirective(validTemplate, compile, scope);
+    });
+
+    /* reserved for potential future specs */
```
megalord: I'm not a big fan of having a spec file without any tests in it.  I say either add tests for the watch on `ng-disabled` or remove the file entirely.

CITguy: I only kept it in case someone pointed out something that required testing in this review. If there's nothing worth testing, then I can remove it. I'll defer to @Droogans on this one.

src/rxCheckbox/rxCheckbox.page.js
CITguy: Should rxForm use rxCheckbox, then `rxForm.checkbox.initialize(...)` will initialize an rxCheckbox item.

megalord:
@CITguy ngdocs doesn't use markdown - it's generated from the comments that are placed before each `.directive`, `.service`, etc.

src/rxCheckbox/rxCheckbox.page.js
CITguy: Sorry, I misunderstood.  With the upcoming changes to rxForm, there's going to be a considerable update. `rxForm.checkbox.initialize(...)` only applies to old rxForm functionality (using native checkboxes).  As developers transition to using the new rxForm functionality (using newly styled checkboxes), they would naturally have to transition to the new page objects as well.

megalord: Yeah, there will be some big changes to rxForm, and my concern is that something like this is not obvious and will be lost amidst the other changes, so the tests will break in unexpected places.  Perhaps as the larger change lands, it would be wise to have a migration guide where the big picture changes as well as details like this can be documented.

CITguy: I think we can revisit this topic in the rxForm updates, if needed.

CITguy:
Added "DO NOT MERGE" because functionality is needed to remove the wrapper markup on a destroy (thanks for pointing this out @megalord). This is needed when an `ng-if` or similar removes the `<input rx-checkbox ... />` element from the DOM.

src/rxCheckbox/rxCheckbox.page.js
Droogans: Since I'm the one responsible for this odd naming convention, I feel like I should at least explain why I went with unselect vs. deselect. At first, there was `select` for dropdowns and radio inputs, and `check` for checkboxes.

I didn't like the variation in language. It made it harder for me to remember what to call. So I switched everything over to `select`, and in the process, saw `uncheck` and thought...."should this be `unselect`? What about `deselect`?". After thinking about it, I went with something that worked everywhere (since there was no `decheck`), ending up settling on `unselect`, which may need to change.

TL;DR - I'm all for switching everything over to `deselect`, since there's no more variation between verbs.

Droogans: Also, @megalord, @CITguy and I discussed what'll likely happen to the page objects once all the rxForm components get broken out of it. I think we'll be using `rxForm` as a "require everything please" page object, so it'll keep backwards compatibility. It'll also make pulling in every form input type less lines of code (one `require` instead of three or four).

CITguy:
* Rebased
* removed `rxCheckbox.spec.js` (no functionality to test)
* added `$destroy` support
* updated midways and page object
* updated demo and documentation

CITguy:
rxcheckbox-local-20150508

CITguy:
*unchecked (invalid)*
screen shot 2015-05-08 at 11 25 46 am

*checked (valid), shows next checkbox via `ng-show` logic*
screen shot 2015-05-08 at 11 25 54 am

*both checked and valid*
screen shot 2015-05-08 at 11 26 03 am

CITguy:
**`$destroy` support**
*unchecked, second checkbox is present*
screen shot 2015-05-08 at 11 26 11 am

*checked, second checkbox (and stylistic markup) removed from DOM*
screen shot 2015-05-08 at 11 26 17 am

src/rxCheckbox/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Show/Hide
+
+If you wish to show/hide your `rxCheckbox` element (and its label), we recommend placing the element (and its label) inside of a
+`<div>` or `<span>` wrapper, and performing the show/hide logic on the wrapper.
+
+```html
+<span ng-show="isShown">
+  <input rx-checkbox
+         type="checkbox"
```
parlarjb: Could the `rx-checkbox` directive automatically set the `type="checkbox"`, in the `compile` step?

src/rxCheckbox/README.md
```diff
@@ -0,0 +1,28 @@
...
+         id="chkDemo"
+         ng-model="chkDemo" />
+  <label for="chkDemo">Label for Demo Checkbox</label>
+</span>
+```
+
+It is highly recommended that you use `ng-show` and `ng-hide` for purposes of display logic. Because of the way that `ng-if` and `ng-switch`
+directives behave with scope, they may introduce unnecessary complexity in your code.
```
parlarjb: :+1:

src/rxCheckbox/rxCheckbox.js
```diff
@@ -0,0 +1,34 @@
+angular.module('encore.ui.rxCheckbox', [])
+.directive('rxCheckbox', function () {
+    return {
+        restrict: 'A',
+        scope: {
+            ngDisabled: '=?'
```
parlarjb: This doesn't cause problems with the existing `ngDisabled` directive that ships with Angular?

src/rxCheckbox/README.md
CITguy: Unfortunately, I cannot use both `compile` and `link` in the DDO. Considering that I need to perform DOM manipulation, I need to keep `link`. I tried to automatically set the `type="checkbox"` in the link function, but for some reason, it breaks binding with the scope.

src/rxCheckbox/rxCheckbox.js
CITguy: Nope, just watches the value so that a `scope.$watch()` can add/remove a `disabled` attribute on the wrapper element.

src/rxCheckbox/README.md
CITguy: In addition, if for some reason that the directive doesn't bind properly to the DOM, the `type="checkbox"` will act as a fallback for graceful degradation.

src/rxCheckbox/rxCheckbox.js
```diff
@@ -0,0 +1,34 @@
...
+        },
+        link: function (scope, element) {
+            var wrapper = '<div class="rxCheckbox"></div>';
+            var fakeCheckbox = '<div class="fake-checkbox">' +
+                    '<div class="tick fa fa-check"></div>' +
+                '</div>';
+
+            element.wrap(wrapper);
+            element.after(fakeCheckbox);
+
```
parlarjb: Rather that doing `element.parent()` on every $watch call below, could you do `wrapper = angular.element(wrapper)`,
and then inside the `$watch` you an do `wrapper.attr('disabled', true);` and `wrapper.removeAttr('disabled')`.
Then in the `$destroy` you can also do `wrapper[0].remove();`

src/rxCheckbox/README.md
megalord: Just curious, are you opposed to returning the `link` function from `compile`?

CITguy: I had a chat with @parlarjb about it. He convinced me to convert to using `compile` and returning the function. I'll be submitting an update to the PR soon.

src/rxCheckbox/rxCheckbox.js
CITguy: ah, I will do that. much cleaner.

CITguy:
Updated per comments:

* using `compile` rather than `link` so that the directive automatically sets `type="checkbox"`
* applying logic to wrapper element rather than `element.parent()` repetition

Additionally, I reorganized some of the documentation to make it easier to consume and navigate. The examples still work same as seen in previous screenshots.

rxcheckbox-local-20150511-1

parlarjb:
Could you put back in `.spec.js` to ensure that `type="checkbox"` is being correctly added, as well as checking that your custom `ng-disabled` logic is being applied?

CITguy:
@parlarjb Midways have been updated to test `ng-disabled` and `type="checkbox"` functionality. I didn't feel like this warranted the need for a spec file as the directive isn't adding any functionality on `scope` or the controller. It merely manipulates the DOM based on data binding. @Droogans do you have any input on testing via unit tests versus midways?

parlarjb:
LGTM

CITguy:
Based on feedback for `rxRadio`, I need to add logic to support the native `disabled` attribute in addition to `ng-disabled`.

CITguy:
* added support for native `disabled` attribute

*Checkbox States - updated to include native `disabled` examples*
screen shot 2015-05-13 at 11 29 57 am

src/rxCheckbox/rxCheckbox.js
```diff
@@ -0,0 +1,45 @@
...
+                element.wrap(wrapper);
+                element.after(fakeCheckbox);
+
+                // apply/remove disabled attribute so we can
+                // apply a CSS selector to style sibling elements
+                if (element.attr('disabled')) {
+                    wrapper.addClass(disabledClass);
+                }
+                if (scope.ngDisabled) {
```
megalord: Should this be `attr.ngDisabled`? If the expression passed to `ng-disabled` is initially false, the watch should still be applied.

CITguy: I see what you're saying. I was trying to check for the presence of the variable, not its value. I'll get a quick update submitted.

src/rxCheckbox/rxCheckbox.js
```diff
@@ -0,0 +1,45 @@
...
+
+                element.wrap(wrapper);
+                element.after(fakeCheckbox);
+
+                // apply/remove disabled attribute so we can
+                // apply a CSS selector to style sibling elements
+                if (attrs.disabled) {
+                    wrapper.addClass(disabledClass);
+                }
+                if (angular.isDefined(scope.ngDisabled)) {
```
CITguy: @megalord, how's this look?

megalord: Well, this doesn't work for a value of `undefined`, which would work on a normal checkbox input.  At this point though, I feel I'm getting too nit-picky, so feel free to keep it as is unless someone else chimes in.

parlarjb: `if (_.has(attrs, 'ngDisabled')) {` would probably be the most correct way to do it

CITguy:
Updated to use `if (_.has(attrs, 'ngDisabled'))`

megalord:
LGTM

PR #946, opened 2015-05-12
feat(rxFeedback): Update API endpoint and add fields to feeback

julianbonilla:
Updates API endpoint to `/api/encore/feedback`.  User feedback now includes sso, current page and browser agent.

### DIFF
--- a/src/configs/configs.js
+++ b/src/configs/configs.js
@@ -14,7 +14,7 @@ angular.module('encore.ui.configs', [])
     { value: '/dev/xvdo', label: '/dev/xvdo' },
     { value: '/dev/xvdp', label: '/dev/xvdp' }
 ])
-.constant('feedbackApi', '/api/feedback')
+.constant('feedbackApi', '/api/encore/feedback')
 .provider('routesCdnPath', function () {
 
     this.customURL = null;
--- a/src/rxFeedback/README.md
+++ b/src/rxFeedback/README.md
@@ -2,6 +2,13 @@
 
 Component built to gather and send user feedback
 
+## Default Submission Function
+
+The `rxFeedback` component sends feedback to `/api/encore/feedback`, which routes feedback to `encoreui@lists`. 
+This endpoint also supports a `product` parameter `/api/encore/feedback/:product` for sending feedback to a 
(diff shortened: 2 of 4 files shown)


parlarjb:
@halkon @nickburns2006 Could you two please look at this, as TQ is the only thing currently using the API?

halkon:
From TQ's usage there is no issue with this.  

However some of this stuff seems to be stuff that can be retrieved from `Headers` and the `X-Auth-Token`.  

But then again, if we do try to make it as "API agnostic" then no biggie

:+1:

julianbonilla:
@halkon I like the idea of retrieving these from the headers on the API side. I'll try that.

julianbonilla:
@halkon Moved the current page and user agent to the API side.  Now just sending the id with the feedback.

parlarjb:
@julianbonilla is the endpoint for this currently up and running?

halkon:
it just got merged in yesterday, we can deploy it to `staging` `preprod` if anyone is feeling... daring

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/946/gallery.html)

parlarjb:
@julianbonilla Is the process for getting the product-specific endpoint documented anywhere? And if so, could you add a pointer to it here? Based on what we've got in this PR, teams won't know they can request their own product-specific endpoint

parlarjb:
LGTM

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/946/gallery.html)

chriscantu:
@julianbonilla @parlarjb Should this PR include the removing of the 'mailto' link approach?  Or should that be submitted in separate PR?  

https://github.com/rackerlabs/encore-ui/blob/master/src/rxFeedback/rxFeedback.js#L79

julianbonilla:
@chriscantu I think we should keep `mailto` as a fallback when the API fails.

chriscantu:
@julianbonilla Kool!

Legit

PR #938, opened 2015-05-07
Visual regression

Droogans:
This replaces the full-screenshot visual regression of our Wraith job with a more fine-grained solution.

It uses Selenium (Protractor, really) to grab screenshots *while* running end to end tests. It supports the ability to take a screenshot of just a certain part of the screen, at various resolutions.

Here's what the different visual regression notifications look like, as well as a preview of the diffs.

## Wraith

screen shot 2015-05-13 at 13 30 08

screen shot 2015-05-13 at 13 31 00

## This stuff

screen shot 2015-05-13 at 13 31 58

screen shot 2015-05-13 at 13 32 36

You can see the full diff in github of the new stuff: https://github.com/rackerlabs/encore-ui-screenshots/pull/6/files

## Advantages

 - Manipulate the page in Selenium, take a screenshot, continue
   - Allows for extremely complicated scenarios that might need visual checks along the way
 - Automatically names them in an easily diff-able way
   - `{{browser}}/{{path/to/test/files}}/{{mocha-test-name}}/{{resolution}}-{{selectorName}}.png`
 - Leverages github's built-in image diff tools
 - Add one line of code to get all of this

## Disadvantages

 - Can only store ~25,000 screenshots in a github repo
   - however, by the time we reach 25k screenshots, this limit will likely be raised
 - Can only display the first 25 images in a github pull request
   - you can still pull it down and view all of them yourself, though
 - @comeatmebro will build many screenshot reports if there are multiple commits pushed to the PR
   - can feasibly reach 25k screenshots this way in a short amount of time
 - The screenshot naming convention is very verbose

More is discussed in the [rackerlabs/encore-ui-screenshots readme file](https://github.com/rackerlabs/encore-ui-screenshots).

The library that does most of the heavy lifting is at https://www.npmjs.com/package/snappit-mocha-protractor

PR #951, opened 2015-05-13
Visual regression squashed

Droogans:
See https://github.com/rackerlabs/encore-ui/pull/938 for the original, unsquashed pull request.


Droogans:
On hold until I can figure out if/when an encore 2.0 release can go out.

TheNando:
Don't forget to document the 'Why?' and not just the 'How?'. Also, should vis-reg be its own test run, or is it included with the existing e3e suite?

Droogans:
@glynnis @parlarjb this is now running both Chrome and Firefox screenshots.

Droogans:
@parlarjb I just looked at how this is handling partial screenshots in Chrome. It's naming the file by the resolution of the screenshot, which is good...however, if the element renders outside of the visible area of the screen, it'll only grab the visible part. The name reflects this -- for instance, an 800x600 screenshot that only renders in the right half would get renamed to 400x600! This is bad because as the CSS is fixed to allow for the full 800x600 pixels to be visible, there would be a new screenshot created, instead of overwriting the old one.

This is going to require a pull request in the screenshot naming function, something I can get done soon and release and pull in later.

For now, let's stick with just Firefox.

parlarjb:
LGTM

Droogans:
Ok we have a new visual baseline let's see how travis handles pulling them down and running it again.

Droogans:
@parlarjb the last visual regression check appeared to work just fine. We should make a PR fixing the breadcrumbs and seeing how it looks with a diff.

TheNando:
This also looks good to me.

PR #959, opened 2015-05-15
feat(rxSelect): add rxSelect component

### LINKED ISSUE #896: create styleable <select> directive
In regards to [this comment](https://github.com/rackerlabs/encore-ui/issues/758#issuecomment-84167612), updates are required to get a fully styleable form select element.

CITguy:
Per necessity of rxForm rework, but this directive can stand alone for use elsewhere in the framework.

* `$destroy` support
* apply `rx-disabled` class to wrapper if disabled by either `disabled` or `ng-disabled`
* exercises added to test common functionality
* page objects exports `rxSelect`, `htmlSelectOption`, and `htmlSelect`
  * `htmlSelectOption` and `htmlSelect` provides page object support for
    non-styled elements
* rxCheckbox.page.js updated to export both `rxCheckbox` and `htmlCheckbox`
  * this is to provide page object support for non-styled checkboxes
* rxRadio.page.js updated to export both `rxRadio` and `htmlRadio`
  * this is to provide page object support for non-styled radios
* Lays groundwork for #884

Closes #896

### DIFF
--- a/src/rxCheckbox/rxCheckbox.page.js
+++ b/src/rxCheckbox/rxCheckbox.page.js
@@ -3,7 +3,97 @@ var _ = require('lodash');
 var Page = require('astrolabe').Page;
 
 /**
-   @namespace
+ * @namespace
+ */
+var htmlCheckbox = {
+    /**
+     * @function
+     * @returns {Boolean} Whether the checkbox is currently displayed.
+     */
+    isDisplayed: {
+        value: function () {
+            return this.rootElement.isDisplayed();
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether the checkbox is valid.
+     */
+    isValid: {
+        value: function () {
+            return this.rootElement.getAttribute('class').then(function (classes) {
+                return _.contains(classes.split(' '), 'ng-valid');
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the checkbox is currently selected
+     */
+    isSelected: {
+        value: function () {
+            return this.rootElement.isSelected();
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the checkbox is disabled
+     */
+    isDisabled: {
+        value: function () {
+            return this.rootElement.getAttribute('disabled').then(function (disabled) {
+                return (disabled ? true : false);
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the checkbox exists on the page
+     */
+    isPresent: {
+        value: function () {
+            return this.rootElement.isPresent();
+        }
+    },
+
+    /**
+     * @function
+     * @description Make sure checkbox is selected/checked
+     */
+    select: {
+        value: function () {
+            var checkbox = this.rootElement;
+            return this.isSelected().then(function (selected) {
+                if (!selected) {
+                    checkbox.click();
+                }
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @description make sure checkbox is deselected/unchecked
+     */
+    deselect: {
+        value: function () {
+            var checkbox = this.rootElement;
+            return this.isSelected().then(function (selected) {
+                if (selected) {
+                    checkbox.click();
+                }
+            });
+        }
+    }
+};
+
+/**
+ * @namespace
+ * @extends htmlCheckbox
  */
 var rxCheckbox = {
     eleWrapper: {
@@ -42,29 +132,7 @@ var rxCheckbox = {
 
     /**
      * @function
-     * @returns {Boolean} Whether the native checkbox element is valid.
-     */
-    isValid: {
-        value: function () {
-            return this.rootElement.getAttribute('class').then(function (classes) {
-                return _.contains(classes.split(' '), 'ng-valid');
-            });
-        }
-    },
-
-    /**
-       @function
-       @returns {Boolean} Whether or not the checkbox element is currently selected.
-    */
-    isSelected: {
-        value: function () {
-            return this.rootElement.isSelected();
-        }
-    },
-
-    /**
-     * @function
-     * @returns {Boolean} Whether or not the wrapper is disabled.
+     * @returns {Boolean} Whether or not the wrapper has expected disabled class name
      */
     isDisabled: {
         value: function () {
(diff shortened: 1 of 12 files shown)


CITguy:
Screenshot:
rxselect-local-20150515-1

CITguy:
### Validation

*Invalid*
screen shot 2015-05-15 at 3 29 30 pm

*Valid*
screen shot 2015-05-15 at 3 29 39 pm

CITguy:
### Show/Hide

*Visible*
screen shot 2015-05-15 at 3 29 45 pm

*Hidden*
screen shot 2015-05-15 at 3 29 51 pm

CITguy:
### Destroy

*Created*
screen shot 2015-05-15 at 3 30 06 pm

*Destroyed*
screen shot 2015-05-15 at 3 29 59 pm

src/rxSelect/rxSelect.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {Boolean} [ngDisabled=""] - Angular expression that evaluates to a Boolean
+ * @description This directive is to apply styling to native `<select>` elements
+ */
+.directive('rxSelect', function () {
+    return {
+        restrict: 'A',
+        scope: {
+            ngDisabled: '=?'
+        },
+        link: function (scope, element, attrs) {
```
megalord: The link function is almost exactly the same between `rxSelect`, `rxCheckbox`, and `rxRadio` - the only difference I see is the wrapper and fake elements.  What do you think about having a service that generates these methods?
```javascript
.directive('rxSelect', function (FakeInput) {
    var fakeSelect = '<div class="fake-select">' +
            '<div class="select-trigger">' +
                '<i class="fa fa-fw fa-caret-down"></i>' +
            '</div>' +
        '</div>';
    return {
        link: FakeInput.createLink('rxSelect', fakeSelect);
    };
});
```
This way, we don't have to keep all the components in sync if something ever comes up.  If only one component requires other functionality, you could do
```javascript
.directive('rxSelect', function (FakeInput) {
    return {
        link: function (scope) {
            FakeInput.createLink('rxSelect', fakeSelect).apply(this, arguments);
            // other custom stuff...
        }
    };
});
```

CITguy: I toyed around with the idea of creating a `WrappedElement` service that would take care of the boilerplate (wrapping, disabled suport, $destroy support, etc), but it's probably higher priority getting these elements functional and released before attempting to refactor

CITguy: I was toying around with this kind of interface.
```javascript
...
.directive('somethingWrapped', function (WrappedElement) {
  return new WrappedElement({
    wrapper: '<div class="rxSomethingWrapped"></div>',
    wrappedBefore: 'this goes before the anchor element within the wrapper',
    wrappedAfter: 'this goes after the anchor element within the wrapper',
    link: function (scope, element, attrs) {
      // this functionality is triggered after boilerplate linking happens
    }
  });
});
```

parlarjb: Agreed, I'd rather refactor once they're all in place.

glynnis:
Only one design note here: looks like the caret got darker. Current caret color is `#aaa`. @CITguy, could you use `#aaa` or a similar gray (if there's a better variable for instance) for the caret?

Other than that, LGTM.

CITguy:
@glynnis Updated screenshot

screen shot 2015-05-19 at 4 59 02 pm

lfielder:
@CITguy  This looks awesome. Nice work!

glynnis:
LGTM :cake: :icecream: :pizza:

src/rxSelect/rxSelect.js
```diff
@@ -0,0 +1,49 @@
...
+            var disabledClass = 'rx-disabled';
+            var wrapper = angular.element('<div class="rxSelect"></div>');
+            var fakeSelect = '<div class="fake-select">' +
+                    '<div class="select-trigger">' +
+                        '<i class="fa fa-fw fa-caret-down"></i>' +
+                    '</div>' +
+                '</div>';
```
parlarjb: I like creating these "constant" strings outside of the `link` function, as there's no need to re-create them whenever an `rxSelect` element is drawn, but we can save that for the refactor of this stuff

parlarjb:
LGTM

megalord:
I feel like I may be missing some other conversation that took place, but I'm confused as to why the page objects were changed to support the native html elements.  Is this for backwards compatibility?  If not, is there any direction for when to use the native elements instead of the directives?

src/rxCheckbox/rxCheckbox.page.js
```diff
@@ -116,13 +92,108 @@ var rxCheckbox = {
...
+     * @function
+     * @returns {Boolean} Whether or not the styled element exists on the page
+     */
+    isPresent: {
+        value: function () {
+            return this.eleFakeCheckbox.isPresent();
+        }
+    }
+};
+rxCheckbox = _.assign(htmlCheckbox, rxCheckbox);
```
megalord: Just a personal preference, but I'm usually not a big fan of reassigning a variable using itself - it's a little too sneaky. What about
```javascript
var rxCheckbox = _.assign(htmlCheckbox, {
    eleWrapper: {
    // etc...
});
```
?

src/rxSelect/docs/rxSelect.html
```diff
@@ -0,0 +1,151 @@
...
+      <br />
+      <select rx-select
+              id="selBaconPrep"
+              ng-model="baconPrep"
+              ng-required="true">
+        <option value="">I do not like bacon</option>
+        <option value="thin">Thin (light and crispy)</option>
+        <option value="medium">Medium (perfect balance of flavor)</option>
+        <option value="thick">Thick (borderline jerky)</option>
+        <option value="crumbled">Crumbled (great on salads)</option>
```
megalord: :+1: for creative demo data :)

src/rxSelect/rxSelect.page.js
```diff
@@ -0,0 +1,374 @@
...
+        value: function () {
+            return this.selectedOptionElements.count();
+        }
+    },
+
+    /**
+     * @function
+     * @returns {WebElement[]} List of all selected `<option>` elements in the dropdown.
+     */
+    selectedOptionElements: {
```
megalord: This returns an `ElementFinder`, not an `ElementArrayFinder`.  Both the property's name and description indicate otherwise.

src/rxSelect/rxSelect.page.js
```diff
@@ -0,0 +1,374 @@
...
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Integer} The number of selected `<option>` elements in the dropdown
+     */
+    selectedOptionCount: {
+        value: function () {
+            return this.selectedOptionElements.count();
```
megalord: I'm pretty sure this would throw a error.  See the next comment on this file.

src/rxCheckbox/rxCheckbox.page.js
CITguy: I tried that, but jsdoc didn't pick up the comments.

CITguy:
@megalord, no, there wasn't a discussion around it. I separated the page objects to allow for better composition of advanced components in the future *and* so I could better organize the page object functionality in the rxSelect component.  

If another component wanted to extend `<select>` functionality (for example), then the base  `htmlSelect` and `htmlSelectOption` page objects are available for that new component to extend and override methods it needs for its functionality.

Ideally, there should be a conversation about where these common page objects will live in the framework (my initial gut instinct would be `rxMisc`), but right now, I kept them with their closest framework relative.

* `htmlSelect`, `htmlSelectOption`, `rxSelect`
* `htmlRadio`, `rxRadio`
* `htmlCheckbox`, `rxCheckbox`

src/rxSelect/rxSelect.page.js
CITguy: Oh, I should use `$$`, right?

megalord:
Thanks @CITguy, I see where you're going with this now.
LGTM

src/rxSelect/rxSelect.page.js
CITguy: Updated to use `$$` per our conversation.

CITguy:
Tweaked page object, waiting on Travis

PR #961, opened 2015-05-15
feat(typeahead): show options when focused

### LINKED ISSUE #909: Typeahead styles/component
An emerging need in Encore applications is the ability to provide type-ahead tools for a select box or field. We need to create visual/interactive design for this, and then get it added to the Framework in a way that is flexible and appropriate for use in multiple areas across Encore.

@megalord has some work already put in on this in an application, but not at the Framework level. He's currently using https://select2.github.io/, but we need to vet it for use at the Framework level.

cc @megalord @ElementE @parlarjb

megalord:
Fixes #909 
typeahead

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -1,0 +1,46 @@
+angular.module('encore.ui.typeahead', ['ui.bootstrap'])
+.config(function ($provide) {
+    $provide.decorator('typeaheadDirective', function ($delegate, $filter) {
+        var typeahead = $delegate[0];
+        var link = typeahead.link;
+        var lowercase = $filter('lowercase');
+
+        typeahead.compile = function () {
+            return function (scope, element, attrs, ngModelCtrl) {
+                link.apply(this, arguments);
+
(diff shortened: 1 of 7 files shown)


glynnis:
Design: LGTM

src/typeahead/typeahead.js
```diff
@@ -0,0 +1,47 @@
...
+                            ngModelCtrl.$setViewValue(ngModelCtrl.$viewValue);
+                        });
+                    });
+
+                    scope.allowEmpty = function (actual, expected) {
+                        if (expected === EMPTY_KEY) {
+                            return true;
+                        }
+                        actual = $filter('lowercase')(actual);
+                        expected = $filter('lowercase')(expected);
```
parlarjb: Just a small optimization: Do something like `lower = $filter('lowercase')` outside of the `compile` function, and call out to that each time.

parlarjb:
LGTM! Very nice solution to extending angular-bootstrap's `typeahead`

CITguy:
LGTM

PR #962, opened 2015-05-15
fix(componentTemplate): update to .controller() syntax

CITguy:
Update component template to use the `.controller('myController', function ($scope) {...})` syntax. This is for angular 1.3+ support of new components.


parlarjb:
LGTM

PR #964, opened 2015-05-17
DO NOT MERGE: feat(rxMetadata): add rxMetadata component

### LINKED ISSUE #885: Create Metadata Directives
The current `metadata` CSS class on a definition list `<dl>` is no longer cutting the mustard.

* CSS hacks are required to correct the way that certain browsers select text on double-click (see #870 and #871)
* With the desire to move more toward flexbox-based layouts, a definition list does not contain enough markup to cleanly separate flex containers and flex items.
* Framework consumers have to update their markup to match framework recommendation (this should be handled by the framework via directives).

I propose we create `rxMeta` and `rxMetaItem` directives to be used as such:

```html
<rx-meta>
  <rx-meta-item label="foobar">
    out of working order; seriously, perhaps irreparably, damaged
  </rx-meta-item>
  ...
</rx-meta>
```

With the addition of these directives, we should deprecate the definition list styling (i.e. the `.metadata` CSS class) and steer developers to use the directives. The directives serve three purposes:

1. Simplify source markup for developers
2. Remove browser CSS hacks for proper functionality.
3. Framework has full control over layout. If we wanted to switch to a table-based layout, or back to a definition list, consumers wouldn't need to rework their markup once they're using the new directives.

This would initially be a MINOR level release as it would be adding the new directives, but won't break existing `dl.metadata` source code. It should be made very clear in the documentation, stakeholders mailing list, and the stakeholder meeting that the `.metadata` class **will be removed** on the *following* MAJOR-level release.

### LINKED ISSUE #939: Re-evaluate the usage of description lists <dl> for metadata
The metadata component is one of the most commonly and inconsistently used in Encore apps. Often, the styling is overwritten to change its default behavior. Commonly overwritten properties include: text-align, min-width, margin, clear.

Furthermore, clearing and floating elements becomes problematic when mixed with other inconsistently implemented app elements. In its current state, metadata will not play well with responsive layout attributes. It would be prudent to re-evaluate the intended design of this component and determine if a more appropriate html element should be used instead of dl or if it is a good candidate to make use of the flex module.
 
Examples of how metadata is being inconsistently used can be seen in Account Overview (right aligned dt), Billing Overview (left aligned dt), Billing Current Usage tab, and Support Details (right aligned but misaligned for some reason).

CITguy:
New directive to replace inadequate CSS styles for metadata display.

Closes #885
Closes #939

rxmetadata-local-20150518-2


CITguy:
## Example of custom styling

*Default Styling*
screen shot 2015-05-17 at 4 08 35 pm

*Custom Styling*
screen shot 2015-05-17 at 4 08 44 pm

CITguy:
## Example Show/Hide

*Hidden*
screen shot 2015-05-17 at 4 08 53 pm

*Shown*
screen shot 2015-05-17 at 4 09 02 pm

CITguy:
## Example Destroyed/Created

*Destroyed*
screen shot 2015-05-17 at 4 09 11 pm

*Created*
screen shot 2015-05-17 at 4 09 20 pm

glynnis:
Added "DO NOT MERGE" here, since this is a design issue that needs to be reviewed by the design team before we move forward.

glynnis:
@CITguy I discussed this with @ElementE @lfielder and Omar, and rather than offer many styles, we want to take a closer look at some of the root problems that exist with metadata to try to address them in a unified style and/or come up with alternative solutions when metadata isn't cutting the mustard.

I'm closing this for now, but please keep this branch around in case we want to draw on this work once we've had a chance to meet and make some decisions about the future of metadata. We're setting up a brainstorming session to try and tackle this in the near future. 

Thanks for your work on this! We may pull you in to consult from a tech perspective since you've been in the code recently and have a good understanding of some of the limitations of each style. :cake: :icecream: :hamburger:

CITguy:
The only style in the demo that is built directly into the directive CSS is the "Default" style (definition butted up against the label). The other two were for demonstration purposes to show that you have flexibility with the markup. If needed, I could just remove the additional styles from the demo to match what we have currently.

PR #965, opened 2015-05-18
chore(protractor): Upgrade to protractor 2.0

CITguy:
This is part of #951, but it should be mergeable on its own.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "marked": "^0.3.3",
     "mocha": "^2.1.0",
     "moment": "~2.9.0",
-    "protractor": "1.8.0",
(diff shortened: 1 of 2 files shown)


Droogans:
LGTM

Droogans:
@CITguy @parlarjb there's some breaking changes to rx-page-objects once the `rxForm` rewrite is all said and done (`unselect` becomes `deselect`). I'd be nice to get rid of other inconsistencies and force a bump to protractor 2.0 in a major release, if possible. There's a few projects that will likely struggle pushing past a protractor 2.0 upgrade, so it makes sense to mark rx-page-objects, and therefore encore-ui, as well.

PR #968, opened 2015-05-18
fix(rxRadio): fix broken styles

CITguy:
(Noticed this while playing around with `rxFormOptionTable`.)

*Before - Notice the white corners on the gray and green backgrounds*
screen shot 2015-05-18 at 1 09 00 pm

*After - no more white corners*
screen shot 2015-05-18 at 1 09 31 pm

### DIFF
--- a/src/rxRadio/docs/rxRadio.html
+++ b/src/rxRadio/docs/rxRadio.html
@@ -45,6 +45,7 @@
              ng-model="radCreateDestroy" />
       <label for="radDestroy">Destroyed</label>
     </span>
+    &nbsp;
     <span>
       <input rx-radio
              id="radCreated"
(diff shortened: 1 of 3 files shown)


parlarjb:
LGTM

glynnis:
LGTM

CITguy:
In addition to the background color, I noticed an issue with the way that the browser calculates the radius of the tick mark.

*Second tick is deformed*
screen shot 2015-05-20 at 10 52 14 am

This is due to quirks in the way that the browser calculates border radius based on percentages. The solution is to calculate the sizes explicitly rather than use percentages.

*After fix - perfectly circular*
screen shot 2015-05-20 at 11 16 43 am

glynnis:
Still LGTM :+1:  Nice catch!

PR #982, opened 2015-05-23
removed dead link

dman777:
Removed `["Good CSS has a one-to-many relationship with HTML"](https://speakerdeck.com/csswizardry/big-css-1?slide=22)` which gives `presentation not found`


parlarjb:
LGTM

PR #987, opened 2015-05-26
feat(rxApp): add responsive layout support

TheNando:
Currently, rxPage injects a nested template in place of the directive. Any content within the directive will be wrapped by this markup. Applying the 'layout' attributes to the rxPage markup will not produce the desired effect since the wrong container element will have the attribute.

This moves any layout attributes into the appropriate sub-element so that children respond appropriately.


TheNando:
retest this please

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/987/gallery.html)

parlarjb:
LGTM

demo/index.html
```diff
@@ -17,31 +17,6 @@
...
-                    };
-
-                FNOP.prototype = this.prototype;
-                fBound.prototype = new FNOP();
-
-                return fBound;
-            };
-        }
-    </script>
-
```
CITguy: What's the reason behind removing this? Does a recent version of PhantomJS now support `.bind`?

src/rxApp/rxApp.js
```diff
@@ -203,6 +203,24 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
             // Remove the title attribute, as it will cause a popup to appear when hovering over page content
             // @see https://github.com/rackerlabs/encore-ui/issues/251
             element.removeAttr('title');
+
+            function isLayoutAttrib (attr) {
+                return _.isString(attr.name) && attr.name.startsWith('layout');
```
CITguy: Do we need a String function to do this or can we achieve the same result using a regex? (`/^layout/`)

demo/index.html
TheNando: We already had this polyfill in the browser-helpers along with a new one I added. I just updated the grunt build to include that file with the tests so it doesn't need to be duplicated.

megalord:
LGTM

src/rxApp/rxApp.js
```diff
@@ -203,6 +203,24 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+            var pageBodyDiv = pageDiv.querySelector('.page-content');
+            function moveAttrib (attr) {
+                pageBodyDiv.setAttribute(attr, pageDiv.getAttribute(attr));
+                pageDiv.removeAttribute(attr);
+            }
+
+            // Relocate layout attributes
+            _(pageDiv.attributes)
+                .filter(isLayoutAttrib)
+                .pluck('name')
```
CITguy: Can this be done with a single `.map`? This implementation iterates over values in an attribute list twice, whereas a `.map()` would iterate once.

```javascript
_(pageDiv.attributes)
  .map(function (attr) {
    if (isLayoutAttrib(attr)) { // replaces .filter(isLayoutAttr)
      return attr.name; // replaces .pluck('name')
    }
  }).forEach(moveAttrib);
```

demo/index.html
parlarjb: This code was duplicated in `browser-helpers.js` and in `index.html`. Nando added a task to copy `browser-helpers.js` over so we could import that into `index.html`

src/rxApp/rxApp.js
CITguy: This can be done without monkey-patching the `String.prototype`.
```javascript
return _.isString(attr.name) && attr.name.match(/^layout/)
```

demo/index.html
CITguy: Gotcha

src/rxApp/rxApp.js
parlarjb: I like Nando's approach for the readability. The number of attributes is small, and `O(n)` vs. `O(2n)` comes out to the same thing. The total work done is the same, the second loop through is just "delaying" pulling out the `.name` attr

src/rxApp/rxApp.js
TheNando: I went with the startsWith because it was available in Chrome and Firefox, is cleaner to read, and performs faster (which is technically negligible). The only drawback was that I didn't realize PhantomJS wouldn't like it. The polyfills are only to keep Phantom happy. I could change it, since it looks like the methos is missing even in PhantomJS 2

src/rxApp/rxApp.js
```diff
@@ -203,6 +203,27 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+                    return;
+                }
+
+                pageBodyDiv.setAttribute(attr.name, pageDiv.getAttribute(attr.name));
+                pageDiv.removeAttribute(attr.name);
+            }
+
+            // Relocate all layout attributes
+            var i = pageDiv.attributes.length;
+            while (i--) {
```
TheNando: I would have liked to use the _.forEach here, but I discovered a bug. If the NamedNodeMap contains an element whose key string includes a hyphen (ex. layout-fill) then the function called when iterated receives an undefined instead of the collection attribute element. This method is nearly as clean and doesn't produce the bug.

src/rxApp/rxApp.js
```diff
@@ -203,6 +203,27 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+                moveLayoutAttrib(pageDiv.attributes[i]);
+            }
         },
```
TheNando: I would have liked to use the _.forEach here, but I discovered a bug. If the NamedNodeMap contains an element whose key string includes a hyphen (ex. layout-fill) then the function called when iterated receives an undefined instead of the collection attribute element. This method is nearly as clean and doesn't produce the bug.

src/rxApp/rxApp.js
```diff
@@ -203,6 +203,27 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+
+                pageBodyDiv.setAttribute(attr.name, pageDiv.getAttribute(attr.name));
+                pageDiv.removeAttribute(attr.name);
+            }
+
+            // Relocate all layout attributes
+            var i = pageDiv.attributes.length;
+            while (i--) {
+                moveLayoutAttrib(pageDiv.attributes[i]);
+            }
```
TheNando: I would have liked to use the _.forEach here, but I discovered a bug. If the NamedNodeMap contains an element whose key string includes a hyphen (ex. layout-fill) then the function called when iterated receives an undefined instead of the collection attribute element. This method is nearly as clean and doesn't produce the bug.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/987/gallery.html)

CITguy:
LGTM

PR #996, opened 2015-05-27
feat(rxFieldName): add rxFieldName component

CITguy:
* Lays groundwork for #884

rxfieldname-local-20150527-1

### DIFF
--- a/src/rxFieldName/rxFieldName.js
+++ b/src/rxFieldName/rxFieldName.js
@@ -1,0 +1,11 @@
+angular.module('encore.ui.rxFieldName', [])
+.directive('rxFieldName', function () {
+    return {
+        restrict: 'E',
+        transclude: true,
+        scope: {
+            ngRequired: '=?'
+        },
+        templateUrl: 'templates/rxFieldName.html'
+    };
+});
--- a/src/rxFieldName/rxFieldName.less
+++ b/src/rxFieldName/rxFieldName.less
@@ -1,0 +1,21 @@
+/*
+ * rxFieldName
+ */
+rx-field-name {
+  .box-sizing(border-box);
+  display: block;
+  margin: 0px;
+  padding: 0px;
+  color: @rxFieldName-color;
+
+  .wrapper {
+    .box-sizing(border-box);
+    margin: 0px;
+    padding: 0px;
+
+    .required-symbol {
+      font-weight: 900;
+      color: @rxFieldName-symbol-color;
+    }
+  }
+}//rx-field-name
--- a/src/rxFieldName/README.md
+++ b/src/rxFieldName/README.md
@@ -1,0 +1,7 @@
+[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+
+rxFieldName is an **element directive** that provides a standardized UI for displaying form field names.
+
+## Styling
+* Directive results in a **block element**
+* You can style the output against the decendents of the `rx-field-name` element.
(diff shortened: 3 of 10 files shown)


CITguy:
*Required*
screen shot 2015-05-27 at 5 25 53 pm

*Optional*
screen shot 2015-05-27 at 5 25 45 pm

megalord:
Quite frankly, I'm not convinced that this really needs to be a directive at all - it seems like a few CSS rules would do just fine.
```html
<span class="rx-field-name-content required">Checkbox Input Field</span>
<!-- or -->
<span class="rx-field-name-content" ng-class="{required: fieldRequired}">Checkbox Input Field</span>
```
I'm guessing the asterisk could be applied with `::before`.

CITguy:
@megalord, with the way it is currently set up, yes, it could be done with a `::before` but that leaves it brittle to future modification (what if we wanted to use a FontAwesome icon?).  

The directive is still necessary for upcoming work.  The logic isn't here yet, but an upcoming PR will be adding additional logic to the DDO for hierarchical validation.  For style/layout purposes, `<rx-field-name>` must be an immediate child of `<rx-field>`. If it's placed anywhere else, a console error should be thrown to denote that it is expecting its parent to be a `<rx-field>`.

megalord:
Had a conversation with @CITguy offline.  I'm still concerned that this is flirting the line of bloat, but it sounds like there is a desire to increase standardization of forms, which will be done with these hierarchy validations.

src/rxFieldName/rxFieldName.js
```diff
@@ -0,0 +1,11 @@
+angular.module('encore.ui.rxFieldName', [])
+.directive('rxFieldName', function () {
+    return {
+        restrict: 'E',
+        transclude: true,
+        scope: {
+            ngRequired: '=?'
```
megalord: This seems to add some duplication when used in conjunction with an input.
```html
<rx-field>
  <rx-field-name ng-required="fieldRequired">Name</rx-field-name>
  <rx-field-content>
    <input type="text" ng-required="fieldRequired" />
  </rx-field-content>
</rx-field>
```
Giving `rx-field` (which I'm aware doesn't exist yet) a controller so that `rx-field-name` can communicate with `ng-required` on `<input />` would work, but that would definitely be some extra effort.

CITguy: Can you move this question to issue #884?  It pertains to the overhaul as a whole and should be discussed there.

parlarjb: Either a controller in `rx-field`, or a sibling controller in `rx-field-content`, that `rx-field-name` can communicate with. I agree that. Having the duplicated `ng-required` attributes is sticking out to me.

CITguy: Please see https://github.com/rackerlabs/encore-ui/issues/884#issuecomment-107741918

parlarjb:
I think I'm ok with the `ng-required` for now. Ideally we'll find a way to remove it, but we've got plenty of time before this stuff gets used heavily, and practical usage scenarios might show us a better way.

CITguy:
Rebased PR

src/rxFieldName/rxFieldName.less
```diff
@@ -0,0 +1,21 @@
...
+  .wrapper {
+    .box-sizing(border-box);
+    margin: 0px;
+    padding: 0px;
+
+    .required-symbol {
+      font-weight: 900;
+      color: @rxFieldName-symbol-color;
+    }
+  }//.rxFieldName
```
megalord: This has changed since first written, I believe.

megalord:
LGTM

parlarjb:
LGTM

PR #998, opened 2015-05-28
fix(ng1dot3): modify for angular 1.3 support

CITguy:
This is to provide support for angular 1.3+

Though we're not entirely sure what caused the issue, it seems that the `wrapper` variable used to wrap the native form elements seems to be disconnected from the DOM in angular 1.3.0 and above.  The workaround is to use `element.parent()` instead of `wrapper`.

### DIFF
--- a/src/rxCheckbox/rxCheckbox.js
+++ b/src/rxCheckbox/rxCheckbox.js
@@ -12,32 +12,34 @@ angular.module('encore.ui.rxCheckbox', [])
 
             return function (scope, element, attrs) {
                 var disabledClass = 'rx-disabled';
-                var wrapper = angular.element('<div class="rxCheckbox"></div>');
+                var wrapper = '<div class="rxCheckbox"></div>';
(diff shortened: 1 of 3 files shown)


parlarjb:
Instead of calling `element.parent()` multiple times, can you do `var parent = element.parent()` one time, and use that?

CITguy:
@parlarjb: Yeah, that seems to work in both 1.2 and 1.3. I'll make that change.

TheNando:
LGTM

CITguy:
@parlarjb, this PR has been upaded to use `var parent = element.parent()` as suggested.

parlarjb:
LGTM

parlarjb:
@CITguy this needs a rebase

CITguy:
balls

CITguy:
rebased, waiting on Travis

CITguy:
rebased with Travis corrections

CITguy:
re-rebased

PR #999, opened 2015-05-28
feat(rxApp): Add LDAP filtering in the nav

### LINKED ISSUE #958: LDAP filtering in the nav
Ideally, we should be able to specify in the nav JSON file that certain LDAP roles are required for an item to be visible in the nav. This would help simplify the nav for typical users, as it would prevent _everything_ from being visible.

parlarjb:
Closes #958

This adds the ability to filter individual items in the nav
by requiring certain LDAP roles. If a logged-in user doesn't
have the necessary roles, then the item won't appear in the nav
for them.

Note: Before we can really start to use this, all Encore products will have to be running whatever version of the framework incorporates this change. Otherwise all nav items will be displayed when the user is in a non-updated product.

Note that I haven't explicitly documented how to add these roles to the nav. I don't know yet where the right place is for that. Likely better to increase documentation in the encore-ui-nav repo.

### DIFF
--- a/src/rxApp/rxApp.js
+++ b/src/rxApp/rxApp.js
@@ -1,5 +1,5 @@
 angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnvironment', 'ngSanitize',
-    'ngRoute', 'cfp.hotkeys', 'encore.ui.rxSession', 'encore.ui.rxLocalStorage'])
+    'ngRoute', 'cfp.hotkeys', 'encore.ui.rxSession', 'encore.ui.rxLocalStorage', 'encore.ui.rxPermission'])
 /**
 * @ngdoc service
 * @name encore.ui.rxApp:encoreRoutes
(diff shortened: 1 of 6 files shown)


src/rxPermission/rxPermission.js
```diff
@@ -16,24 +16,58 @@ angular.module('encore.ui.rxPermission', ['encore.ui.rxSession'])
...
+            return _.pluck(permissionSvc.getRoles(), 'name');
+        };
+
+        /*
+         * @description Takes a function and a list of roles, and returns the
+         * result of calling that function with `roles`, and comparing to userRoles()
+         * @param {function} fn - Comparison function to use. _.any, _.all, etc.
+         * @param {array} roles - List of desired roles
+         */
+        var checkRoles = function (fn, roles) {
```
CITguy: This seems backward given the typical callback argument pattern. It might be better to have it as 

```javascript
var checkRoles = (roles, fn) { ... }
```
This way the list is before the callback, which is more commonplace in the JS community.

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/999/gallery.html)

src/rxPermission/rxPermission.js
parlarjb: Good call, thanks

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/999/gallery.html)

src/rxApp/rxApp.js
```diff
@@ -301,16 +301,38 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
 
-            $scope.isVisible = function (visibility) {
+            var roleCheck = function (roles) {
+                if (_.isUndefined(roles)) {
+                    return true;
+                }
+
+                if (!_.isUndefined(roles.any)) {
+                    return Permission.hasRole(roles.any);
+                } else if (!_.isUndefined(roles.all)) {
```
TheNando: You don't strictly need the 'else' since each clause is returning.

src/rxApp/rxApp.spec.js
```diff
@@ -725,6 +774,39 @@ describe('rxAppNavItem', function () {
 
         expect(childHeader.textContent).to.equal(menuItem.childHeader);
     });
+
+    describe('Roles visibility', function () {
```
TheNando: Lines breaks @ 783 and 803

TheNando:
LGTM

src/rxApp/rxApp.spec.js
parlarjb: Thanks, fixed

comeatmebro:
[Wraith UI diff](http://eb4d4842ea70055541c8-b6a3ba129b5eb76af13f4d17e4c1a53b.r51.cf5.rackcdn.com/999/gallery.html)

CITguy:
LGTM

PR #1001, opened 2015-06-02
feat(rxSelectFilter): Leverage new rxCheckbox and rxSelect components

### LINKED ISSUE #990: Update rxMultiSelect to use rxCheckbox and rxSelect

megalord:
Fixes #990 
Now, this input looks the same as `rx-select` and can be disabled.  Also, preview text (e.g. "2 Selected") that exceeds the width of the control will now be clipped instead of wrapping to a new line.
screen shot 2015-06-02 at 9 31 39 am
screen shot 2015-06-02 at 9 31 26 am

### DIFF
--- a/src/rxSelect/rxSelect.js
+++ b/src/rxSelect/rxSelect.js
@@ -27,7 +27,7 @@ angular.module('encore.ui.rxSelect', [])
 
             // apply/remove disabled class so we have the ability to
             // apply a CSS selector for purposes of style sibling elements
-            if (attrs.disabled) {
+            if (_.has(attrs, 'disabled')) {
                 wrapper.addClass(disabledClass);
             }
             if (_.has(attrs, 'ngDisabled')) {
--- a/src/rxSelect/rxSelect.less
+++ b/src/rxSelect/rxSelect.less
@@ -4,7 +4,6 @@
 .rxSelect {
   .box-sizing(border-box);
   display: block;
-  overflow: hidden;
   position: relative;
   color: @rxSelect-color;
   background: @rxSelect-background;
@@ -12,7 +11,7 @@
 
   // Element should be resizable via wrapper alone
   min-width: @rxSelect-min-width;
-  min-height: @rxSelect-min-height;
+  height: @rxSelect-height;
 
   &.rx-disabled {
     color: @inputColorDisabled;
@@ -21,6 +20,7 @@
 
   // position elements in wrapper
   select,
+  rx-multi-select,
   .fake-select {
     .box-sizing(border-box);
     position: absolute;
@@ -30,9 +30,9 @@
     height: 100%;
   }
 
-  select {
+  select,
+  rx-multi-select {
     z-index: 0;
-    padding: 5px;
     cursor: pointer;
 
     // reset native styling
--- a/src/rxSelectFilter/README.md
+++ b/src/rxSelectFilter/README.md
@@ -165,4 +165,4 @@ Merely calls the `applyTo()` method of a `SelectFilter` instance to an input arr
 
 Remember to adhere to the Null Pattern for [tables](#/styleguide/tables) in Encore UI.  When there are no results to populate the table, the table content should say "No [x] were found."  However, if the table is empty because of the applied filters, the message should instead say "No results match those criteria."  This has been included in the demo below for an example implemtation.
 
-The developer is reponsible for ensuring that the multi-select input is wide enough to show the preview text. Especially for the "All except [x]" case, the text could wrap if the proper width is not set.  To facilitate this, a class is provided that adds a '-filter' suffix to the filtered property.  See the demo markup for an example.
+The developer is reponsible for ensuring that the multi-select input is wide enough to show the preview text. Especially for the "All except [x]" case, the text could clip if the proper width is not set.  To facilitate this, a class is provided that adds a '-filter' suffix to the filtered property.  See the demo markup for an example.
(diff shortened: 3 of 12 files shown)


src/rxSelect/rxSelect.js
```diff
@@ -27,7 +27,7 @@ angular.module('encore.ui.rxSelect', [])
 
             // apply/remove disabled class so we have the ability to
             // apply a CSS selector for purposes of style sibling elements
-            if (attrs.disabled) {
+            if (_.has(attrs, 'disabled')) {
```
parlarjb: These aren't semantically equivalent. Was it wrong before?

src/rxSelect/rxSelect.less
```diff
@@ -4,15 +4,14 @@
...
-  overflow: hidden;
   position: relative;
   color: @rxSelect-color;
   background: @rxSelect-background;
   .border-radius(@rxSelect-border-radius);
 
   // Element should be resizable via wrapper alone
-  min-width: @rxSelect-min-width;
-  min-height: @rxSelect-min-height;
+  width: @rxSelect-width;
```
CITguy: Please revert to `min-width` and `@rxSelect-min-width`. If you need to set an explicit width, please set `width` on the `.rx-select-filter` wrapper.

-----

`rx-select` was designed to take the full width of its parent, because there's no way to know how a `rx-select` will be used in context with other markup. It needs a minimum width so that it can:

1. grow to the size it needs, but
2. retain a sane minimum width for usability

*(fyi: This is part of the master plan concerning the `rxForm` rework.)*

src/rxSelect/rxSelect.less
```diff
@@ -21,6 +20,7 @@
 
   // position elements in wrapper
   select,
+  rx-multi-select,
```
CITguy: Component styles should remain in their respective LESS file. Please define `rx-multi-select` styles in `src/rxSelectFilter/rxSelectFilter.less`.

src/rxSelect/rxSelect.less
```diff
@@ -4,15 +4,14 @@
 .rxSelect {
   .box-sizing(border-box);
   display: block;
-  overflow: hidden;
```
CITguy: Per our conversation, this is fine to remove. *(It is a remnant of an old prototype.)*

src/rxSelect/rxSelect.less
megalord: The way it was prevented a `width` that was less than `min-width`, but I'd like to be able to create multi-selects that are smaller than 250px.  Setting a `width` on the wrapper will not work.  Why are all select inputs being given a minimum width?

src/rxSelect/rxSelect.less
```diff
@@ -4,15 +4,14 @@
...
   position: relative;
   color: @rxSelect-color;
   background: @rxSelect-background;
   .border-radius(@rxSelect-border-radius);
 
   // Element should be resizable via wrapper alone
-  min-width: @rxSelect-min-width;
-  min-height: @rxSelect-min-height;
+  width: @rxSelect-width;
+  height: @rxSelect-height;
```
CITguy: This is okay to keep. There's not much vertical growth happening in form elements anyway.

src/rxSelect/rxSelect.less
CITguy: How small are we talking?

src/rxSelect/rxSelect.js
megalord: The old way works fine for `select`, but not for `rx-multi-select`.  When the browser encounters the `disabled` attribute on a `select` or `input`, it will set the element's `disabled` property to `true`.  However, I want add support for `rx-multi-select`.  The other alternative is to put
```javascript
attrs.disabled = _.has(attrs, 'disabled');
```
in the linking function of `rx-multi-select`, but I felt this was simpler.

As far as I know, there isn't any way to put the `disabled` attribute on an element and also set it to `false`, so merely checking for it's presence is still valid.

src/rxSelect/rxSelect.less
megalord: I added to the selector here to avoid duplication.  The way I see it, I'm adding support for `rx-multi-select` to `rx-select`.  I'd really rather not just copy half of `rxSelect.less` into `rxSelectFilter.less`; would you rather see a mixin used?

src/rxSelect/rxSelect.js
parlarjb: Ok. Out of curiosity, what value does `attrs.disabled` have when used with `rx-multi-select`?

megalord: It's an empty string - it basically turns into `disabled=""`.

src/rxSelect/rxSelect.less
megalord: I personally have a need for `130px` or so, but looking at the current [ticketing app](https://staging.encore.rackspace.com/ticketing/account/323676), there are some inputs with `90px`.

As per our discussion, I'll change the `min-width` to `100px`.

src/rxSelect/rxSelect.less
CITguy: :ok: After thinking about it a bit, you're right. This won't affect anything, currently. But as soon as we need to do anything special with `rx-multi-select` in regards to `rx-select`, we'll need to split the styles out.

src/rxSelect/rxSelect.js
CITguy: FYI: this is because the boolean `disabled` attribute is designed to be placed only on HTML form controls. On a `select` it is a known boolean attribute and the browser returns the proper boolean value when calling `attrs.disabled`. Because `<rx-multi-select>` is a custom element that doesn't extend from one of these form controls, `disabled` doesn't register as a boolean attribute and you get an empty string as the value.

megalord:
Added a little padding between the options
screen shot 2015-06-02 at 3 59 21 pm

CITguy:
LGTM

parlarjb:
LGTM

Issue #984, opened 2015-05-26
rxSelectFilter does not populate "All except ..." properly on initialization


cmartinez1499:
When initializing an rxSelectFilter through a controller and selecting all but one of the available options, the UI should display "All except [excluded item]", but instead displays "All except". This only happens during initialization.
![screen shot 2015-05-26 at 1 08 02 pm](https://cloud.githubusercontent.com/assets/1155834/7819755/6ffce3da-03a9-11e5-9e61-b1d8a309d5b9.png)
![screen shot 2015-05-26 at 1 07 48 pm](https://cloud.githubusercontent.com/assets/1155834/7819758/7354aeb4-03a9-11e5-932a-72cacbb05b8e.png)

megalord:
Yeah, I noticed this as well the other day, but it isn't partial to the "All except" case.  When using dynamic options for the multi-select (e.g. using the options attribute instead of rxSelectOption), the label in the dropdown isn't rendered before the component grabs it for the preview text.

PR #1006, opened 2015-06-03
fix(rxSelectFilter): correctly render preselected options

megalord:
Fixex #984 
The preview text requires that the DOM has been updated, as it uses an option's label in some cases.  Also, the component internally tracks the list of available options, and this list is used in the preview text logic.  However, items weren't being removed if the option element was removed.


parlarjb:
What was the reason for putting it into `$evalAsync`?

megalord:
`$evalAsync` is like `$timeout`, but instead of creating a new digest cycle, it simply waits until the next "tick" of the current cycle.  The delay is needed because the preview text is (sometimes) pulled from the DOM, but the text isn't in the DOM on the current "tick".

parlarjb:
LGTM

CITguy:
LGTM

PR #1010, opened 2015-06-03
docs(styleguide): Document format for empty or null values

### LINKED ISSUE #1008: Document style for empty/null values
See https://jira.rax.io/browse/UXENCORE-197

Adding this as a story on the Framework side.

While related to https://github.com/rackerlabs/encore-ui/issues/972, this issue is separate in that it only covers the need to document in the style guide this preference for "N/A" (rather than how it can be cleanly implemented by devs).

glynnis:
Closes #1008

### DIFF
--- a/demo/styleguide/basics.html
+++ b/demo/styleguide/basics.html
@@ -141,6 +141,14 @@
 
     <rx-styleguide code-url="styleguide/basics/metadata.html"></rx-styleguide>
 
+    <h3 class="title clear" id="no-data">When There is No Data to Display</h3>
+
+    <ul class="list">
+        <li>In table cells, metadata lists, or other displays of data, <code>N/A</code> should be used as an indicator to the user that there is no relevant data to display.</li>
+        <li>Do not use <code>–</code>, <code>Empty</code>, <code>null</code>, or no indicator to communicate this, as there may be situations where those are valid inputs or data to display to a user depending on the context.</li>
+        <li>Note that a separate style exists to communicate <a href="/#/styleguide/basics#wells">when data isn't present due to permissions</a>, or when an entire table is empty (see "Null Patterns" section of <a href="/#/styleguide/tables">Tables style guide</a>).
+    </ul>
+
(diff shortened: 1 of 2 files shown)


demo/styleguide/basics.html
```diff
@@ -134,6 +134,14 @@ <h3 class="title clear" id="descriptions-metadata">Descriptions / Metadata</h3>
 
     <rx-styleguide code-url="styleguide/basics/metadata.html"></rx-styleguide>
 
+    <h3 class="title clear" id="no-data">When There is No Data to Display</h3>
+
+    <ul class="list">
+        <li>In table cells, metadata lists, or other displays of data, <code>N/A</code> should be used as an indicator to the user that there is no relevant data to display.</li>
+        <li>Do not use <code>–</code>, <code>Empty</code>, <code>null</code>, or no indicator to communicate this, as there may be situations where those are valid inputs or data to display to a user depending on the context.</li>
+        <li>Note that a separate style exists to communicate <a href="/#/styleguide/basics#wells">when data isn't present due to permissions</a>, or when an entire table is empty (see "Null Patterns" section of <a href="/#/styleguide/tables">Tables style guide</a>).
```
glynnis: @parlarjb This anchor link doesn't work no matter what way I format it. I can't remember -- is that just part of how the IDs work/are set up? Or does it need the full URL? I seem to remember there was some trick to this but I couldn't figure it out.

parlarjb: I checked out the branch and tried it, and both the `wells` and `tables` links seem to work. What behaviour are you seeing?

glynnis: Just checked this out again and it's working for me now too? :confused:

Droogans:
@glynnis @parlarjb The failure in the end to end tests appears to be from glitchy hover actions in Selenium.

I'll check out the branch locally, if I can't reproduce then LGTM. There's no good reason that documentation (mostly) changes would alter the behavior of a component.

PR #1012, opened 2015-06-04
Visual regression reject forks

Droogans:


CITguy:
lgtm - if travis agrees

Droogans:
@parlarjb I've simply suppressed any and all failures around visual regression. This makes sense given that it's a "nice to have" and less critical than other parts of the build.

However, should there be a legitimate issue with the visual regression system, you'll never know what went wrong without spending a lot of time guessing and testing with Travis CI. It's difficult isolating problems when you're waiting up to 30 minutes between attempts.

I've [created a question on stackoverflow](http://stackoverflow.com/questions/30649196/skipping-parts-of-a-travis-job-if-building-from-a-fork) that seeks an answer to my fundamental problem. So far, no luck.

Droogans:
@CITguy are you looking for me to squash these two together?

```
test(visualRegression): Reject if building from fork         d2b561c
fix(travis): Don't fail the build for screenshots            5592aca
```

The rest of the commits are unrelated but still important. I can squash everything but it'll be a bit more of a generic commit message.

CITguy:
I was under the assumption that all PRs should be squashed. If @parlarjb is okay with keeping these as separate commits, I'm okay with it.

parlarjb:
I prefer full squashes, but I'm ok with not doing a full squash if the commits do separate things, and it might be useful in the future to see the history.

PR #1020, opened 2015-06-04
Increase Travis' stability in end to end tests

Droogans:


Droogans:
@parlarjb Please merge this after reviewing and discussing it with the relevant parties. I am certain it will restore order to Travis test runs in the future.

Droogans:
I think by switching away from sharding the tests between two firefox instances back to just one for everything, we may reduce the chances of the tooltips tests failing for no reason.

Droogans:
Alright, squashing all of this and doing one last run.

grunt-tasks/options/shell.js
```diff
@@ -27,7 +27,7 @@ module.exports = function (grunt) {
         },
 
         screenshotsClone: {
-            command: ['[ ${ghToken} ] &&',
+            command: ['[ ! -z "${ghToken}" ] &&',
```
parlarjb: To make sure I'm parsing it properly, "if the `ghToken` is not a null string" ?

Droogans: Yeah, @hussamd told me it's a better way to test if something is a non-empty string than what I was doing earlier.

parlarjb:
LGTM

grunt-tasks/options/shell.js
CITguy: `-z` tests for zero length, so `! -z` is a non-zero length value

CITguy:
LGTM

PR #1025, opened 2015-06-05
test(visualRegression): Avoid grunt-shell tasks

Droogans:


Droogans:
Everything looks good from this fork.

Let's try it from upstream now.

PR #1026, opened 2015-06-05
rxCollapse second version "See More / Less"

ccarteronline:
- [x] Design LGTM
- [ ] QE LGTM
- [x] Dev LGTM
- [x] Dev LGTM

screen shot 2015-06-05 at 4 47 38 pm
screen shot 2015-06-05 at 4 47 46 pm
This is an add on to rxCollapse for story linked here: https://jira.rax.io/browse/EOD-614

### DIFF
--- a/grunt-tasks/options/karma.js
+++ b/grunt-tasks/options/karma.js
@@ -13,8 +13,8 @@ var files = [
     'demo/bower_components/html2canvas/build/html2canvas.js',
     'demo/bower_components/mousetrap/mousetrap.js',
     'demo/bower_components/angular-hotkeys/build/hotkeys.min.js',
-    'demo/bower_components/angular-mocks/angular-mocks.js',
     'demo/bower_components/ng-debounce/angular-debounce.js',
+    'demo/bower_components/angular-mocks/angular-mocks.js',
     // endbower
     'utils/browser-helpers.js',
     'utils/resource-helpers.js'
--- a/src/rxCollapse/docs/rxCollapse.html
+++ b/src/rxCollapse/docs/rxCollapse.html
@@ -1,3 +1,29 @@
-<rx-collapse title="A Custom Title" expanded="true">
+<rx-collapse class="demo-with-title" title="A Custom Title" expanded="true">
     You can put whatever content you want to inside here
 </rx-collapse>
+
+<br/>
+
+<div class="metadata">
+    <dl>
+        <dt>Name:</dt>
+        <dd>Lorem ipsum dolor sit amet</dd>
+        <dt>ID:</dt>
+        <dd>1aa2bfa9-de8d-42f7-9f6d-e6437855b36e</dd>
+        <dt>Region:</dt>
+        <dd>ORD</dd>
+        <dt>Created:</dt>
+        <dd>December 2, 2014 @ 14:28</dd>
+        <rx-collapse class="demo-no-title" expanded="false">
+	        <dt>Name:</dt>
+	        <dd>Lorem ipsum dolor sit amet</dd>
+	        <dt>ID:</dt>
+	        <dd>1aa2bfa9-de8d-42f7-9f6d-e6437855b36e</dd>
+	        <dt>Region:</dt>
+	        <dd>ORD</dd>
+	        <dt>Created:</dt>
+	        <dd>December 2, 2014 @ 14:28</dd>
+        </rx-collapse>
+    </dl>
+</div>
+
--- a/src/rxCollapse/rxCollapse.exercise.js
+++ b/src/rxCollapse/rxCollapse.exercise.js
@@ -6,10 +6,14 @@ var rxCollapse = require('./rxCollapse.page').rxCollapse;
    @exports encore.exercise.rxCollapse
    @param {Object} [options=] - Test options. Used to build valid tests.
    @param {string} [options.cssSelector=] - Fallback selector string to initialize widget with.
+   @param {String} title - The title of the rxCollapse element.
+   @param {Boolean} expanded - Whether or not the rxCollapse element is currently expanded.
    @example
    ```js
    describe('default exercises', encore.exercise.rxCollapse({
-       cssSelector: '.secondary-info rx-paginate', // select one of many widgets on page
+       cssSelector: '.secondary-info rx-collapse', // select one of many widgets on page
+       title: 'My Custom rxCollapse Element',
+       expanded: true
    }));
    ```
  */
@@ -20,7 +24,7 @@ exports.rxCollapse = function (options) {
 
     options = _.defaults(options, {
         title: undefined,
-        expanded: true,
+        expanded: false,
     });
 
     return function () {
@@ -39,34 +43,30 @@ exports.rxCollapse = function (options) {
         });
 
         it('should expand and collapse with toggle', function () {
-            expect(component.isExpanded).to.eventually.eq(options.expanded);
+            expect(component.isExpanded()).to.eventually.eq(options.expanded);
 
-            // Collapse
             component.toggle();
-            expect(component.isExpanded).to.eventually.eq(!options.expanded);
+            expect(component.isExpanded()).to.eventually.eq(!options.expanded);
 
-            // Expand
             component.toggle();
-            expect(component.isExpanded).to.eventually.eq(options.expanded);
+            expect(component.isExpanded()).to.eventually.eq(options.expanded);
         });
 
         if (!_.isUndefined(options.title)) {
             it('should show a custom title', function () {
-                expect(component.titleText()).to.eventually.equal(options.title);
+                expect(component.title).to.eventually.equal(options.title);
             });
         } else {
-            it('should show see more for title of other version', function () {
-                expect(component.titleText()).to.eventually.equal('See More');
+            it('should show "See More" for the title', function () {
+                expect(component.title).to.eventually.equal('See More');
             });
 
-            it('should toggle between see more or see less as text', function () {
-                //Expand
+            it('should toggle between "See More" and "See Less"', function () {
                 component.toggle();
-                expect(component.titleText()).to.eventually.equal('See Less');
+                expect(component.title).to.eventually.equal('See Less');
 
-                //Collapse
                 component.toggle();
-                expect(component.titleText()).to.eventually.equal('See More');
+                expect(component.title).to.eventually.equal('See More');
             });
         }
 
--- a/src/rxCollapse/rxCollapse.exercise.js
+++ b/src/rxCollapse/rxCollapse.exercise.js
@@ -18,7 +18,10 @@ exports.rxCollapse = function (options) {
         options = {};
     }
 
-    options = _.defaults(options, { /* defaults go here */ });
+    options = _.defaults(options, {
+        title: undefined,
+        expanded: true,
+    });
 
     return function () {
         var component;
@@ -35,21 +38,37 @@ exports.rxCollapse = function (options) {
             expect(component.isDisplayed()).to.eventually.be.true;
         });
 
-        it('should show a custom title', function () {
-            expect(component.title.getText()).to.eventually.equal('A Custom Title');
-        });
-
         it('should expand and collapse with toggle', function () {
-            expect(component.isExpanded).to.eventually.be.true;
+            expect(component.isExpanded).to.eventually.eq(options.expanded);
 
             // Collapse
-            component.btnToggle.click();
-            expect(component.isExpanded).to.eventually.be.false;
+            component.toggle();
+            expect(component.isExpanded).to.eventually.eq(!options.expanded);
 
             // Expand
-            component.btnToggle.click();
-            expect(component.isExpanded).to.eventually.be.true;
+            component.toggle();
+            expect(component.isExpanded).to.eventually.eq(options.expanded);
         });
+
+        if (!_.isUndefined(options.title)) {
+            it('should show a custom title', function () {
+                expect(component.titleText()).to.eventually.equal(options.title);
+            });
+        } else {
+            it('should show see more for title of other version', function () {
+                expect(component.titleText()).to.eventually.equal('See More');
+            });
+
+            it('should toggle between see more or see less as text', function () {
+                //Expand
+                component.toggle();
+                expect(component.titleText()).to.eventually.equal('See Less');
+
+                //Collapse
+                component.toggle();
+                expect(component.titleText()).to.eventually.equal('See More');
+            });
+        }
 
     };
 };
--- a/src/rxCollapse/rxCollapse.js
+++ b/src/rxCollapse/rxCollapse.js
@@ -7,12 +7,13 @@ angular.module('encore.ui.rxCollapse', [])
  * @description
  * Hide and show an element with a transition.
  *
- * @param {string} [title] The title to display next to the toggle button.
- * @param {string} [expanded] Initially expanded or collapsed.  Default is expanded.
+ * @param {string} [title] The title to display next to the toggle button. Default is "See More/See Less" toggle.
+ * @param {string} [expanded] Initially expanded or collapsed. Default is expanded.
  *
  * @example
  * <pre>
- *     <rx-collapse title="Filter results" expanded="true"></rx-collapse>
+ *     <rx-collapse title="Filter results" expanded="true">Text Here</rx-collapse>
+ *     <rx-collapse expanded="true">Text Here</rx-collapse>
  * </pre>
  */
 .directive('rxCollapse', function () {
@@ -25,6 +26,10 @@ angular.module('encore.ui.rxCollapse', [])
         },
         link: function (scope, element, attrs) {
             scope.isExpanded = (attrs.expanded === 'false') ? false : true;
+
+            scope.toggleExpanded = function () {
+                scope.isExpanded = !scope.isExpanded;
+            };
         }
     };
 });
--- a/src/rxCollapse/rxCollapse.less
+++ b/src/rxCollapse/rxCollapse.less
@@ -20,14 +20,14 @@ rx-collapse {
           display: inline-block;
           color: @white;
           font-size: 28px;
-          .chevron-mixin(-1);
+          .chevron-mixin;
         }
       }
 
       &.expanded {
         .double-chevron {
           &:after {
-            .chevron-mixin;
+            .chevron-mixin(-1);
           }
         }
       }
@@ -45,5 +45,14 @@ rx-collapse {
       margin: 0;
       padding: 10px;
     }
+  }
+  .sml-title {
+    cursor: pointer;
+    color:#898989;
+    font-size: 10pt;
+  }
+
+  .hide-border {
+      border:none;
   }
 }
--- a/src/rxCollapse/rxCollapse.spec.js
+++ b/src/rxCollapse/rxCollapse.spec.js
@@ -1,8 +1,10 @@
 /* jshint node: true */
 
 describe('rxCollapse', function () {
-    var scope, compile, rootScope, el;
+    var scope, compile, rootScope, el, elCollapsed, elExanded;
     var validTemplate = '<rx-collapse title="Filter results"></rx-collapse>';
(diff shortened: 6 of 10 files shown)


CITguy:
First glance, the arrows seem backward.

ElementE:
@ccarteronline @CITguy - To Ryan's point, the arrows are reversed here. See more (arrow-down), and See less (arrow-up).  Can we show these in context to the metadata lists? Just wanna make sure this pattern can be used in that context.

grunt-tasks/options/karma.js
```diff
@@ -13,8 +13,8 @@ var files = [
     'demo/bower_components/html2canvas/build/html2canvas.js',
     'demo/bower_components/mousetrap/mousetrap.js',
     'demo/bower_components/angular-hotkeys/build/hotkeys.min.js',
-    'demo/bower_components/angular-mocks/angular-mocks.js',
     'demo/bower_components/ng-debounce/angular-debounce.js',
+    'demo/bower_components/angular-mocks/angular-mocks.js',
```
parlarjb: Was this causing a problem?

src/rxCollapse/rxCollapse.js
```diff
@@ -21,10 +22,47 @@ angular.module('encore.ui.rxCollapse', [])
...
+                scope.isExpanded = !scope.isExpanded;
+                if (scope.isExpanded) {
+                    scope.toggleTitle = scope.storedTitles[0];
+                } else {
+                    scope.toggleTitle = scope.storedTitles[1];
+                }
+                scope.setChev(scope.isExpanded);
+            };
+
+            scope.setChev = function (val) {
```
parlarjb: If this function is not accessed in a template, then it doesn't need to be on the `scope`. Instead just do `var setter = function (val) {`, rather that adding more things to the `scope`

src/rxCollapse/rxCollapse.js
```diff
@@ -21,10 +22,47 @@ angular.module('encore.ui.rxCollapse', [])
...
         scope: {
-            title: '@'
+            title: '@',
+            toggleTitle: '&'
         },
         link: function (scope, element, attrs) {
+            scope.storedTitles = ['See Less', 'See More'];
             scope.isExpanded = (attrs.expanded === 'false') ? false : true;
+
+            scope.doSeeMoreOrLess = function (expVal) {
```
parlarjb: If this function is not accessed in a template, then it doesn't need to be on the `scope`. Instead just do `var doSeeMoreOrLess = function (val) {`, rather that adding more things to the `scope`

src/rxCollapse/rxCollapse.js
```diff
@@ -21,10 +22,47 @@ angular.module('encore.ui.rxCollapse', [])
         templateUrl: 'templates/rxCollapse.html',
         transclude: true,
         scope: {
-            title: '@'
+            title: '@',
+            toggleTitle: '&'
         },
         link: function (scope, element, attrs) {
+            scope.storedTitles = ['See Less', 'See More'];
```
parlarjb: This doesn't need to be on the `scope`, do `var storedTitles = [...]`

parlarjb:
napkin 15-06-08 4 54 39 pm

parlarjb:
napkin 15-06-08 4 55 59 pm

grunt-tasks/options/karma.js
ccarteronline: No, not to my knowledge. Seems to be alright to me. Just need to configure tests a bit I think.

ElementE:
Arrows are now in correct context, and code provided Im assuming would allow this to be used in metadata lists.
Design LGTM

ccarteronline:
screen shot 2015-06-09 at 10 48 19 am
screen shot 2015-06-09 at 10 48 25 am

ElementE:
@ccarteronline - Just noticed after I saw the latter screens, please change font-size as follows:

screen shot 2015-06-10 at 10 36 15 am

src/rxCollapse/rxCollapse.exercise.js
```diff
@@ -51,5 +51,19 @@ exports.rxCollapse = function (options) {
...
+        it('should toggle between see more or see less as text', function () {
+            //Expand
+            component.seeMoreTitle.click();
+            expect(component.seeMoreTitle.getText()).to.eventually.equal('See Less');
+
+            //Collapse
+            component.seeMoreTitle.click();
+            expect(component.seeMoreTitle.getText()).to.eventually.equal('See More');
+        });
+
```
Droogans: This should be passed into the exercises `options` object. Not everyone will use the default "See More" message for this component.

Also, you'll need an if statement to not run those tests should they be a classic rxCollapse element.

In fact, looking at the original rxCollapse exercises, I think they'll need some fixes. I can get to that, but not until later.

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+<div class="collapse-container" ng-class="cContainer">
+	<div class="tabStyle_1" ng-show="!tabToShow">
+		<div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+        	<a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+	    </div>
+	    <h3 class="rx-collapse-title title">{{title}}</h3>
+	</div>
+    <div ng-show="isExpanded" ng-class="ispad" ng-transclude></div>
+	<div class="tabStyle_2" ng-show="tabToShow">
+		<div class="smlTitle" >
```
CITguy: Please convert to `<span>`. As you can see from this screenshot, the div is spanning the entire width of its container. With the `cursor: pointer;` css property applied to this element, I see a pointer cursor well to the right of the "See More" text (unfortunately, you can't see it in the screenshot).  This leads me to believe that I can click to expand (which is incorrect).

**as a DIV**
screen shot 2015-06-10 at 1 14 41 pm 2

**as a SPAN**
screen shot 2015-06-10 at 1 19 38 pm

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+			<span class="title" ng-click="toggleTitleAndCollapse()">{{toggleTitle}}</span>
```
CITguy: In regards to the `cursor: pointer;` CSS property on the `.smTitle` and previous comment about clicking the double chevron, it would probably be best to move the `ng-click` functionality to the `.smTitle` element.  This will allow the user to click anywhere in the `.smTitle` (including the double chevron) to expand/collapse the content.

src/rxCollapse/rxCollapse.js
```diff
@@ -7,12 +7,13 @@ angular.module('encore.ui.rxCollapse', [])
  * @description
  * Hide and show an element with a transition.
  *
- * @param {string} [title] The title to display next to the toggle button.
+ * @param {string} [title] The title to display next to the toggle button and toggle different rxCollapse
```
julianbonilla: Can we add something more descriptive like: 

`@param {string} [title] The title to display next to the toggle button.  Default is "See More/See Less" toggle.`

src/rxCollapse/rxCollapse.js
```diff
@@ -21,10 +22,40 @@ angular.module('encore.ui.rxCollapse', [])
...
-            title: '@'
+            title: '@',
+            toggleTitle: '&'
         },
         link: function (scope, element, attrs) {
+            var storedTitles = ['See More', 'See Less'];
             scope.isExpanded = (attrs.expanded === 'false') ? false : true;
+
+            var doSeeMoreOrLess = function () {
+                scope.toggleTitle = storedTitles[+scope.isExpanded];
```
parlarjb: Please don't do this boolean->int conversion. Instead,

```
var storedTitles = { false: 'See Less', true: 'See More' };
...
scope.toggleTitle = storedTitles[scope.isExpanded]
```

And frankly, even that isn't as readable. I'd prefer something like

```
var storedTitle = function (isExpanded) {
    return isExpanded ? 'See Less' : 'See More';
}

...

scope.toggleTitle = storedTitle(scope.isExpanded)
```

halkon: Even that may be a bit much, if `storeTitle` is not even to change just do:

```javascript
scope.toggleTitle = scope.isExpanded === true ? 'See Less' : 'See More';
```
Though my... loathe on scope drives me to:
```javascript
scope.toggleTitle = attrs.expanded === 'true' ? 'See Less' : 'See More';
```
But that's just me and that one really doesn't matter.  It's going to check every time anyways so the extra work is really not needed for something that only gets used in one place

halkon: @ccarteronline the reason for not doing any JS hacks like that is that they decrease readability on the code.  We tend to try to be as explicit as possible. 

It's hard to get out of those habits sometimes, but in this case we have to, we have a lot of people working in projects and when we switch contexts we have to at least be able to read and know easily what's going on. 

That being said we are all guilty of not doing that but the goal of these reviews is to try and catch it and clear things up

src/rxCollapse/rxCollapse.spec.js
```diff
@@ -19,6 +21,9 @@ describe('rxCollapse', function () {
         });
 
         el = helpers.createDirective(validTemplate, compile, scope);
+        elOne = helpers.createDirective(otherTemplate1, compile, scope);
+        elTwo = helpers.createDirective(otherTemplate2, compile, scope);
```
parlarjb: Would you mind updating these to use more descriptive names? Looking through the tests, it's tough to remember which is why.

These might be clearer:
```
        elCollapsed = helpers.createDirective(otherTemplate1, compile, scope);
        elExanded = helpers.createDirective(otherTemplate2, compile, scope)
```

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
-    <h3 class="rx-collapse-title title">{{title}}</h3>
-    <div class="collapse-body" ng-show="isExpanded" ng-transclude>
-    </div>
+<div class="collapse-container" ng-class="cContainer">
+	<div class="tabStyle_1" ng-show="!tabToShow">
+		<div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+        	<a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+	    </div>
+	    <h3 class="rx-collapse-title title">{{title}}</h3>
+	</div>
```
parlarjb: The indent of this `div` is off, could you bring it back one level?

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+    <div ng-show="isExpanded" ng-class="ispad" ng-transclude></div>
+	<div class="tabStyle_2" ng-show="tabToShow">
+		<div class="smlTitle" ng-click="toggleTitleAndCollapse()">
+			<span class="title" >{{toggleTitle}}</span>
+			<i ng-class="arrowChange" class="fa"></i>
+		</div>
+	</div>
```
parlarjb: The indent of this `div` is off, could you bring it back one level?

src/rxCollapse/docs/rxCollapse.html
```diff
@@ -1,3 +1,30 @@
 <rx-collapse title="A Custom Title" expanded="true">
     You can put whatever content you want to inside here
-</rx-collapse>
\ No newline at end of file
+</rx-collapse>
+
+<br/>
+
+
```
CITguy: Please use a maximum of one blank line between elements.

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+<div class="collapse-container" ng-class="cContainer">
+	<div class="tabStyle_1" ng-show="!tabToShow">
+		<div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+			<a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+		</div>
+		<h3 class="rx-collapse-title title">{{title}}</h3>
+	</div>
+	<div ng-show="isExpanded" ng-class="ispad" ng-transclude></div>
+	<div class="tabStyle_2" ng-show="tabToShow">
+		<div class="smlTitle" ng-click="toggleTitleAndCollapse()">
```
CITguy: In regards to https://github.com/rackerlabs/encore-ui/pull/1026#discussion-diff-32148993, this should be a `<span>` to prevent full width and to reduce unnecessary CSS definitions.

I realize that you are setting `width: 70px;` in the LESS file, but this CSS property is unnecessary if you use a `<span>`.

src/rxCollapse/rxCollapse.less
```diff
@@ -46,4 +46,19 @@ rx-collapse {
...
     }
   }
+  .smlTitle {
+    width: 70px;
+  }
+  .smlTitle, .smlTitle .title {
+    cursor: pointer;
+    color:#898989;
+    font-size: 10pt;
+  }
```
CITguy: * `.smlTitle .title` is unnecessary. `.smlTitle` will cover all elements within itself.
* `width: 70px;` is unnecessary if `.smlTitle` is a `<span>` (see [comment on template](https://github.com/rackerlabs/encore-ui/pull/1026#discussion_r32329084))

These two selectors can be reduced to the following, once `.smlTitle` is a `<span>`:
```less
.smlTitle {
  cursor: pointer;
  color: #898989;
  font-size: 10pt;
}
```

src/rxCollapse/templates/rxCollapse.html
ccarteronline: Hey there @CITguy. Could you please be more specific as to which elements you want more concisely surrounded by the `<span>` tag?

ccarteronline:
Okie Dokie, I've removed the `.smlTitle .title` and the font reverts back to the default size of the original rxCollapse object.

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+	<div class="tabStyle_1" ng-show="!tabToShow">
+		<div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+			<a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+		</div>
+		<h3 class="rx-collapse-title title">{{title}}</h3>
+	</div>
+	<div ng-show="isExpanded" ng-class="ispad" ng-transclude></div>
+	<div class="tabStyle_2" ng-show="tabToShow">
+		<span class="smlTitle" ng-click="toggleTitleAndCollapse()">
+			<span class="title" >{{toggleTitle}}</span>
```
CITguy: Remove `title` class to correct font-size issue. The `title` class is meant to be used with headings (h1-h6) and sets the font-size to 18px;

src/rxCollapse/rxCollapse.less
```diff
@@ -46,4 +46,16 @@ rx-collapse {
       padding: 10px;
     }
   }
+  .smlTitle{
+    cursor: pointer;
+    color:#898989;
+    font-size: 10pt;
+  }
+  .fa {
```
CITguy: This is unnecessary. The font-awesome icon is nested within the `.smlTitle` so it automatically gets its color from the definition on line 51.

src/rxCollapse/rxCollapse.less
```diff
@@ -46,4 +46,13 @@ rx-collapse {
       padding: 10px;
     }
   }
+  .smlTitle {
```
TheNando: Css class names should use kebob case, meaning all lowercase with hyphens '-' between words. Also, use full words, ie 'small' instead of 'sml'. When in doubt, scan the file and see if your conventions are consistent with the rest of the file.

CITguy: @TheNando, `sml` I think stands for `see more less`. However, I agree with the `snake-case-class` convention. So, @ccarteronline, please update to use `.sml-title`.

src/rxCollapse/rxCollapse.js
```diff
@@ -7,12 +7,13 @@ angular.module('encore.ui.rxCollapse', [])
  * @description
  * Hide and show an element with a transition.
  *
- * @param {string} [title] The title to display next to the toggle button.
- * @param {string} [expanded] Initially expanded or collapsed.  Default is expanded.
+ * @param {string} [title] The title to display next to the toggle button. Default is "See More/See Less" toggle.
```
CITguy: You should be able to set the default value like so: `[title="See More/See Less"]`

src/rxCollapse/rxCollapse.js
```diff
@@ -25,6 +26,34 @@ angular.module('encore.ui.rxCollapse', [])
...
+                    scope.arrowChange = 'fa-angle-double-down';
+                } else {
+                    scope.arrowChange = 'fa-angle-double-up';
+                }
+            };
+
+            if (!scope.title) {
+                scope.toggleTitle = getTitle();
+                scope.cContainer = 'hideBorder';
+                scope.hasTitle = false;
```
CITguy: I think you can get away without this scope variable. See template comments below.

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
-<div class="collapse-container">
-    <div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
-        <a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
-    </div>
-    <h3 class="rx-collapse-title title">{{title}}</h3>
-    <div class="collapse-body" ng-show="isExpanded" ng-transclude>
-    </div>
+<div class="collapse-container" ng-class="cContainer">
+	<div ng-if="hasTitle" class="collapse-title-wrap collapse-title-wrap-custom">
```
CITguy: You should be able to just use the `title` variable already on the scope for logic.
```
<div ng-if="title" ...>
```

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
...
+		<div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
+			<a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
+		</div>
+		<h3 class="rx-collapse-title">{{title}}</h3>
+	</div>
+	<div ng-show="isExpanded" ng-class="ispad" ng-transclude></div>
+	<div ng-if="!hasTitle" class="collapse-title-wrap collapse-title-wrap-default">
```
CITguy: You should be able to just use the `title` variable already on the scope for logic.
```
<div ng-if="!title" ...>
```

src/rxCollapse/templates/rxCollapse.html
```diff
@@ -1,8 +1,15 @@
-<div class="collapse-container">
-    <div class="double-chevron-cell" ng-class="{ expanded: isExpanded }">
-        <a class="double-chevron" ng-click="isExpanded = !isExpanded"></a>
-    </div>
-    <h3 class="rx-collapse-title title">{{title}}</h3>
-    <div class="collapse-body" ng-show="isExpanded" ng-transclude>
-    </div>
+<div class="collapse-container" ng-class="cContainer">
```
CITguy: You can simplify the link function by defining the tag like so:

```HTML
<div class="collapse-container" ng-class="{ hideBorder: !title }">
```

src/rxCollapse/rxCollapse.js
CITguy: Not a deal breaker, though.

CITguy:
@ccarteronline, can you add one last set of screenshots for good measure?

ccarteronline:
screen shot 2015-07-02 at 1 47 28 pm
screen shot 2015-07-02 at 1 47 36 pm

CITguy:
LGTM

CITguy:
@ElementE can you verify the final screenshot

ccarteronline:
6459zfrf1wxna
giphy

ElementE:
Excellent! Design LGTM

parlarjb:
LGTM

PR #1028, opened 2015-06-05
feat(FlexLayout): remove outer margins on layout-margin

TheNando:
Using layout-margin produces margins around all edges of the child elements by default. For keeping a consistent visual look in our framework and considering that the layout module is intended to replace the Pure Grid system, the outer margins on outer child elements should be removed. This will mean that layout-margin will no longer be compatible with flex-order, as css selectors currently work based on DOM order, (ie, depending on the flex-order, the first and last child selector may not represent the outer elements).

### DIFF
--- a/src/layout/docs/responsiveLayout.html
+++ b/src/layout/docs/responsiveLayout.html
@@ -88,18 +88,20 @@
     <h2 class="title">Flex Order Attribute</h2>
 
     <div layout="row" layout-margin>
-      <div flex flex-order="3" class="flex-box dark-blue">
+      <div flex flex-order="3" class="flex-box dark-blue margin-left">
         [flex-order="3"]
       </div>
-      <div flex flex-order="2" class="flex-box light-blue">
+      <div flex flex-order="2" class="flex-box light-blue margin-right margin-left">
         [flex-order="2"]
       </div>
-      <div flex flex-order="1" class="flex-box dark-green">
+      <div flex flex-order="1" class="flex-box dark-green margin-right">
         [flex-order="1"]
       </div>
     </div>
 
     <p>Add the <code>flex-order</code> attribute to a layout child to set its position within the layout. Any value from 0-9 is accepted.</p>
+
+    <p>Note that the <code>flex-order</code> attribute is not compatible with the <code>layout-margin</code> attribute. This is because the CSS selector engine selects based on DOM markup order and the <code>layout-margin</code> attribute makes use of <code>:first-child</code> and <code>:last-child</code> to apply margins to only the inner elements in the container. As a work-around, <code>[flex-order].left-margin</code> and <code>[flex-order].right-margin</code> classes are availabe to manually add margins. The most likely use cases for these classes is for programatic ordering of children.</p>
 
     <table>
         <tr>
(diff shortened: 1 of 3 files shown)


src/layout/docs/responsiveLayout.html
```diff
@@ -88,19 +88,21 @@ <h2 class="title">Flex Percent Values</h2>
...
       </div>
-      <div flex flex-order="1" class="flex-box dark-green">
+      <div flex flex-order="1" class="flex-box dark-green margin-right">
         [flex-order="1"]
       </div>
     </div>
 
     <p>Add the <code>flex-order</code> attribute to a layout child to set its position within the layout. Any value from 0-9 is accepted.</p>
 
+    <p>Note that the <code>flex-order</code> attribute is not compatible with the <code>flex-margin</code> attribute. This is because the CSS selector engine selects based on DOM markup order and the <code>flex-margin</code> attribute makes use of <code>:first-child</code> and <code>:last-child</code> to apply margins to only the inner elements in the container. As a work-around, <code>[flex-order].left-margin</code> and <code>[flex-order].right-margin</code> classes are availabe to manually add margins. The most likely use cases for these classes is for programatic ordering of children.</p>
```
CITguy: Is it `flex-margin` or `layout-margin`?

src/layout/layout.less
```diff
@@ -60,8 +55,30 @@
...
+    margin-top: @layout-gutter-width / @divisor;
+    margin-bottom: @layout-gutter-width / @divisor;
+  }
+  [layout-margin] > [@{selector}]:not(:first-child):not([flex-order])  {
+    margin-left: @layout-gutter-width / @divisor;
+  }
+  [layout-margin] > [@{selector}]:not(:last-child):not([flex-order])  {
+    margin-right: @layout-gutter-width / @divisor;
+  }
+}
```
CITguy: This can be written more efficiently as:

```less
.layout-marign-for-name(@divisor, @suffix: null) {
  .concat-selector('flex', @suffix);
  [layout-margin] > [@{selector}]:not([flex-order]) {
    // set all margins
    margin: (@layout-gutter-width / @divisor);

    // override first and last child only
    &:first-child { margin-left: 0px; }
    &:last-child { margin-right: 0px; }
  }  
}
```
This is also more efficient when the browser goes to apply styles.  The original logic instructs the browser to apply specific properties for **all** children (not just first and last).

parlarjb:
Where are we with this PR?

TheNando:
I made some changes and am testing them. I'll have something up within the hour.

TheNando:
I know that when I first implemented the layout-margin fix, there was a reason I opted to do it the way I did. I know that with the syntax I used, instead of setting all the margins and then overwriting the first and last to 0, I was update all but the first and last, which while takes more cycles, it doesn't prescribe a 0 margin that will need to be overwritten in the case where a new class is affecting the margins. I know that we have had a bevy of cases where !important is abused to overwrite declarations like this.

However, I was not able to recall or reproduce a case where the layout-margins were affected, so I implemented the suggestions, which is admittedly much cleaner than what I had before.

CITguy:
LGTM

parlarjb:
LGTM

PR #1030, opened 2015-06-08
docs(rxApp): Nav docs

parlarjb:


src/rxApp/README.md
```diff
@@ -128,3 +126,339 @@ The EncoreUI default font is Roboto. This is used for all text on the page and i
...
+        "type": "highlight"
+    }, { 
+        "title": "Section 2"
+    }
+]
+```
+
+In this example, `Section 1` will have a `nav-section-highlight` class applied to it, while `Section 2` will receive the default `nav-section-all` class.
+
+The default Encore nav menu does not currently use the `type` property, and products being added to Encore should avoid it.
```
CITguy: Should we even document this attribute if we aren't using it and recommend avoiding it?

parlarjb: I had that same thought on a lot of these attributes :)

My intent was more along the lines of: "Our designers might choose to use this property, but random product XYZ should not plan on it"

src/rxApp/README.md
```diff
@@ -128,3 +126,339 @@ The EncoreUI default font is Roboto. This is used for all text on the page and i
...
+
+`visibility` can take a few types of values. The original form used in EncoreUI was to pass an expression, filtering values with `rxEnvironmentMatch`, i.e.
+
+```
+"visibility": "('unified-preprod' | rxEnvironmentMatch) || ('local' | rxEnvironmentMatch)",
+```
+
+This expression would be evaluated, checking if the user is currently viewing the app in the `unified-preprod` environment or the `local` environment, and only display the item if one of those was true. (See [rxEnvironment](#/rxEnvironment) for more details on environemnts). This was used to prevent items from being displayed in a production environment if they were only currently available in staging.
+
+This use has somewhat tailed off. We now have different JSON files for each environment, so checking the current environment is not necessary.
```
CITguy: I'd append a `*Note:*`

src/rxApp/README.md
```diff
@@ -128,3 +126,339 @@ The EncoreUI default font is Roboto. This is used for all text on the page and i
...
+            }
+        ]
+```
+
+This example will pull the `accountNumber` from the `pathParams`, and insert `Current Account: 1234` above the children.
+
+
+
+### `roles` (optional)
+
```
CITguy: It might not be a bad idea to note the minimum framework version required to use this feature.

src/rxApp/README.md
CITguy: In that case, I'd add some wording around that attribute being reserved more for design purposes.

src/rxApp/README.md
parlarjb: I've noted it, but a line below, so it's not clearing out this comment.

CITguy:
LGTM

megalord:
I'm not sure how important it is that all the examples in the markup are valid json, but many of them are missing a closing curly brace, and there are lots of errors with trailing commas.  I'm fine if they don't get fixed though - I doubt it'll throw anyone off.

src/rxApp/README.md
```diff
@@ -128,3 +126,341 @@ The EncoreUI default font is Roboto. This is used for all text on the page and i
...
+                "key": "overview",
+                "linkText": "Overview"
+            }, {
+                "href": "/about/{{foobar}}",
+                "key": "about",
+                "linkText": "About"
+            },
+        ]
+```
+If `foobar` is currently in `$route.current.pathParams`, then its value will automatically be inserted into the final URL.
```
megalord: `$route.current.params`?

megalord:
:+1: w/ comments
It would be nice if this would use the ENCORE-SPECIFIC styled callouts that are in the styleguide, but I'm not sure how you feel about putting html in your markdown.

parlarjb:
I'll fix the trailing commas. Too much copy paste to put this together.

src/rxApp/README.md
parlarjb: `$route.current.params`

Issue #1004, opened 2015-06-02
Styleguide layouts need to be rebuilt with flexbox grid


glynnis:
After adding the deprecation warning to pure grids (#1003), I realized that we still need to convert our styleguide layout examples (http://rackerlabs.github.io/encore-ui/#/styleguide/layouts) to use the new grid, especially before we deprecate pure altogether.

anur2213:
Related Jira story:
https://jira.rax.io/browse/FRMW-443
Sub-task of Jira story:
https://jira.rax.io/browse/FRMW-341

PR #1031, opened 2015-06-09
feat(StyleGuide): update layout templates, removing pure, adding flex

TheNando:
- [ ] Dev LGTM
- [ ] Dev LGTM
- [ ] Design LGTM
- [ ] QE LGTM

Addresses #1004

@Peer Reviewers, use https://github.com/rackerlabs/encore-ui/pull/1031/files?w=1 to ignore whitespace during review.

### DIFF
--- a/demo/styleguide/layout-1.html
+++ b/demo/styleguide/layout-1.html
@@ -1,111 +1,134 @@
-<rx-page title="'Layout 1'">
+<rx-page
+        title="'Layout 1'" subtitle="'Layout 1 Subtitle'"
+        layout="column" layout-fill>
 
-<div class="page-actions">
-<ul class="actions-area">
-    <li class="msg-action"><a href="#/styleguide/layouts/1"><i class="fa fa-fw fa-plus fa-lg"></i> Add Object</a></li>
-    <li class="msg-info"><a href="#/styleguide/layouts/1"><i class="fa fa-fw fa-plus fa-lg"></i> Pop-up a Modal</a></li>
-    <li class="msg-warn"><a href="#/styleguide/layouts/1"><i class="fa fa-fw fa-times fa-lg"></i> Delete Object</a></li>
-</ul>
-</div>
+    <h3 class="title">Section Title</h3>
 
-<div class="pure-g clear metadata-section">
-    <div class="pure-u-2-5">
+    <div layout="row">
+        <div class="metadata" flex>
+            <dl>
+                <dt>Name:</dt>
+                <dd>Lorem ipsum dolor sit amet</dd>
 
-<div class="clear-left metadata">
-    <dl>
-        <dt>Name:</dt>
-        <dd>Lorem ipsum dolor sit amet</dd>
-        <dt>Region:</dt>
-        <dd>ORD</dd>
-        <dt>Created:</dt>
-        <dd>December 2, 2014 @ 14:28</dd>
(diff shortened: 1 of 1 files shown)


CITguy:
LGTM given #1028 is merged before or with this PR

Droogans:
@TheNando did you mean to make the div for the page actions so tall?

https://github.com/rackerlabs/encore-ui-screenshots/pull/16/files#diff-c8888eead74b7a4afbe904bdbb7b3b02

CITguy:
@Droogans, I'm wondering if there's a glitch with the visual regression around that. Locally, the page actions look like this (sans blue border):

screen shot 2015-06-10 at 2 04 51 pm

CITguy:
Also, the screen resolution plays into the height.

**568x320**
layout 1 encoreui

glynnis:
It looks like the page actions got moved. They should be adjacent to the page title, not the metadata section.

For comparison:

### Billing (correct page actions position)
screen shot 2015-06-11 at 9 34 27 am

### Cloud (correct page actions position)
screen shot 2015-06-11 at 9 34 09 am

### Layout 1 (page actions position too low)
screen shot 2015-06-11 at 9 32 31 am

TheNando:
That somebody is me. I'll fix it up.

CITguy:
I think it may just be a matter of rearranging items. In rough pseudo code, you might consider this layout
```
#row
  #col
    Title & Subtitle
  #col
    Actions
#row
  #col
    page content
```
This way, the title and actions will be in the same row (as expected with the current layout), and the content will be in rows below that.  The issue that will arise is when you have lots of actions that push the top row height further down the page, but this is already something that we run into today with the existing layout. (see image below - Server Details)

*STAGING: Server Details*
screen shot 2015-06-17 at 9 48 23 am
* blue : would translate to a `layout="row"`
* red : would translate as a `flex` attribute

TheNando:
The problem is the entry point into the page. rxPage's renders to this

```
<rx-page>
   <header></headers>
   <div class="page-body>
      <div class="rx-notifications></div>
      <div class="page-titles"></div>
      <div class="page-content">
         {
            rx-page content injected here, including:
            <div class="page-actions>
         }
      </div>
      ...
   </div>
</rx-page>
```

Roughly, with the layout module, it looks like this, with the page actions not aligned to the top:
image

The red is the first entry point and shows how the layout is separated out into collections of rows and columns.

I think this layout below would work really well, but we'd need to address the window bar title and how to programmatically place the title H1:

image

demo/styleguide/layout-3.html
```diff
@@ -1,86 +1,88 @@
 <rx-page title="'Layout 3'" subtitle="'Sample Create Page'">
 
-<form class="rx-form clear">
```
CITguy: Considering that this is purely cosmetic for code style, (and considering that my rxForm stuff heavily modifies this file), I would feel better if this change was removed from the PR.

parlarjb:
Current status on this?

TheNando:
@parlarjb: I'd like to discuss the issue with the team in more detail and see if we can come up with a solution.

Droogans:
@TheNando Did this ever get resolved?

TheNando:
I'm closing this without it being resolved. I really think the best solution is to add a new rxFlexPage directive that will avoid all these cases.

PR #1033, opened 2015-06-10
fix(rxTokenInterceptor): Only look at hostname for token exclusion

### LINKED ISSUE #1009: rxTokenInterceptor domain name matching
when matching for `rackcdn.com` (the current exclusion default in `rxTokenInterceptor`) it is omitting anything that possibly has a `rackcdn.com` in any part of the path (rather than just the `hostname`)

https://github.com/rackerlabs/encore-ui/blob/1b935aff641198095cc89bea172a481ccc2dc347/src/rxTokenInterceptor/rxTokenInterceptor.js#L29

In a test edge case, and in cases where files retrieved from `rackcdn.com`, the domain name is in the file, when uploading it to a ticket as an attachment (which has happened already a few times), as soon as a request for that attachment is asked for, it fails as it's going through the `TQ API` rather than `rackcdn.com` 

a proposed changed could be:

```javascript
(var url = document.createElement('a')).href = config.url // - "templates/rxAccountSearch.html"
// url.host - "localhost:9000"
// url.hostname - "localhost"

// Don't add the X-Auth-Token if the request URL matches
// something in exclusionList
var exclude = _.some(exclusionList, function (item) {
    if (_.contains(url.hostname, item)) {
        return true;
    }
});
```
(bad javascript used in order to point out this is not a final solution)

This way it only matches on domain name.

halkon:
On behalf of @parlarjb:
Closes #1009

### DIFF
--- a/src/rxTokenInterceptor/rxTokenInterceptor.js
+++ b/src/rxTokenInterceptor/rxTokenInterceptor.js
@@ -20,13 +20,18 @@ angular.module('encore.ui.rxTokenInterceptor', ['encore.ui.rxSession'])
     .provider('TokenInterceptor', function () {
         var exclusionList = this.exclusionList = [ 'rackcdn.com' ];
 
-        this.$get = function (Session) {
+        this.$get = function (Session, $document) {
+            var url = $document[0].createElement('a');
             return {
                 request: function (config) {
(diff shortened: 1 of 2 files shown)


CITguy:
LGTM

halkon:
LGTM (tested locally against Ticketing)

PR #1034, opened 2015-06-10
feat(rxApp): EN-1912 default to an accounts admin user

kyleondata:
When a user searches for an account they will, by default, have the admin user for that account selected.  They can then select a different user if they choose.

JIRA: https://jira.rax.io/browse/EN-1912

### DIFF
--- a/src/rxApp/rxApp.js
+++ b/src/rxApp/rxApp.js
@@ -438,6 +438,19 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
         link: function (scope) {
             scope.isCloudProduct = false;
 
+            // This function is attached to the scope for the sole purpose of making
+            // it easier to test this functionality. A reorganization of this and/or 
+            // the tests is needed in order to pull this off the scope.
+            scope.switchToAdmin = function () {
+                // If the user in the params is not the admin swtich to the admin
+                // this causes the $route.current.params.user to become the admin user
+                var adminUser = _.first(_.where(scope.users, { admin: true }));
+                if (adminUser && ($route.current.params.user !== adminUser.username)){
+                    scope.currentUser = adminUser.username;
+                    scope.switchUser(adminUser.username);
+                }
+            };
+
             var checkCloud = function () {
                 encoreRoutes.isActiveByKey('accountLvlTools').then(function (isAccounts) {
                     if (isAccounts) {
@@ -458,6 +471,8 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
             var loadUsers = function () {
                 var success = function (account) {
                     scope.users = account.users;
+                    scope.switchToAdmin();
+
                     scope.currentUser = $route.current.params.user;
                     if (!scope.currentUser) {
                         // We're not in Cloud, but instead in Billing, or Events, or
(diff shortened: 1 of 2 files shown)


src/rxApp/rxApp.js
```diff
@@ -438,6 +438,16 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
         link: function (scope) {
             scope.isCloudProduct = false;
 
+            scope.switchToAdmin = function () {
```
parlarjb: Could you please take the function off of the scope? If it doesn't need to be accessed from a template, then it's better to do

`var switchToAdmin = function () {`,

then we're not putting too many things onto the scope.

CITguy: I spoke with Kyle about this, and the reason it's on the scope is to access the function within the unit test.  If he changes it back, how can he update the unit tests to call the function?

parlarjb: You'd do it by testing the expected behaviour of the directive, vs. the individual functions.

Namely, if the directive receives a list of users with no `admin`, then it should use the first one. If it receives a list of users with an `admin`, then it should use that.

kyleondata: the problem is that we are hard coding the scope.currentUser.  If I take that line away it breaks other tests.  If I leave it there then the selected user never changes. thoughts?

parlarjb: Sorry, I don't understand what you mean. What do you mean about hard coding? If you take which line away?

You'd definitely have to update the tests to get this to work. If it would be a huge amount of work to update the tests, we can consider not making this change. But I really prefer methods only be added to `scope` when they're accessed from the template, as it gives us an easy way to look at a `link` function and see what interface it provides to templates.

src/rxApp/rxApp.spec.js
```diff
@@ -1014,8 +1015,8 @@ describe('rxAccountUsers', function () {
 
             scope.currentUser = 'hub_cap';
```
CITguy: @parlarjb, this is what @kyleondata is referring to. And yes, it will likely require reorganizing the tests a bit.

src/rxApp/rxApp.js
parlarjb: Ok, rather than reorganize the tests right now, could you put a comment above the function to note that it's _not_ intended as an interface method, but it's only being exposed to the `scope` to make testing easier?

CITguy: :+1:

kyleondata: OK I will add the comment to the function for better understanding of what is happening here.

halkon: I would suggest that if the desired task is to unit test it, then you can just mock out the rest of the stuff to test it:

```javascript
Encore.getAccountUsers = function (params, callback) {
    callback({
        users: [...]
    });
};
```

It feels kind of odd, but in my opinion it puts less on scope

kyleondata:
@parlarjb @CITguy I added the comment as instructed. :)

CITguy:
LGTM

parlarjb:
LGTM

parlarjb:
Just to verify, did you test this locally against Cloud?

kyleondata:
yes

CITguy:
@parlarjb, please communicate in the release notes for this PR that midway tests may break if they expect data to be loaded based on the old behaviour.

kyleondata:
does this branch need to be deleted?

parlarjb:
We delete branches when they're part of the repo, but you brought the branch in from your private fork. So it's up to you if you want it gone from your fork (I'd recommend "yes")

Issue #279, opened 2014-05-21
Filter/sort components by stability


klamping:
It would be helpful for the stability of the components to be a little clearer from an overview standpoint. This may mean allowing components to be organized by stability, or include the stability somewhere in the nav...

glynnis:
@parlarjb Update this to reflect conversation on 2/27

CITguy:
Closing, completed in #1036

PR #1036, opened 2015-06-11
feat(StyleGuide): add searchable, consolidated 'All Components' page

TheNando:


.gitignore
```diff
@@ -38,5 +38,6 @@ results
 /utils/rx-page-objects/index.js
 /utils/rx-page-objects/exercise.js
 /utils/rx-page-objects/API.md
+/demo/assets/generated_demo.css
```
parlarjb: My recollection is that this has to be checked in for github pages to work properly

demo/styleguide/components.html
```diff
@@ -0,0 +1,26 @@
...
+                    <rx-search-box
+                        rx-placeholder="'Filter by any...'"
+                        class="filter-box"
+                        ng-model="searchText">
+                    </rx-search-box>
+                </td>
+            </tr>
+            <tr>
+                <th>Component Name</th>
+                <th>Last Updated</th>
```
megalord: How do you feel about adding stability as one of the columns?

TheNando:
@parlarjb I'll check it back in.

@megalord Adding the stability column sounds good to me. With the badge?

megalord:
Yeah I think the badges are fine.  If it looks strange to have that many badges on a page then just the text works too.

grunt-tasks/modules.js
```diff
@@ -82,9 +83,51 @@ module.exports = function (grunt) {
         return deps;
     }
 
+    var lastModified = {};
+    function moduleLastModified () {
+        grunt.log.writeln('Retriving module dates. This may take up to 10 seconds.');
+
+        // Git shell command to get most recent commit info for every file, including date
+        var command = 'git ls-tree -r --name-only HEAD | ' +
+                'while read filename; do   echo "$(git log -1 --format="%ai" -- $filename) $filename"; done';
```
megalord: Is the reason you used this instead of `ls -l src/` for Windows compatibility or to differentiate between code and documentation changes?  I ask because `git ls-tree` is the main reason I see for this taking up to 10 seconds.

demo/assets/demo.less
```diff
@@ -145,3 +145,17 @@ table.component-attributes {
     vertical-align: top;
   }
 }
+
+.components-table {
```
CITguy: It's confusing having a class with `table` in the name if it's not used directly on a table. Can you update this to be something more appropriate like `.components-list` or `.components-wrapper`?

grunt-tasks/modules.js
megalord: Also, you could use something like `awk` to filter the result, e.g. `ls -l src | awk '{print $7,$8,$10}'`

demo/demo.js
```diff
@@ -26,6 +26,10 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                 rxBreadcrumbsSvc.set();
             }
         })
+        .when('/styleguide/components', {
```
CITguy: Should this path be under `styleguide`? I would think that we might want to organize all components under the `/components/*` umbrella.

glynnis:
@TheNando pics or it didn't happen

grunt-tasks/modules.js
TheNando: `git ls-tree -r --name-only HEAD` is really quick. I found that adding ` src` gets only the output I need which allows me to remove some checks. All the time is spent in the while loop. I do need to use the `git log` command to get the date of the last commit for that file. If I use `ls` I only get the date it was written to my disk.

megalord: sho nuff

TheNando: I ended up using ls after all. As long as I run it through git log, I works out the same.

TheNando:
Ok. Not only do I have stability badges in, I got the whole !# running in ~1.5 seconds! That makes it not-the-longest-running-task which is great, considering it was around 10 seconds when I started.

I also added some smart filtering. As you type in the filter box, it will check to see if your search string:
* Matches anything within the component name (ie. 'menu' -> 'rxActionMenu)
* Matches the beginning of the stability value (ie. 'stable' -> 'stable' but not 'unstable')
* Matches the formatted date, not the raw date (ie. 'Nov' -> 'November 12, 2015' but '11' doesn't match 'November')
* Is all case insensitive (ie. You know what this means =P )

Some things that still need work:
* There are a bunch of files for components now. They should go in their own sub-dir like the styleguide files. I tried to do this, but I must have not done it correctly because the grunt template string were no longer rendering. If one of you guys can add a commit, I'd be grateful.
* Breadcrumbs are busted. If you click a component, the breadcrumbs don't add a ' > Components >' you can click to go back, and if you use the browser's back, the breadcrumbs keep whatever component you last clicked on them. I'm gonna try and work this out next week.

image

@glynnis @ElementE So this work needs to be done sooner or later, but I understand there are some other prerequisites before this goes in. Could you add a comment regarding that so everyone knows not to merge in until x, please?

glynnis:
Please move the "Stability" column to the far right, so the column order is "Component Name", "Last Updated", then "Stability".

@TheNando I'm not aware of any pre-requisites that need to be merged before this.

TheNando:
Now with reordered columns (@glynnis) and working breadcrumbs (at least for the components)! 
image

TheNando:
It keeps failing the build and I'm not sure why. I'll try and fix this tomorrow.

glynnis:
@TheNando Gah, sorry I didn't catch this earlier: can you make the table full width (to match how Cloud Servers handles tables)? We also need to use the standard date format (http://rackerlabs.github.io/encore-ui/#/styleguide/basics#date-formatting).

Once those are addressed, this has a Design LGTM! :rainbow: :cake: :octopus:

glynnis:
PS: Looks like your column titles are mis-matched after your last update.

TheNando:
@glynnis Dang. I can't believe I missed that. I have pushed up updates
image

ElementE:
Design LGTM

TheNando:
This is ready for another review, now that I got all the build kinks worked out. There are still two test failures related to the rxFloatHeader, but these are caused by the nature of the tests, not by an actual problem. None of this added content affects the rxFloatHeader directive.

demo/components.html
```diff
@@ -0,0 +1,32 @@
...
+                <th>Last Updated</th>
+                <th>Stability</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr ng-repeat="comp in components | componentsFilter:searchText">
+                <td><a href="#/components/{{comp.name}}">{{ comp.name }}</a></td>
+                <td>{{ comp.lastModified }}</td>
+                <td>
+                    <span ng-if="!!comp.stability">
```
megalord: Why is this type casting necessary?

TheNando: It isn't necessary, but I've read that it's a good practice because it increases clarity of intent and is also more performant than Boolean(x) explicit or logical evaluation implicit conversion (==), and thus make use of it. No one has to know what type 'stability' is to know that it is expected behavior that it is coerced into a boolean. I'd be happy to hear counterpoints.

megalord: In my mind, using `if` is enough of a clue that the expression is evaluated for truthiness, but I think an even better solution is to have a filter for `_.isEmpty`.  It's all preference though, no need to change this on my accord.

megalord:
LGTM
I'll also point out that this fixes #279  so we remember to close it.

TheNando:
I corrected the typos and implemented the suggestions. I also think I might have a fix for the rxFloatingHeaders that isn't _too_ hacky. It works locally, so we'll see if it passes.

TheNando:
Looks like this guy is ready to go.

grunt-tasks/modules.js
```diff
@@ -32,6 +32,16 @@ module.exports = function (grunt) {
...
         }
 
+        function convertDateFormat (dateStr) {
+            // dateStr recieved as "2015-02-20 13:34:24 -0600"
+            // Covert to "Fri Feb 20 2015 13:34:24 GMT-0600 (CST)" then split into array
+            var d = new Date(dateStr).toString().split(' ');
+
+            // Return as "Feb 20, 2015 @ 13:34:24 (UTC-0600)"
+            return [
+                d[1], parseInt(d[2]) + ',', d[3], '@', d[4].substr(0, 5), d[5].replace('GMT', '(UTC') + ')'].join(' ');
```
parlarjb: When I run `grunt server`, I get

```
Running "modules" task
Warning: Cannot read property 'substr' of undefined Use --force to continue.
```

parlarjb: The reason for this is that I had some stray directories in my `src`, and thus the `getLastModified` function was failing when trying to get `git` information on them.

parlarjb:
Was sorting of the All Components page discussed?

grunt-tasks/modules.js
TheNando: I can add in a check.

TheNando: Added. I don't know why the change isn't showing up here, but if you click to view the full changes, you'll see that there is a check now.

TheNando:
@parlarjb, I don't recall anyone mentioning sorting on the table, but then you just mentioned it, so I added it. I've also rebased, so it should merge in nice.

demo/components.html
```diff
@@ -0,0 +1,41 @@
...
+                    <span ng-show="vm.showCaret('name')" ng-class="vm.getCaretType()"></span>
+                </th>
+                <th ng-click="vm.sortColumn('rawLastModified')">
+                    Last Updated
+                    <span ng-show="vm.showCaret('rawLastModified')" ng-class="vm.getCaretType()"></span>
+                </th>
+                <th ng-click="vm.sortColumn('stability')">
+                    Stability
+                    <span ng-show="vm.showCaret('stability')" ng-class="vm.getCaretType()"></span>
+                </th>
```
TheNando: Now with sorting.

demo/componentsController.js
```diff
@@ -0,0 +1,77 @@
...
+            search = search.toLowerCase();
+
+            return _.filter(components, function (item) {
+                return item.name.toLowerCase().indexOf(search) !== -1 ||
+                    (item.stability && item.stability.substring(0, search.length) === search) ||
+                    item.lastModified.toLowerCase().indexOf(search) !== -1;
+            });
+        };
+    })
+    .controller('componentsController', function ($interpolate, components, rxBreadcrumbsSvc) {
```
TheNando: I rewrote this all to include the sorting and make use of the new Angular style guide recommendations (at least the ones we all seemed to agree on).

demo/componentsController.js
```diff
@@ -0,0 +1,77 @@
...
+            search = search.toLowerCase();
+
+            return _.filter(components, function (item) {
+                return item.name.toLowerCase().indexOf(search) !== -1 ||
+                    (item.stability && item.stability.substring(0, search.length) === search) ||
+                    item.lastModified.toLowerCase().indexOf(search) !== -1;
+            });
+        };
+    })
+    .controller('componentsController', function (components, rxBreadcrumbsSvc) {
```
TheNando: I rewrote this to include the sorting functions as well as to follow the parts of the Angular style guide which we all agreed on.

grunt-tasks/modules.js
```diff
@@ -40,6 +54,9 @@ module.exports = function (grunt) {
             tplFiles: grunt.file.expand('src/' + name + '/*.tpl.html'),
             tplJsFiles: grunt.file.expand('templates/' + name + '/templates/*.html'),
             dependencies: dependenciesForModule(name),
+            stability: stabilities[name],
+            lastModified: convertDateFormat(lastModified[name]),
+            rawLastModified: lastModified[name],
```
TheNando: Including the raw date for sorting.

PR #1039, opened 2015-06-12
fix(rxFloatingHeader): incorrect styles applied to filter header row

CITguy:
Correcting markup missed when applying `rxSearchBox` to floating headers.

**BEFORE**
screen shot 2015-06-12 at 12 00 04 pm

**AFTER**
screen shot 2015-06-12 at 11 58 42 am


parlarjb:
@CITguy rxFloatingHeader is supposed to apply that class automatically. The fact that it's not means there's a bug in rxFloatingHeader w.r.t. rxSearchBox https://github.com/rackerlabs/encore-ui/blob/master/src/rxFloatingHeader/rxFloatingHeader.js#L123

CITguy:
Except it has nothing to do with rxSearchBox. If you remove rxSearchBox from the td, it still has problems calculating the width. The issue is that rxFloatingHeader doesn't calculate the width of a `thead > tr > td`, only `thead > tr > th`.  The `.filter-header` class overrides the dark background on the `th` and is not applied automatically by rxFloatingHeader.

parlarjb:
That's because when rxFloatingHeader was designed, we were specifically looking for `input` fields, and applying `.filter-header` to any `<th>` that had an `input` in it. (Look at the loop that https://github.com/rackerlabs/encore-ui/blob/master/src/rxFloatingHeader/rxFloatingHeader.js#L123 lives in)

Once `rxSearchBox` was introduced, it meant the `input` was taken away, and that code doesn't work anymore.

CITguy:
Ok, I see what you're talking about now. So we'll have to update that logic to apply the `.filter-header` class if finds an `rx-search-box` element, as well, correct?

parlarjb:
Yeah. I'd rather not have to hard-code different elements in there, but it's what we might have to do.

It's a little more complicated though, because we also track the `input` elements, and act on them later. But at this point in the `link` function, `rx-search-box` has not had its template expanded, so we can't yet find the `input`

CITguy:
Closing, @parlarjb will be submitting a PR to address the actual problem with applying the styles.

PR #1042, opened 2015-06-15
test(visualRegression): Remove from dev conf

### LINKED ISSUE #1041: Development Snappit Configuration
Snappit [configurations](https://github.com/rackerlabs/encore-ui/blob/master/protractor.conf.js#L31) should not be included in `protractor.conf.js` if `snappit-mocha-protractor` is not an NPM dev-dependency.  This is a hidden dependency that is only useful to a small percentage of developers and causes local midways to fail unexpectedly for the rest. Considering that snappit is not in the normal workflow for the majority of developers, please remove or comment out these configurations in `protractor.conf.js` so that the developers who do need this functionality can easily uncomment/add the configuration for local development.

Droogans:
Developers who create new components need to run end to end tests
against them. This requires that they remove the `snappit` stuff from
it, since it's been left out of all environments (except the PR builder)
on purpose.

It was originally left in there in case a developer should decide to add
a visual regression test inside of the existing end to end test suite,
but it really doesn't make much sense.

Closes #1041


parlarjb:
LGTM

CITguy:
LGTM

Issue #1044, opened 2015-06-15
Use the "exported" rx-page-objects on Travis.


Droogans:
When a new contributor adds a component that needs to use another page object, this error happens frequently.

https://github.com/rackerlabs/encore-ui/blob/9bc0cfa59c11dfd671f15872b28072e0d3426312/guides/testing.md#some-odd-patterns-in-the-page-object-source-code

If we publish the page object to the file system (`grunt rxPageObjects` does this), we could use that instead and avoid this issue altogether.

PR #1045, opened 2015-06-15
fix(rxBulkSelect): Require from exports

Droogans:
https://goo.gl/OEY6EE

This link explains how you must try to require from both exports (for
the rx-page-objects finished product) vs. using the page object under
src. If you don't do this users who download rx-page-objects will get
errors in their end to end tests from bad require statements.

https://github.com/rackerlabs/encore-ui/issues/1044 is related.


CITguy:
Do we need to attempt importing `rxCheckbox.page` twice? Can't we define `rxCheckbox` at the top of the file?

Droogans:
@CITguy if you pull it in at the root level of the file, it'll get concatenated together with the official `rxCheckbox` component.

You could do `var checkbox` and I (think) it wouldn't collide in greater namespace. It's easier to pull it in twice than to hope it never gets broken.

This is all related to #1044, which I'm very close to pushing up. It addresses all of this.

PR #1046, opened 2015-06-15
fix(e2e): Use published version for testing

### LINKED ISSUE #1041: Development Snappit Configuration

### LINKED ISSUE #1044: Use the "exported" rx-page-objects on Travis.

Droogans:
## What is this all about?!

When we develop new components, we do so in the `src` directory. It looks like this.

```
src/rxActionMenu
├── README.md
├── docs
│   ├── rxActionMenu.html
│   ├── rxActionMenu.js
│   └── rxActionMenu.midway.js
├── rxActionMenu.js
├── rxActionMenu.less
├── rxActionMenu.page.js
├── rxActionMenu.spec.js
└── templates
    └── rxActionMenu.html
```

So, should we need another component inside of the page objects to borrow functionality from (in this case `rxActionMenu` requires functionality from `rxModalAction`), we'd write something like this.

```js
var modal = require('../../rxModalAction/rxModalAction.page').rxModalAction;
```

This will work...in the framework. But once this gets published to npm, the end result looks something like this.

```
rx-page-objects/
├── API.md
├── README.md
├── exercise.js
├── index.js
└── package.json
```

There is no `src` directory. Every page object is concatenated into one large `index.js`. In this environment, you'd "borrow" another component's functionality like this.

```js
var modal = exports.rxModalAction;
```

Since it's in the same file, this works fine. So, in order to make both the framework end to end tests pass *and* have proper require statements work in the published page objects, we write this.

```js
var modal = exports.rxModalAction || require('../../rxModalAction/rxModalAction.page').rxModalAction;
```

Ugly! Error prone! Confusing!

This pull request looks to concatenate all page object files into a single `index.js` file every time a page object file is changed. Then, when end to end tests run in the framework, the large `index.js` file will be used instead of the entire `src` directory. Errors similar to the one above will be caught from now on. There should be no more bad rx-page-objects releases in the future around this type of error.

fix(rxBulkSelect): Require from exports

https://goo.gl/OEY6EE

This link explains how you must try to require from both exports (for
the rx-page-objects finished product) vs. using the page object under
src. If you don't do this users who download rx-page-objects will get
errors in their end to end tests from bad require statements.

https://github.com/rackerlabs/encore-ui/issues/1044 is related.

fix(rxSelectFilter): Require from exports

test(visualRegression): Remove from dev conf

Developers who create new components need to run end to end tests
against them. This requires that they remove the `snappit` stuff from
it, since it's been left out of all environments (except the PR builder)
on purpose.

It was originally left in there in case a developer should decide to add
a visual regression test inside of the existing end to end test suite,
but it really doesn't make much sense.

Closes #1041

fix(testing): Unify to a single style of imports

Closes #1044

### DIFF
--- a/grunt-tasks/options/watch.js
+++ b/grunt-tasks/options/watch.js
@@ -30,8 +30,9 @@ module.exports = {
         files: ['src/**/docs/*.html', 'src/**/*.md', 'demo/**/*', '!demo/bower_components/**/*'],
         tasks: ['build']
     },
-    rxPageObjectsDocs: {
+    rxPageObjects: {
         files: ['src/**/*.page.js', 'src/**/*.exercise.js'],
-        tasks: ['jsdoc2md:rxPageObjects', 'shell:rxPageObjectsDemoDocs']
+        tasks: ['concat:rxPageObjects', 'concat:rxPageObjectsExercises',
+                'jsdoc2md:rxPageObjects', 'shell:rxPageObjectsDemoDocs']
     }
 };
--- a/src/rxActionMenu/rxActionMenu.page.js
+++ b/src/rxActionMenu/rxActionMenu.page.js
@@ -21,8 +21,7 @@ var action = function (actionElement) {
             */
             value: function (customFunctionality) {
                 actionElement.$('.modal-link').click();
-                var rxModal = exports.rxModalAction || require('../rxModalAction/rxModalAction.page').rxModalAction;
-                return rxModal.initialize(customFunctionality);
+                return exports.rxModalAction.initialize(customFunctionality);
             }
         },
 
--- a/src/rxApp/docs/rxApp.midway.js
+++ b/src/rxApp/docs/rxApp.midway.js
@@ -1,5 +1,5 @@
-var rxAppPage = require('../rxApp.page').rxApp;
-var rxPage = require('../rxApp.page').rxPage;
+var rxAppPage = encore.rxApp;
+var rxPage = encore.rxPage;
 
 describe('rxApp', function () {
     var rxAppCustom, rxAppStandard;
--- a/src/rxFeedback/rxFeedback.page.js
+++ b/src/rxFeedback/rxFeedback.page.js
@@ -95,9 +95,8 @@ var rxFeedback = {
     confirmSuccess: {
         value: function (within, fn) {
             if (fn === undefined) {
-                var notifications = exports.rxNotify || require('../rxNotify/rxNotify.page').rxNotify;
                 fn = function () {
-                    return notifications.all.exists('feedback', 'success');
+                    return exports.rxNotify.all.exists('feedback', 'success');
                 };
             }
 
@@ -112,11 +111,10 @@ var rxFeedback = {
 exports.rxFeedback = {
 
     initialize: function (rxFeedbackElement) {
-        var modal = exports.rxModalAction || require('../rxModalAction/rxModalAction.page').rxModalAction;
         rxFeedback.eleFeedback = {
             get: function () { return rxFeedbackElement; }
         };
-        return modal.initialize(rxFeedback);
+        return exports.rxModalAction.initialize(rxFeedback);
     }
 
 };
--- a/src/rxSelectFilter/rxSelectFilter.page.js
+++ b/src/rxSelectFilter/rxSelectFilter.page.js
@@ -1,11 +1,10 @@
 /*jshint node:true*/
 var Page = require('astrolabe').Page;
 var _ = require('lodash');
-var rxCheckbox = require('../rxCheckbox/rxCheckbox.page').rxCheckbox;
 
 var selectOptionFromElement = function (optionElement) {
 
-    return Object.create(rxCheckbox.initialize(optionElement.$('input')), {
+    return Object.create(exports.rxCheckbox.initialize(optionElement.$('input')), {
 
         /**
          * @memberof rxMultiSelect.option
--- a/.jshintrc
+++ b/.jshintrc
@@ -37,6 +37,7 @@
         "demoPage": true,
         "describe": false,
         "element": false,
+        "encore": true,
         "expect": true,
         "exports": false,
         "helpers": true,
--- a/demo/styleguide/layout-3.html
+++ b/demo/styleguide/layout-3.html
@@ -3,82 +3,82 @@
 <form class="rx-form clear">
   <div class="form-area">
     <rx-form-item label="Server Name">
-        <input type="text" id="serverName" required>
+	<input type="text" id="serverName" required>
     </rx-form-item>
 
     <rx-form-item label="Text area">
-        <textarea rows="10" cols="50"></textarea>
+	<textarea rows="10" cols="50"></textarea>
     </rx-form-item>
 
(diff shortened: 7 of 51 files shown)


grunt-tasks/options/watch.js
```diff
@@ -30,8 +30,9 @@ module.exports = {
         files: ['src/**/docs/*.html', 'src/**/*.md', 'demo/**/*', '!demo/bower_components/**/*'],
         tasks: ['build']
     },
-    rxPageObjectsDocs: {
+    rxPageObjects: {
         files: ['src/**/*.page.js', 'src/**/*.exercise.js'],
-        tasks: ['jsdoc2md:rxPageObjects', 'shell:rxPageObjectsDemoDocs']
+        tasks: ['concat:rxPageObjects', 'concat:rxPageObjectsExercises',
+                'jsdoc2md:rxPageObjects', 'shell:rxPageObjectsDemoDocs']
```
CITguy: Where are these new `concat` tasks being defined?

Droogans: [`/options/concat.js`](https://github.com/rackerlabs/encore-ui/blob/b26d21660fddba789a55fb4fc1bc161047e701f9/grunt-tasks/options/concat.js#L24-L56)

They've been around for a while...we currently only use them to publish a final product.

Now, they'll build on any change to the app, so that midways run locally during development use an up to date version of the "published" page objects.

CITguy: ok, cool

CITguy:
So the thing that's bothering me is that we are using the page objects and midways in one of two ways:

1. using the page object / midway directly (`exports.rxFoobar`, `encore.rxFoobar`)
2. assigning to a convenience variable to be used (`var rxFoobar = exports.rxFoobar;`)

Considering that these files are concatenated, convenience variable assignment would be dangerous/confusing to use.  For example:
```
/* rxCheckbox.page.js */
var rxCheckbox = { ... }; // page object definitions

/* someOther.page.js */
var rxCheckbox = exports.rxCheckbox; // convenience variable
```

This can be confusing to debug because `rxCheckbox` is defined twice in the compiled output and may potentially have unexpected side effects.  I would rather explicitly use the objects directly (`encore.*`, `exports.*`) rather than assigning convenience variables.

Droogans:
@CITguy I did my best to catch those cases and set a good example.

https://github.com/rackerlabs/encore-ui/pull/1046/files#diff-dd46895d940bc37f2fc29641a2526f77

If you find anymore "temp" vars, they can likely be removed and replaced with a prepended `exports.`!

src/rxActionMenu/rxActionMenu.page.js
```diff
@@ -21,7 +21,7 @@ var action = function (actionElement) {
             */
             value: function (customFunctionality) {
                 actionElement.$('.modal-link').click();
-                var rxModal = exports.rxModalAction || require('../rxModalAction/rxModalAction.page').rxModalAction;
+                var rxModal = exports.rxModalAction;
                 return rxModal.initialize(customFunctionality);
```
Droogans: This temp var can go away.

```js
return exports.rxModal.initialize(customFunctionality);
```

src/rxFeedback/rxFeedback.page.js
```diff
@@ -95,7 +95,7 @@ var rxFeedback = {
     confirmSuccess: {
         value: function (within, fn) {
             if (fn === undefined) {
-                var notifications = exports.rxNotify || require('../rxNotify/rxNotify.page').rxNotify;
+                var notifications = exports.rxNotify;
                 fn = function () {
                     return notifications.all.exists('feedback', 'success');
                 };
```
Droogans: This temp var can be replaced with `exports.rxNotify`.

src/rxFeedback/rxFeedback.page.js
```diff
@@ -112,7 +112,7 @@ var rxFeedback = {
 exports.rxFeedback = {
 
     initialize: function (rxFeedbackElement) {
-        var modal = exports.rxModalAction || require('../rxModalAction/rxModalAction.page').rxModalAction;
+        var modal = exports.rxModalAction;
         rxFeedback.eleFeedback = {
             get: function () { return rxFeedbackElement; }
         };
```
Droogans: This temp var can be replaced with `exports.rxModalAction`.

src/rxSelectFilter/rxSelectFilter.page.js
```diff
@@ -1,10 +1,10 @@
 /*jshint node:true*/
 var Page = require('astrolabe').Page;
 var _ = require('lodash');
-var rxCheckbox = require('../rxCheckbox/rxCheckbox.page').rxCheckbox;
 
 var selectOptionFromElement = function (optionElement) {
 
+    var rxCheckbox = exports.rxCheckbox;
     return Object.create(rxCheckbox.initialize(optionElement.$('input')), {
```
Droogans: This temp var can be replaced with `exports.rxCheckbox`.

Droogans:
@CITguy good thinking, I found several more.

I'll get this cleared up real fast.

src/rxApp/docs/rxApp.midway.js
```diff
@@ -1,5 +1,5 @@
-var rxAppPage = require('../rxApp.page').rxApp;
-var rxPage = require('../rxApp.page').rxPage;
+var rxAppPage = encore.rxApp;
+var rxPage = encore.rxPage;
```
parlarjb: Needs a change?

CITguy:
So, to clarify:

* `*.page.js` - use `exports.*`, don't use temp variables
* `*.midway.js` - use `encore.*`, temp variables ok

Is this correct?

Droogans:
Yes. Hopefully they're short enough to use inline without the need to define them once at the top of the midway file.

parlarjb:
LGTM

Droogans:
The screenshots are my fault.

I'll need to revert a bit of the commit to get that working again.

I had no idea that would affect the layouts demo!

Droogans:
Merging this before https://github.com/rackerlabs/encore-ui/pull/1060 will require a rebase over there, and some re-work to get it aligned.

It's for the best though since currently, rx-page-objects is entirely broken from the recent changes to rxForm.

halkon:
LGTM

PR #1049, opened 2015-06-16
fix(travis): Valid visual regression installs only

Droogans:
Don't install visual regression tools, which take nearly 3 to 5 minutes,
if the tools aren't even going to be used.

By placing these steps inside of the `after_success` block, it prevents
these installs should the tests fail.

By placing these steps after a check for secure variables existing, it
prevents these installs for PRs that will never run the visual
regression tests in the first place, like from a fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,9 @@ language: node_js
 node_js:
 - '0.10'
 before_install:
-- sudo apt-get install libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++
 - npm install -g grunt-cli bower
 install:
 - travis_retry npm install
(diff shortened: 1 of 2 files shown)


Droogans:
If this builds OK, and everything looks good, then this will still need to be checked by having me push from my fork, too.

Once both pass, it can be merged.

Droogans:
Retest this once https://github.com/rackerlabs/encore-ui/issues/1041 is closed.

parlarjb:
@Droogans Could you look at this?

parlarjb:
@Droogans This is passing, any more work needed here?

Droogans:
Let's do it!

This will save about 5 minutes off of the Travis builds from forks.

Droogans:
Pushing from a fork, I'll link back here with the results.

PR #1052, opened 2015-06-19
fix(rxBulkSelect): only add bulk message to root table

megalord:
The `rx-bulk-select-message` directive was being added to every `<thead>` in the table, which is fine until you need nested tables (like in the [docs](http://rackerlabs.github.io/encore-ui/#/styleguide/tables)).


parlarjb:
LGTM

CITguy:
LGTM

PR #1053, opened 2015-06-19
fix(rxSelect) main methods do not return

jj4th:
Fix rxSelect.main to return an rxSelect object.
Fix htmlSelect.main to return an htmlselect object.


CITguy:
Ooops. LGTM

parlarjb:
LGTM

Issue #1024, opened 2015-06-05
Flexbox Mixins Obsolete


CITguy:
Chrome 29+ and Firefox 28+ have full support for the non-prefixed CSS flexbox properties (see Resources below).  `styles/flexbox.less` can be removed and all mixins replaced with their standardized CSS counterpart.  This will aid in reducing CSS complexity as well as reduce the size of the generated CSS files.

### Resources
* [MDN Flexible Boxes Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Flexible_boxes#Browser_compatibility)
* [caniuse.com Flexbox Compatibility](http://caniuse.com/#feat=flexbox)

CITguy:
I'll handle this after rxForm improvements have been merged.

CITguy:
Closing, captured in JIRA. https://jira.rax.io/browse/FRMW-83

Issue #948, opened 2015-05-12
Update style guide for <label> best practices


CITguy:
It is considered best practice to have a `<label for="inputId">` associated with checkbox and radio inputs (by clicking the label, the input is activated).  This should be documented in our style guide.

**Checkbox w/ Label**
```html
<input type="checkbox" id="chkDemo" />
<label for="chkDemo">Demo Checkbox</label>
```

**Radio w/ Label**
```html
<input type="radio" id="radDemo" />
<label for="radDemo">Demo Radio</label>
```

megalord:
It's also valid to have a single `input` inside a `label`, which eliminates the need for the `for` and `id` attributes.
```html
<label>
  <input type="checkbox" />
  Demo Checkbox
</label>
```

CITguy:
Interesting, I wasn't aware of that trick. :+1:

CITguy:
So, while the second style is easier to implement, it brings challenges when attempting to style the label text based on whether the input is disabled.

**With the "label for" method**
```html
<style>
  /* Sibling selector allows us to style the label */
  input[disabled] + label { color: #ccc; }
</style>
<input type="checkbox" id="chkDemo" disabled />
<label for="chkDemo">Disabled Checkbox</label>
```

**Using the "label around" method"**
```html
<label>
  <input type="checkbox" disabled />
  How do I style this text when the input is disabled?
</label>
```
CSS Selectors only allow to style sideways or downward, never upward (which is needed in this case). This markup makes it impossible to style the label text.

CITguy:
In order to perform CSS sibling selector, you'd require more markup.

```html
<style>
  /* Sibling within label */
  label input[disabled] + label span { color: #ccc; }
</style>
<label>
  <input type="checkbox" disabled />
  <span>I can style this, but it requires more markup</span>
</label>
```

CITguy:
Two goals are desired:

1. Clicking the label should activate the control
2. The label should be styleable based on if the input is disabled.

```LESS
input[type="checkbox"] {
	&[disabled] {
		cursor: not-allowed;

		/* OPTION 1 - Sibling Label
			<input type="checkbox" id="chkDemo" disabled />
			<label for="chkDemo">Checkbox 1</label>
		
		  PROS:
		    * Specific CSS Styles - Low chance of 
		      unwanted sibling <label> styling
		  CONS:
		    * Additional markup required
		*/
		& + label {
			color: red;
			cursor: not-allowed;
		}

[...]

PR #1060, opened 2015-06-22
feat(rxForm): rxForm 2.0

CITguy:
* add `rxNestedElement` service to aid in DOM hierarchy validation
  on new directives
* lots of JSDOC improvements
  * proper "module" naming provides better search capabilities in
    generated ngdocs and easier linking between documentation
* added missing unit tests for
  * rxFieldName
  * rxCheckbox
  * rxRadio
  * rxSelect
* added new directives
  * rx-form
  * rx-form-section
  * rx-field
  * rx-field-content
  * rx-input
  * rx-prefix
  * rx-suffix
  * rx-inline-error
  * rx-help-text
* `rxForm.less` and `vars.less` updated for new code
* eliminated `rx-form-item` from demos and templates
  * `Layout 3: Create Form` markup
  * `rxFeedback` template markup
  * `rxIdentity` demo markup
  * `rxModalAction` demo markup
  * `typeahead` demo markup
* rxNotify demo updated
  * now using latest directives and rxForm patterns
  * using new layout style functionality
* `rxFieldName` moved within `rxForm`

------

* [FRMW-1](https://jira.rax.io/browse/FRMW-1)
* [FRMW-2](https://jira.rax.io/browse/FRMW-2)
* [FRMW-3](https://jira.rax.io/browse/FRMW-3)


CITguy:
Screenshots to follow

CITguy:
### rxForm
* compressed layout is no longer a thing
* rxFormFieldset is no longer a thing
  * Use headings to properly label sections
* Supports inline or stacked forms
  * Stacked via `<rx-form-section stacked>`
  * Can use inline and stacked within the same form, if desired.
* Many New Directives
  * rx-form
  * rx-form-section
  * rx-field
  * rx-field-content
  * rx-input
  * rx-prefix
  * rx-suffix
  * rx-inline-error
  * rx-help-text
* most new directives now subject to proper DOM hierarchy nesting
  * made possible via `rxMisc > rxNestedElement` service
* Updated rx-field-name to adhere to DOM hierarchy validation.
* Lots of updated documentation
  * rxForm demo
    * Updated existing examples to use new rxForm patterns.
    * In addition, new examples demonstrate some more advanced inputs.
  * Forms Styleguide

#### BEFORE
rxform-before

#### AFTER
rxform-after

CITguy:
### Layout 3: Create Form
* Styleguide updated to use new rxForm patterns.

#### BEFORE
layout3-before

#### AFTER
layout3-after

CITguy:
### rxFeedback
* Template updated to use new rxForm patterns.

#### BEFORE
rxfeedback-before

#### AFTER
rxfeedback-after

CITguy:
### rxIdentity
* Demo updated to use new rxForm patterns.

#### BEFORE
rxidentity-before

#### AFTER
rxidentity-after

CITguy:
### rxModalAction
* Demo updated to use new rxForm patterns.

#### BEFORE
rxmodalaction-password-before

#### AFTER
rxmodalaction-password-after

CITguy:
### rxNotify
* Demo updated to make use of new rxForm patterns and new directives.
* Demo also uses new `layout` functionality to place form and example side-by-side.

#### BEFORE
rxnotify-before

#### AFTER
rxnotify-after

CITguy:
### typeahead
* Demo updated to use new rxForm patterns.

#### BEFORE
typeahead-before

#### AFTER
typeahead-after

CITguy:
## rxForm Responsiveness
* rxForm styles are automatically responsive to screen width (and width of containers)
* The immediate children of `rx-form-section` (`rx-field` and `div`) will grow starting at 250px (current field width) to a max of full width. If the width cannot hold all elements, they will wrap to the next line.

**NOTE:** All three examples below are using the exact same markup.

### Wide
The `rx-field` elements will share 1/3 of the width (if wide enough)
rxform-responsive-textinputs-large

### Medium
The first two `rx-field` elements share 1/2 width, while the third consumes full width.
rxform-responsive-textinputs-medium

### Skinny
Same markup, `rx-field` elements wrap automatically.
rxform-responsive-textinputs-small

CITguy:
FYI: mixins for vendor-prefixed CSS properties were not used as per #1024

CITguy:
* Addresses #948 in Styleguide > Forms
* Addresses #884

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+    <div class="component-description">
+      <ul>
+        <li>
+          By default, <code>rx-form-section</code> positions <code>rx-field</code>
+          and <code>div</code> children in an inline fashion (left to right).
+        </li>
+        <li>
+          You can use a stacked layout with these children by placing the
+          <code>stacked</code> HTML attribute on the <code>rx-form-section</code> to
+          position these children in a columnar fashion (top to bottom).
```
megalord: When reading this, I started wondering why there were html tags in a readme.  I haven't been part of the conversations around improving documentation, but I don't think it's a good idea to be putting implementation details in the styleguide.  That just means any updates have to change both places.  I would think that the styleguide just says that you can use inline or stacked fields.  It might also say *why*, but it doesn't say *how*.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+      </ul>
+    </div>
+
+    <h3 class="title">Labels and Inputs</h3>
+    <div class="component-description">
+      <ul class="list">
+        <li>Please use the <code>&lt;label for="formControlId"&gt;</code> format when defining your HTML labels.
+          <ul>
+            <li>When a <code>label</code> is clicked (wherever it is placed in the DOM), it should focus or activate its corresponding form control.</li>
+            <li>If used inline with the form control, place the <code>label</code> <strong>immediately after the form control</strong>.
```
megalord: Does this apply to `<input type="text">` or `<textarea>`?

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+                  </ul>
+                </li>
+                <li>
+                  <strong>NOTE:</strong> While this technique currently does not work for
+                  <code>rxToggleSwitch</code> to activate the control by clicking the label,
+                  CSS styles are still in place to style the label in the event that the control
+                  is disabled in an <code>rxForm</code>.
+                </li>
+                <li>
+                  For consistency, and future compatibility, assume that rxTogleSwitch + label
```
megalord: ~~rxTogleSwitch~~ rxToggleSwitch

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+                  works as expected.
+                </li>
+              </ul>
+            </li>
+          </ul>
+        </li>
+        <li>Do not confuse <code>rxFieldName</code> with a <code>label</code> element.
+          <ul>
+            <li><code>rxFieldName</code> is a descriptive element for a particular form field.</li>
+            <li><code>rxFieldName</code> may contain one or more <code>label</code> elements.</li>
```
megalord: Can you elaborate as to when this would be the case?

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+          about how we style and color buttons. In a nutshell:
+          <ul>
+            <li>
+              Submit buttons should use the green <code>submit</code> class unless
+              you're deleting or formatting. In this case, it should use the
+              <code>negative</code> class.
+            </li>
+          </ul>
+        </li>
+        <li>If you need to use a button in your field for auxillary purposes, use the default blue.</li>
```
megalord: ~~auxillary~~ auxiliary

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+          attribute for use with the label's <code>for</code> attribute. This allows
+          the label to toggle the checkbox no matter where it is placed in the DOM.
+        </li>
+        <li>
+          <strong>DO NOT</strong> wrap the <code>rxCheckbox</code> in a <code>label</code>
+          element.  This prevents CSS rules from styling the label when the checkbox is
+          disabled.
+        </li>
+        <li>
+          If you intend to use a <code>label</code> element inline, place the
```
megalord: The word "inline" throws me off here.  It makes me think that I can use a label element, but not inline, and then this rule would not apply. But that isn't the desired meaning, right?
> If you intend to use a <code>label</code> element, place the

This way, there is no margin for interpretation about what to do with a label element.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+          <span class="msg-info">ENCORE-SPECIFIC:</span> While the majority of radio
+          controls have been used in the context of a <code>rxOptionTable</code>, there
+          is at least one instance in Encore where they are used to make a choice when
+          asking the user to do a hard/soft reboot while viewing server details.
+        </li>
+      </ul>
+    </div>
+    <rx-styleguide code-url="styleguide/forms/checkboxes.html"></rx-styleguide>
+
+    <h3 class="title ">Using a Character Counter</h3>
```
megalord: Extra space in the class.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+              you're deleting or formatting. In this case, it should use the
+              <code>negative</code> class.
+            </li>
+          </ul>
+        </li>
+        <li>If you need to use a button in your field for auxillary purposes, use the default blue.</li>
+      </ul>
+    </div>
+
+    <h3 class="title clear">Selects</h3>
```
megalord: I don't believe `clear` class is needed for this title, or any of the subsequent ones.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+    <div class="component-description">
+      <ul class="list">
+        <li>
+          When an input is disabled, styles are automatically applied to gray out
+          the field and any text within, with a <code>not-allowed</code> pointer
+          style.
+        </li>
+        <li>
+          When label rules (seen above) are applied correctly, the label will also
+          get some styling when a radio or checkbox is disabled.
```
megalord: Too many "when"s in the same sentence.
> When label rules (seen above) are applied correctly to a radio or checkbox,
the label will also get some styling.

demo/styleguide/forms.html
CITguy: Good suggestion. I'll make a tweak.

demo/styleguide/forms.html
CITguy: Inline? No.  I'm referring to inline with like radios and checkboxes.

demo/styleguide/forms.html
CITguy: good catch.

demo/styleguide/forms.html
CITguy: Not really, and looking at it again, the sentence doesn't make much sense. It should just say that it may contain an optional `label` element, not "one or more".

demo/styleguide/forms.html
CITguy: good catch, will fix.

demo/styleguide/forms.html
CITguy: good catch, will fix

demo/styleguide/forms.html
CITguy: good catch. this was a remnant of removing the `clear` class from the original markup (which is unnecessary)

demo/styleguide/forms.html
CITguy: indeed, removed remaining `clear` classes. the only one that should remain is the one on line 2

demo/styleguide/forms.html
CITguy: good call. will update with reference to toggle switches, too

src/rxFieldName/docs/rxFieldName.html
```diff
@@ -28,41 +34,33 @@ <h2 class="title">Field Name States</h2>
...
         <td>
-          <rx-field-name id="eleFieldNameStateOptional">
-            Optional Field
-          </rx-field_name>
+          <div rx-form>
+            <rx-form-section>
+              <rx-field>
+                <rx-field-name id="eleFieldNameStateOptional">
+                  Optional Field
+                </rx-field_name>
```
megalord: `s/_/-/`

src/rxFieldName/rxFieldName.js
```diff
@@ -1,11 +1,56 @@
...
+ *     </rx-field>
+ *   </rx-form-section>
+ * </form>
+ * ...
+ * </pre>
+ *
+ * @param {Boolean=} [ng-required=false]
+ * Is field required? This will add/remove the required symbol to the left of the name.
+ */
+.directive('rxFieldName', function (rxNestedElement) {
```
megalord: `rxForm` now contains both `rxField` and `rxFieldContent`.  With this in mind, I don't see why `rxFieldName` gets its own component (i.e. directory and module).

src/rxForm/README.md
```diff
@@ -1,45 +1,194 @@
...
+use this service as usual.*
+
 
 # Directives
+## Hierarchical
+To provide a standard layout of form fields (and so CSS rules can apply that
+layout), most of the new directives must be nested in a specific hierarchy.
+If you do not nest these elements properly, Angular will throw an error (this
+is by design). So, rule of thumb, aim for 0 console errors.  These directives
+must be nested in the proper order or angular will throw a console error.
```
megalord: This sentence is pretty much a duplicate of the one two lines above.

src/rxForm/README.md
```diff
@@ -1,45 +1,194 @@
...
 
-## rxFormItem
+* [rxForm](/ngdocs/index.html#/api/rxForm.directive:rxForm)
+  * [rxFormSection](/ngdocs/index.html#/api/rxForm.directive:rxFormSection)
+    * [rxField](/ngdocs/index.html#/api/rxForm.directive:rxField)
+      * [rxFieldName](/ngdocs/index.html#/apirxFieldname.directive:rxFieldName)
```
megalord: This link is missing a `/` after `/api`.

src/rxForm/README.md
```diff
@@ -1,45 +1,194 @@
...
-**DEPRECATED - This directive has been marked as deprecated and *will be removed* in a future release of the EncoreUI framework.**
+# Custom Validators
 
-Please use [`rxOptionTable`](/#/component/rxOptionTable) as a stand-in replacement.
+Angular provides its own validator when you use `type="email"`, and you can use
+`<rx-inline-error>` to turn email validation errors into a styled message.  You
+can also use this element if you define a custom validator.
 
-# Services
+### Foocheck validator
```
megalord: Do we really need to include this?  It's basically an excerpt from the angular docs.

src/rxForm/docs/rxForm.html
```diff
@@ -1,117 +1,419 @@
-<div class="forms rx-form" ng-controller="rxFormDemoCtrl" ng-class="{'layout-compressed': compressedLayout.value}">
-    <form name="demoForm">
-        <rx-form-item label="Switch to Compressed Layout">
-            <input type="checkbox" ng-model="compressedLayout.value" />
-        </rx-form-item>
-        <h3 class="title">Text inputs</h3>
-
-        <rx-form-item label="Plain textbox">
+<div class="" ng-controller="rxFormDemoCtrl">
```
megalord: Please remove the `class` attribute altogether.

src/rxForm/docs/rxForm.html
```diff
@@ -1,117 +1,419 @@
...
+              ng-pattern="/^\b(\w)?(\w)\w?\2\1$/" />
+          </rx-input>
+          <rx-help-text>Must be 2-5 letter palindrome (e.g. 'dewed')</rx-help-text>
+          <rx-help-text>Bound Value: {{simple.volumeName}}</rx-help-text>
+
+          <rx-input>
+            <input rx-checkbox
+              id="chkNameRequired"
+              ng-model="simple.isNameRequired" />
+            <label for="chkNamerequired">
```
megalord: `s/r/R/`

src/rxForm/rxForm.js
```diff
@@ -1,21 +1,516 @@
...
+ * <form rx-form>
+ *   <rx-form-section>
+ *     <rx-field>...</rx-field>
+ *     <div>...</div>
+ *   </rx-form-section>
+ * </form>
+ * ...
+ * </pre>
+ *
+ * @param {Boolean=} stacked
```
megalord: Labeling this param as a boolean is misleading.  It isn't really a boolean; it's *presence* is.  `stacked="false"`, `stacked="true"`, and `stacked` are no different.

src/rxForm/rxForm.less
```diff
@@ -307,3 +317,202 @@
...
+
+  input:not([type="checkbox"]):not([type="radio"]) {
+    min-height: @rxForm-input-min-height;
+  }
+
+  rx-field-name,
+  rx-help-text,
+  rx-inline-error,
+  rx-input {
+    margin-bottom: (@rxForm-gutter / 2); // ~8px
```
megalord: This is the type of content that I doubt gets maintained, and could lead to confusion in the future.  How do you feel about removing it?

src/rxForm/rxForm.js
```diff
@@ -1,21 +1,516 @@
...
+        restrict: 'A'
+    });
+})
+/**
+ * @name rxForm.directive:rxFormSection
+ * @ngdoc directive
+ * @restrict E
+ * @description
+ * Structural element directive used for layout of sub-elements.
+ * You may place an optional `.rx-form-actions` class on this element
```
megalord: I did a quick ctrl+f on this page, and this class isn't defined or used in the rest of the PR.  Is that by design?  If so, what purpose does the class serve?

src/rxMisc/rxMisc.js
```diff
@@ -115,38 +129,167 @@ angular.module('encore.ui.rxMisc', ['debounce', 'encore.ui.rxSessionStorage'])
...
+            /*
+             * bare minimum function definition needed for "require"
+             * validation logic
+             *
+             * NOTE: `angular.noop` and `_.noop` WILL NOT trigger validation
+             */
+            opts.link = function () {};
+        }
 
+        return _.assign(defaults, opts);
```
megalord: I think there is a bug here.  `_.assign` mutates the destination object, so for subsequent function calls, the value of `defaults` is a mixture of it's previous value and the `opts` provided to the previous call.

Also, I think `_.defaults` better communicates your intention.

src/styles/vars.less
```diff
@@ -115,10 +115,21 @@
 @inputBackgroundDisabled: #dededd;
 @inputBackgroundError: @inputColorError;
 @inputBorderRadius: 0px;
-@inputBorderColor: #ccc;
+@inputBorderColor: #ccc; // use @rxForm-input-border-color
```
megalord: Was this comment intended for you to come back and fix before the PR is merged?

demo/styleguide/forms.html
megalord: Ok, so are inline labels for text inputs no longer acceptable from a design standard?  If they are, I think it should be mentioned as an exception here.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,340 @@
...
+        </li>
+      </ul>
+    </div>
+
+    <h3 class="title">Labels and Inputs</h3>
+    <div class="component-description">
+      <ul class="list">
+        <li>Please use the <code>&lt;label for="formControlId"&gt;</code> format when defining your HTML labels.
+          <ul>
+            <li>When a <code>label</code> is clicked (wherever it is placed in the DOM), it should focus or activate its corresponding form control.</li>
```
parlarjb: How about mentioning that this is what the `for` attribute does? Or, at a minimum, change the wording. It currently sounds like making it focus/activate is something that the developer will have to figure out

demo/styleguide/forms.html
CITguy: Inline HTML labels should only be used for radios, checkboxes, and toggle switches because those controls are the only ones that make sense to use them.
screen shot 2015-06-24 at 1 14 31 pm

Forms are inline when `rx-field` elements are side-by-side. Elements *within* `rx-field` will still be stacked, and HTML labels for these controls should go in the `rx-field-name`.

screen shot 2015-06-24 at 1 13 50 pm

This will be reflected in my update.

src/rxFieldName/docs/rxFieldName.html
CITguy: good catch. will fix

src/rxFieldName/rxFieldName.js
CITguy: @parlarjb, do you have any objection to moving rxFieldName into rxForm?

parlarjb: No objection

src/rxForm/README.md
CITguy: the intent was to be a segway to the list below, I'll make a tweak.

src/rxForm/README.md
CITguy: good catch. will fix

src/rxForm/README.md
CITguy: This was existing documentation. I'm merely bringing it forward.

src/rxForm/docs/rxForm.html
CITguy: good catch. will fix

src/rxForm/docs/rxForm.html
CITguy: good catch. will fix

src/rxForm/rxForm.js
CITguy: Unfortunately, there's no concept of a boolean-by-presence attribute for ngdocs/jsdocs. This was the closest alternative I could find. Do you have any alternative suggestions?

src/rxForm/rxForm.less
CITguy: the comment, i presume?

CITguy: will remove comment (and similar comments)

src/rxForm/rxForm.js
CITguy: remnant of old code. will remove

src/rxMisc/rxMisc.js
CITguy: Ah, yes. I keep getting those two confused.

CITguy: will change to `_.defaults`

src/styles/vars.less
CITguy: I was originally intending on adding deprecation comments, but I'll be leaving that discussion for another PR. will remove comment

demo/styleguide/forms.html
CITguy: good point. will update

src/rxFieldName/rxFieldName.js
CITguy: TL;DR - I'll move it into `rxForm`

So, thinking about this some more, I realized why I split it into it's own component. I didn't want to clutter rxForm considering the recent refactoring to pull `rxOptionTable` *out* of rxForm.  If we had the ability to define sub-components in demo nav, display, etc., I would place the following components underneath `rxForm`

* rxCharacterCount
* rxCheckbox
* rxSelect
* rxMultiSelect
* rxRadio
* rxOptionTable
* rxToggleSwitch
* etc.

This way, all form control elements and directives would be housed within `rxForm`. However, our build process currently doesn't support this. However, considering the upcoming requirements to documentation, this probably wouldn't be a bad idea to look into.

For this PR, I'll move `rxFieldName` into `rxForm`.

parlarjb: I'll say this: If an element has `rxForm` as a required controller (or has `rxForm` in a required controller "dependency chain"), then I'm fine with it living in rxForm. Ideally we'd make a little better use of Angular's module system, and at least put them in separate files, but our current build system wouldn't allow that.

CITguy: Agreed. I'd love to be able to split each directive into its own source file. `rxForm.js` is getting lengthy with all of the documentation added for the directives.

src/rxForm/rxForm.js
CITguy: @megalord, how would you feel about `@param {*=} stacked`?

src/rxMisc/rxMisc.js
```diff
@@ -115,38 +129,167 @@ angular.module('encore.ui.rxMisc', ['debounce', 'encore.ui.rxSessionStorage'])
...
+            /*
+             * bare minimum function definition needed for "require"
+             * validation logic
+             *
+             * NOTE: `angular.noop` and `_.noop` WILL NOT trigger validation
+             */
+            opts.link = function () {};
+        }
 
+        return _.defaults(defaults, opts);
```
CITguy: I realized I have the two arguments backward. This will be fixed in the next PR update.

src/rxForm/rxForm.js
megalord: I tried several different syntaxes, but they all resulted in errors.  I guess this will have to do.

Droogans:
Let me pull down this branch and see if I can't get it mergable. If everything that conflicts is just my changes, I should be able to get through it. Otherwise I'll grab @CITguy and see if we can't work through this by lunch.

demo/styleguide/forms.html
```diff
@@ -1,132 +1,250 @@
...
+        <li>rxAutoSave interacts exclusively with your model layer. Your UI/template code will be unaware that its state is being saved.</li>
+        <li>See <a href="#/component/rxMisc/">rxAutoSave</a> (which lives within the rxMisc component) for further details.</li>
+      </ul>
+    </div>
+    <rx-styleguide code-url="styleguide/forms/autosave.html"></rx-styleguide>
+
+    <h3 class="title">Manual Form Saving</h3>
+    <div class="component-description">
+      <ul class="list">
+        <li>When an entity is being edited in place, rather than submitting an edit or create form, a notification with a save button is used.</li>
```
glynnis: This is actually not the correct pattern. See https://projects.invisionapp.com/d/main#/projects/3750442

I'm not sure where @megalord and his team are with updating/changing this pattern on the dev side, but if it's still in progress I would prefer not documenting it until the correct pattern exists over documenting something that's incorrect.

megalord: Ah, my bad on this.  I need to update the Encore UI docs to reflect the design change.  Though the diff shows an addition here, this content is no different than the existing documentation, which is also incorrect.  I'll put in a fix, but I don't think it should hold up this PR.

CITguy: @megalord is correct, this looks like an addition, but it's the same content that is already in the [styleguide](http://rackerlabs.github.io/encore-ui/#/styleguide/forms) under "Manual Form Saving". Can the documentation update be submitted in a fast follow PR?

demo/assets/generated_demo.css
```diff
@@ -105,3 +105,21 @@ table.component-attributes th,
...
+  padding: 10px;
+}
+code {
+  padding: 2px 4px;
+}
+pre code {
+  padding: 0;
+  border: 0;
+  border-radius: 0;
+}
```
CITguy: This is being added because it was missed in a previous PR that updated `demo/assets/demo.less`.

CITguy:
* `rxFieldName` moved within `rxForm`
* updated to reflect new syntax in page objects, midways, and exercise test files

Droogans:
LGTM. Overall, I'm ok with this as is. It is too large for a single pull request (as most refactor PRs tend to be), and there needs to be some more work on the rxForm page objects, but those don't need to be here in this PR. It can wait.

There are lots of new features to interact with. Some of the ones that stand out to me is all of the invalid/disabled states for form fields, as well as a better way to deal with groups of radio elements and groups of checkbox elements.

demo/styleguide/forms.html
glynnis: I'm good with a fast follow :+1:

glynnis:
I'm with @Droogans -- LGTM on the design side :+1:, but wouldn't be surprised given the size of the PR if we'll need to make some tweaks and additions.

Thank you for this work @CITguy ! Very excited about the new form and how all this hard work is going to pay off :smile: :clap: :clap: :clap:

CITguy:
Updated to make use of `rx-help-text` within rxOptionTable to display the selected label text. This is to better differentiate the data value from the help text (selected label).

*Compare the position and style of "(Already saved data)" in the Before & After below.*

### Before
screen shot 2015-06-26 at 10 58 43 am

### After
screen shot 2015-06-26 at 10 57 27 am

CITguy:
If all is well, please let me know when I should perform a squash for the final update.

Droogans:
Squash label's up! I say go for it.

demo/styleguide/forms/validation.html
```diff
@@ -1,3 +1,71 @@
...
+          <input rx-checkbox always-invalid id="chkInvalidOne" value="1" ng-model="chkInvalidOne" />
+          <label for="chkInvalidOne">Checked Option</label>
+        </rx-input>
+        <rx-input>
+          <input rx-checkbox always-invalid id="chkInvalidTwo" value="2" ng-model="chkInvalidTwo" />
+          <label for="chkInvalidTwo">Unchecked Option</label>
+        </rx-input>
+      </rx-field-content>
+
+      <!-- Uncomment if/when "invalid" styles available -->
```
parlarjb: Is there an issue filed for this?

demo/styleguide/forms.html
```diff
@@ -1,132 +1,250 @@
...
+    <h3 class="title">Form Validation</h3>
+    <div class="component-description">
+      <ul class="list">
+        <li>
+          Because Encore is built on Angular, we leverage ngModel directives used for validation.
+          As a result, the EncoreUI framework has custom CSS styling if an element has the
+          "ng-invalid" and/or "ng-dirty" classes (which are dynamically added by ngModel validation logic).
+        </li>
+        <li>Some components will style based on both classes, while others will only use <code>ng-invalid</code>.</li>
+        <li><strong>NOTE:</strong> Designing an EncoreUI specific design pattern for placement of form validation messages is on our roadmap.</li>
```
parlarjb: "Creating an EncoreUI-specific design pattern" sounds a little less awkward, but this is minor, I'll leave it up to you

demo/styleguide/forms/validation.html
CITguy: Yes, https://github.com/rackerlabs/encore-ui/issues/1065

demo/styleguide/forms.html
CITguy: Sounds good. will update.

parlarjb:
Could you add something for `[API Documentation](/ngdocs/index.html#/api/rxButton.directive:rxButton)` to the `component-template`?

CITguy:
@parlarjb, that may be something that will need to be addressed in #1063. Right now, the jsdoc paths are inconsistent, which would result in many links being broken.

parlarjb:
Ok, thanks

CITguy:
Squashed!

src/rxCheckbox/rxCheckbox.js
```diff
@@ -1,4 +1,19 @@
...
+ * @name rxCheckbox.directive:rxCheckbox
+ * @ngdoc directive
+ * @restrict A
+ * @scope
+ * @description
+ * Attribute directive that wraps a native checkbox element in markup required for styling purposes.
+ *
+ * @example
+ * <pre>
+ * <input rx-checkbox />
```
parlarjb: How do you feel about adding more attributes here? I don't want someone to look at it and think that the `rx-checkbox` will handle the `ng-model` automatically.

src/rxCheckbox/rxCheckbox.spec.js
```diff
@@ -0,0 +1,118 @@
...
+            beforeEach(function () {
+                parent = el.parent();
+            });
+
+            it('should be disabled', function () {
+                expect(parent.attr('class')).to.contain('rx-disabled');
+            });
+
+            it('should have "rxCheckbox" class', function () {
+                expect(parent.attr('class')).to.contain('rxCheckbox');
```
parlarjb: It's outside the scope of this PR, but what was your thinking on naming the class `rxCheckbox` instead of `rx-checkbox`? I can't recall if we spoke on that before.

CITguy: We have element directives that are referenced as `rx-help-text` or `rx-field-name` (in spinal case), and IMO adding a dot to the beginning for a class name could be confusing (`.rx-field-name` vs `rx-field-name`).  So by using lowerCamelCase for class names, we avoid the similarity confusion (`.rxFieldName` vs `rx-field-name`).  But this is something that should be discussed around CSS standards (among other items).

src/rxCheckbox/rxCheckbox.js
CITguy: ah, yes. I'll update.

src/rxCheckbox/rxCheckbox.spec.js
Droogans: I say follow Angular's [normalization naming methods](https://docs.angularjs.org/guide/directive#normalization).

src/rxForm/README.md
```diff
@@ -1,45 +1,204 @@
...
 
-Creates a field row wrapped in a label. Used for fields which include a single input/field.
+* [rxForm](/ngdocs/index.html#/api/rxForm.directive:rxForm)
+  * [rxFormSection](/ngdocs/index.html#/api/rxForm.directive:rxFormSection)
+    * [rxField](/ngdocs/index.html#/api/rxForm.directive:rxField)
+      * [rxFieldName](/ngdocs/index.html#/api/rxForm.directive:rxFieldName)
+      * [rxFieldContent](/ngdocs/index.html#/api/rxForm.directive:rxFieldContent)
+        * [rxInput](/ngdocs/index.html#/api/rxForm.directive:rxInput)
+          * [rxPrefix](/ngdocs/index.html#/api/rxForm.directive:rxPrefix)
+          * [rxSuffix](/ngdocs/index.html#/api/rxForm.directive:rxSuffix)
```
parlarjb: Can we added a bracketed number after each line to represent the "has-a" relationship here? i.e. `rxForm` can have `n` `rxFormSection` children, but each `rxField` can only have `1` `rxFieldName`,

```
rxForm
    rxFormSection (n)
        rxField (n)
            rxFieldName (1)
…
```

src/rxForm/README.md
```diff
@@ -1,45 +1,204 @@
...
 
-Angular provides its own validator when you use `type="email"`, and you can use `.inline-error` to turn email validation errors into a styled message.
+* [rxButton](/#/component/rxButton)
+* [rxCharacterCount](/#/component/rxCharacterCount)
+* [rxCheckbox](/#/component/rxCheckbox)
+* [rxOptionTable](/#/component/rxOptionTable)
+* [rxRadio](/#/component/rxRadio)
+* [rxSelect](/#/component/rxSelect)
+* [rxToggleSwitch](/#/component/rxToggleSwitch)
+* [typeahead](/#/component/typeahead)
```
parlarjb: Do you know if `rxMultiSelect` works properly? If not, we should file an issue.

src/rxForm/README.md
```diff
@@ -1,45 +1,204 @@
...
+
+* If your legend pertains to at least one row, place a `.title` heading variant
+  before the desired `rx-form-section` element.
+* If your legend pertains to controls in a single column, place a `.title` heading
+  variant within the `rx-field` element, at the top.
+
+## Before &amp; After
+The new markup may look a little wordy, but it is designed to provide flexibility
+(to a point) around content for advanced field inputs. To be explicit, the new
+directives were designed based on feedback around:
```
parlarjb: Please feel free to add something about `rxFormItem` being incredibly brittle and prone to breakage.

src/rxForm/README.md
```diff
@@ -1,45 +1,204 @@
...
+*You can see this in sections below if you resize the width of your browser.*
 
-**DEPRECATED - This directive has been marked as deprecated and *will be removed* in a future release of the EncoreUI framework.**
 
-Please use [`rxOptionTable`](/#/component/rxOptionTable) as a stand-in replacement.
+# Validation
 
-# Services
+## Required Fields
+When displaying a field should be required, please make use of the `ng-required`
```
parlarjb: `When displaying _that_ a field should be required`, or `When displaying a field _that_ should be required`

src/rxForm/README.md
CITguy: I've not tested `rxMultiSelect` with the `rxForm` revamp. However, `rxMultiSelect` uses `rxSelect` as a base, so I would imagine that it would be fine. That being said: yes, we should probably file an issue to **ensure** `rxMultiSelect` compatibility with `rxForm`.

src/rxForm/README.md
CITguy: I like the second option. will update

src/rxForm/README.md
CITguy: good call. will do

src/rxForm/README.md
CITguy: good idea. will do

src/rxForm/README.md
CITguy: Actually, yes! It does work.

screen shot 2015-06-26 at 4 57 31 pm

I'll add it as a compatible directive.

CITguy: Added `rxMutliSelet` before `rxOptionTable`

CITguy:
Addressed feedback from @parlarjb and added examples around using `rxMultiSelect`.

screen shot 2015-06-26 at 5 17 58 pm

screen shot 2015-06-26 at 5 18 08 pm

src/rxForm/README.md
```diff
@@ -1,45 +1,206 @@
...
+These directives must be nested in the following hierarchy (*parens denote how many items should be nested within its parent*):
 
-Creates a field row wrapped in a label. Used for fields which include a single input/field.
+* [rxForm](/ngdocs/index.html#/api/rxForm.directive:rxForm)
+  * [rxFormSection](/ngdocs/index.html#/api/rxForm.directive:rxFormSection) (N)
+    * [rxField](/ngdocs/index.html#/api/rxForm.directive:rxField) (N)
+      * [rxFieldName](/ngdocs/index.html#/api/rxForm.directive:rxFieldName) (1)
+      * [rxFieldContent](/ngdocs/index.html#/api/rxForm.directive:rxFieldContent) (1)
+        * [rxInput](/ngdocs/index.html#/api/rxForm.directive:rxInput) (N)
+          * [rxPrefix](/ngdocs/index.html#/api/rxForm.directive:rxPrefix) (N)
```
parlarjb: What's the use case for having `N` `rxPrefix` or `rxSuffix` elements?

src/rxForm/rxForm.js
```diff
@@ -1,21 +1,568 @@
...
+ *       <rx-field-name>Salary</rx-field-name>
+ *       <rx-field-content>...</rx-field-content>
+ *     </rx-field>
+ *   </rx-form-section>
+ * </form>
+ * ...
+ * </pre>
+ *
+ * @param {Boolean=} [ng-required=false]
+ * Is field required? This will add/remove the required symbol to the left of the name.
```
parlarjb: Can we change this to "Is _this_ field required?"

src/rxForm/rxForm.js
```diff
@@ -1,21 +1,568 @@
...
+        },
+        templateUrl: 'templates/rxFieldName.html'
+    });
+})
+/**
+ * @name rxForm.directive:rxFieldContent
+ * @ngdoc directive
+ * @restrict E
+ * @description
+ * Sructural element directive used for layout of sub-elements.
```
parlarjb: "Structural"

src/rxForm/rxForm.js
```diff
@@ -1,21 +1,568 @@
...
+ *
+ * @example
+ * <pre>
+ * ...
+ * <form rx-form>
+ *   <rx-form-section>
+ *     <rx-field>
+ *       <rx-field-name>Salary:</rx-field-name>
+ *       <rx-field-content>
+ *         <rx-input>...</rx-input>
```
parlarjb: Could you please show an actual element inside of the `<rx-input>`, for clarity?

src/rxForm/README.md
CITguy: I'm not aware of any present use cases, but I didn't want our present designs to limit what we could potentially accomplish.  

#### Inspiration
I realize that it's not officially a supported framework pattern, but I wanted `rxPrefix` and `rxSuffix` to have similar functionality to how Bootstrap does ["Input Groups"](http://getbootstrap.com/components/#input-groups).  The only difference between EncoreUI and Bootstrap, in this regard, is styling.

src/rxForm/rxForm.js
CITguy: Sure. will do

src/rxForm/rxForm.js
CITguy: Good catch. will fix

src/rxForm/rxForm.js
CITguy: sure, will update

src/rxForm/rxForm.less
```diff
@@ -307,3 +317,222 @@
...
+  font-size: 0.8em;
+  font-style: italic;
+  margin: 1em 0em;
+}
+
+rx-field-name {
+  box-sizing: border-box;
+  display: block;
+  margin: 0px;
+  padding: 0px;
```
parlarjb: These four lines are the same four as are defined above (for `[rx-form], `rx-form-section`, `rx-field-name`, etc.), do they need to be repeated?

src/rxForm/rxForm.less
```diff
@@ -307,3 +317,222 @@
...
+          justify-content: flex-start;
+          align-items: stretch;
+
+          input:not([type="checkbox"]):not([type="radio"]),
+          textarea {
+            flex: 1 1 auto; // as dynamic flex item
+          }
+
+          input:not([type="checkbox"]):not([type="radio"]),
+          textarea {
```
parlarjb: Is there a styling reason that you put the `flex: 1 1 auto` into its own selector above, as opposed to putting it into this one?

src/rxForm/README.md
parlarjb: In Bootstrap, they say `We do not support multiple add-ons on a single side.` If we were to follow that, then these should be limited to 1, instead of `N`.

I also think it's more likely to confuse people with `N`, as it's counter to what people would expect to do.

If you do choose to keep `N`, then can you change the line above these examples to say "parens denote how many items _can_ be nested within its parent"

parlarjb: A final note on this stuff: To be safe, we should probably say `0..N`, `0..1` for all of these, instead of just `N` and `1`. We don't want people to think that (for example) `rxPrefix` is required.

CITguy: Ok, until we actually have a use case, I'll default to `1` on prefix/suffix. I'll update to using the `0..*` notation, as well.

src/rxForm/rxForm.less
CITguy: ah, good catch. will remove duplication.

src/rxForm/rxForm.less
CITguy: ah, the previous definition was left over from some prototyping. will consolidate.

parlarjb:
LGTM

src/rxCheckbox/rxCheckbox.spec.js
CITguy: The key thing is that the class names don't always come into play. 

### Attribute Directives
#### That wrap their target
For attribute directives that wrap their target element in custom markup (`rx-checkbox`, `rx-radio`, `rx-select`, etc.), we don't have a descriptive element for CSS selection.  In this case, I think it would be best to use the angular-normalized css name.

```LESS
.rxCheckbox { ... }
.rxSelect { ... }
.rxRadio { ... }
```

#### Others
For attribute directives that do not wrap their target element, use the attribute name selector.

```LESS
[rx-form] { ... }
```

### Element Directives
For element directives, use the element name.

```LESS
rx-field-name { ... }
rx-help-text { ... }
```

Droogans: > IMO adding a dot to the beginning for a class name could be confusing (`.rx-field-name` vs `rx-field-name`)

That's as far as I had read the first time.

I see now that `.rxCheckbox` is signifying that they are wrapping the whole thing in a single class....at least that's what I'm understanding. `.rx-checkbox` would mean that it's just another class name inside of, say, a regular element directive `rx-checkbox`.

Is that closer to what's happening here?

CITguy:
Wow, could Travis be any slower? :turtle:

src/rxCheckbox/rxCheckbox.spec.js
CITguy: I use `.rxCheckbox` because I needed to scope styles based on the target element's parent (CSS can't do this).  As such, a namespace needed to be created (`.rxCheckbox`) to properly scope the style definitions and not confuse the CSS selector as an element selector.

For other attribute and element directives, I can apply the style definitions downward using the attribute (`[rx-foobar]`) or element (`rx-foobar`) selector as the namespace.

CITguy: To capture conversations around this topic, I've created #1079 for discussion.

megalord:
LGTM

halkon:
I got [̲̅$̲̅(̲̅ιοο̲̅)̲̅$̲̅] [̲̅$̲̅(̲̅ιοο̲̅)̲̅$̲̅] for this LGTM

PR #1070, opened 2015-06-23
docs(styles): fast follow update to <code> styles

CITguy:
While I realize that these styles will probably be out of date once design is finished with their work, but in the mean time, this is a fix to make documentation more legible.

### Before
screen shot 2015-06-23 at 3 43 40 pm

### After
screen shot 2015-06-23 at 3 44 10 pm

### DIFF
--- a/demo/assets/demo.less
+++ b/demo/assets/demo.less
@@ -146,23 +146,33 @@ table.component-attributes {
   }
 }
 
-pre, code {
-    color: #404040;
-    background-color: #D0D0D0;
-    border: 1px solid #BABABA;
-    border-radius: 2px;
+.component-description,
+.component-demo {
+  pre,
+  code {
+      color: #404040;
+      background-color: #D0D0D0;
+      border: 1px solid #BABABA;
+      border-radius: 2px;
+  }
+
+  code {
+      padding: 0 2px;
+      margin: 0 1px;
+  }
+
+  pre {
+      padding: 10px;
+
+      code {
+          border: none;
+          padding: 0;
+          margin: 0;
+          border-radius: 0;
+      }
+  }
 }
 
-pre {
-    padding: 10px;
-}
-
-code {
-    padding: 2px 4px;
-}
-
-pre code {
-    padding: 0;
-    border: 0;
-    border-radius: 0;
+p {
+    line-height: 1.5;
 }
--- a/demo/assets/generated_demo.css
+++ b/demo/assets/generated_demo.css
@@ -105,21 +105,31 @@ table.component-attributes th,
 table.component-attributes td {
   vertical-align: top;
 }
-pre,
-code {
+.component-description pre,
+.component-demo pre,
+.component-description code,
+.component-demo code {
   color: #404040;
   background-color: #D0D0D0;
   border: 1px solid #BABABA;
   border-radius: 2px;
 }
-pre {
+.component-description code,
+.component-demo code {
+  padding: 0 2px;
+  margin: 0 1px;
+}
+.component-description pre,
+.component-demo pre {
   padding: 10px;
 }
-code {
-  padding: 2px 4px;
-}
-pre code {
+.component-description pre code,
+.component-demo pre code {
+  border: none;
   padding: 0;
-  border: 0;
+  margin: 0;
   border-radius: 0;
+}
+p {
+  line-height: 1.5;
 }


demo/assets/demo.less
```diff
@@ -146,23 +146,30 @@ table.component-attributes {
...
+code {
     color: #404040;
-    background-color: #D0D0D0;
+    background-color: #dededd;
     border: 1px solid #BABABA;
     border-radius: 2px;
 }
 
+code {
+    padding: 0px 2px;
```
TheNando: Most CSS Linters will require that values of 0 not have units specified. We link to a linter in our css styleguide that enforces this.

https://github.com/rackerlabs/encore-ui/blob/master/guides/css-styleguide.md
http://csslint.net/

TheNando:
Suggestion regarding unit specification on 0 value properties. Other than that, LGTM.

CITguy:
Updated to not apply styles to "Markup", "LESS", "Protractor", etc. Additionally, I removed the generated demo CSS file (I'm not sure why it was added, anyway).

### Before
screen shot 2015-06-23 at 6 48 00 pm

### After
screen shot 2015-06-23 at 6 49 07 pm

megalord:
LGTM
I wouldn't mind seeing this released asap - it's very difficult to read the docs right now.

CITguy:
@megalord you and me both :smile:

CITguy:
Reverted removal of `generated_demo.css`.

TheNando:
LGTM

CITguy:
FYI: To be clear, this is a temporary fix to correct readability of docs until official documentation styles are ready.

CITguy:
After speaking with @glynnis, part of this was a bug (overriding colors for syntax highlighting), but the original colors were part of the grand re-envisioning of the documentation work that is being done on the design team. I'll be waiting on specific feedback from @glynnis before we do anything else with this PR.

demo/assets/demo.less
```diff
@@ -146,23 +146,33 @@ table.component-attributes {
...
-    border: 1px solid #BABABA;
-    border-radius: 2px;
-}
+.component-description,
+.component-demo {
+  pre,
+  code {
+      color: #404040;
+      background-color: #dededd;
+      border: 1px solid #BABABA;
```
glynnis: Changes to parent elements/scope for these styles is :+1:, however, let's leave colors as is because of more design changes coming down the pipeline (most code blocks will appear on the app grey background instead of white with the docs redesign).

demo/assets/demo.less
```diff
@@ -146,23 +146,33 @@ table.component-attributes {
...
+      }
+  }
 }
 
-pre code {
-    padding: 0;
-    border: 0;
-    border-radius: 0;
+p {
+    line-height: 1.5;
```
glynnis: Line height changes are fine for now re: increasing legibility, but these may be changing shortly

glynnis:
Also noting here for future reference the documented process we should try to stick to with PRs w/r/t design involvement and pre-reqs for PRs:
- https://github.com/rackerlabs/encore-ui/blob/master/CONTRIBUTING.md#preparing-an-issue-for-development
- https://github.com/rackerlabs/encore-ui/blob/master/CONTRIBUTING.md#pr-steps

@CITguy and I already spoke and resolved this (and we figured out how to modify this PR so it can be merged :ok_hand: :ok_woman:), but posting here for others who may find this relevant: Many times visual fixes like this that aren't outlined or approved by design upfront can feel like a step backwards for the design team, which has a strategy for rolling out changes (currently handled through sprint goals, etc.). Sometimes the steps to get to the desired outcome are "good enough" and not "great" because of lack of resources and time on the design side.

My hope is that very soon design will have enough bandwidth to begin focusing more on long-term strategy and better communication of that strategy, so that dev has more context and a better understanding of incremental steps necessary for design changes. I'm sure without that context it can seem like design work cuts corners or introduces changes that don't make sense, but a lot of that is due to the volume and breadth of work that we've had to accommodate across multiple projects (which I'm sure has been frustrating for dev!). 

Thanks for bearing with us as we work toward a better process that accounts for vision and improved communication with dev. 

@CITguy let me know if you need any clarification on requested changes. :+1:

CITguy:
Reverted to original colors.

CITguy:
rebased

PR #1077, opened 2015-06-26
fix(travis): Valid visual regression installs only

Droogans:
Don't install visual regression tools, which take nearly 3 to 5 minutes,
if the tools aren't even going to be used.

By placing these steps inside of the `after_success` block, it prevents
these installs should the tests fail.

By placing these steps after a check for secure variables existing, it
prevents these installs for PRs that will never run the visual
regression tests in the first place, like from a fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,9 @@ language: node_js
 node_js:
 - '0.10'
 before_install:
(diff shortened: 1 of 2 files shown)


parlarjb:
LGTM

Droogans:
The upstream (rackerlabs) branch at https://github.com/rackerlabs/encore-ui/pull/1049 passed...just waiting for this not to mess up.

PR #1081, opened 2015-06-30
docs(rxStatus): Fix button styles

### LINKED ISSUE #1061: Update rxStatus Demo
With the addition of the new `<code>` and `<pre>` styles in the demo, the `rxStatus` demo page now has some ambiguous controls.

screen shot 2015-06-22 at 1 25 14 pm

Please update `rxStatus` demo to use `rx-button` in order to differentiate buttons from code blocks.

megalord:
https://jira.rax.io/browse/FRMW-44
Fixes #1061 
screen shot 2015-06-30 at 12 52 28 pm

### DIFF
--- a/src/rxStatus/docs/rxStatus.html
+++ b/src/rxStatus/docs/rxStatus.html
@@ -2,48 +2,52 @@
     <div style="clear: left;">
         <ol>
             <li>
-                <input type="button" value="setLoading"
-                    ng-click="setLoading('This should set a spinner')"/>
-                        should set a spinner</li>
+                <button class="button" ng-click="setLoading('This should set a spinner')">setLoading</button>
+                should set a spinner
+            </li>
             <li>&nbsp;</li>
             <li>
-                <input type="button" value="setSuccess"
-                    ng-click="setSuccess('This should show a success on $routeChangeSuccess')"/>
-                        should show a success on <strong>$routeChangeSuccess</strong> -
-                        <input type="button" ng-click="triggerRouteChangeSuccess()" value="Trigger it!" /></li>
+                <button class="button" ng-click="setSuccess('This should show a success on $routeChangeSuccess')">setSuccess</button>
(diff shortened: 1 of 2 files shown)


parlarjb:
LGTM

megalord:
rebased

src/rxStatus/docs/rxStatus.html
```diff
@@ -2,48 +2,52 @@
     <div style="clear: left;">
         <ol>
             <li>
-                <input type="button" value="setLoading"
-                    ng-click="setLoading('This should set a spinner')"/>
-                        should set a spinner</li>
+                <button class="button" ng-click="setLoading('This should set a spinner')">setLoading</button>
```
CITguy: Can we just add a rule to `rxButton.less` so that styles apply to the button element as well?
```
button,
.button { ... }
```
I feel like `button.button` is a bit redundant.

CITguy: @glynnis, is this something that would need discussed in a separate issue?

glynnis: @CITguy Good question :+1:. I think it should be filed as an issue, yes. 

My concern in making this change would be that at the project level, folks might be using `button` as an element and applying their own styles, and we'd need to give them fair warning about changes (especially without visual regression in place at the project level yet). It'd be the projects' responsibility to update/refactor, of course, but I think that kind of change warrants its own PR and a call out in release notes.

CITguy: #1084 created to capture conversation

CITguy:
LGTM

src/rxStatus/docs/rxStatus.html
glynnis: Perfect! Thanks :+1:

parlarjb:
LGTM

PR #1082, opened 2015-06-30
docs(tables): Add proper styling for delete X pattern.

### LINKED ISSUE #930: Fix deletable rows demo in styleguide "tables" section
X's should be red on hover like they are in Cloud. Just realized that this is not the case in the Styleguide, and we're seeing the pattern implemented incorrectly at the product level as a result.

Current state (on hover) in styleguide:
screen shot 2015-05-01 at 9 58 36 am

What it should be on hover (red color, pointer cursor):
screen shot 2015-05-01 at 10 00 29 am 3

megalord:
Fixes #930 
screen shot 2015-06-30 at 1 27 17 pm
To clarify the screenshot, my mouse is hovering the x, but it isn't visible.  The x is not red unless hovered.


parlarjb:
LGTM

CITguy:
LGTM

halkon:
LG....To @CITguy & @parlarjb

PR #1083, opened 2015-06-30
fix(rxAccountUsers): check account number before making API call

TheNando:
https://jira.rax.io/browse/FRMW-28


parlarjb:
Could you please add a test here https://github.com/rackerlabs/encore-ui/blob/master/src/rxApp/rxApp.spec.js#L964 ? 

You could have a test case where `accountNumber` is not set on the `$route.current.params`, and make sure that `Encore.getAccountUsers` doesn't get called

TheNando:
Sure thing.

parlarjb:
LGTM

cc @halkon

halkon:
brentrambooriginal

PR #1086, opened 2015-07-03
feat(rxPaginate): Add ability to disable itemsPerPage persistence

### LINKED ISSUE #1074: Add ability to disable rxPaginate persistence
For an upcoming change to Encore Account Summary, we will have paginated data with page sizes 5, 10, and 20.  However, when the user selects "20" this page size should NOT persist to paginated data in other locations within Encore.

Please add the ability to disable persistence for a single paginated data set.

I'd suggest something like a `persist` attribute that defaults to `true`. When `persist="false"`, changing the page size should NOT persist across URL locations.

parlarjb:
FRMW-55
Closes #1074

### DIFF
--- a/src/rxPaginate/rxPaginate.js
+++ b/src/rxPaginate/rxPaginate.js
@@ -59,15 +59,6 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage', 'debounce'])
 
             var table = parentElement;
 
-            scope.updateItemsPerPage = function (itemsPerPage) {
-                return scope.pageTracking.setItemsPerPage(itemsPerPage).then(function () {
-
-                    // Set itemsPerPage as the new default value for
-                    // all future pagination tables
-                    PageTracking.userSelectedItemsPerPage(itemsPerPage);
-                });
-            };
-
             scope.scrollToTop = function () {
                 table[0].scrollIntoView(true);
             };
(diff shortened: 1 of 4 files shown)


src/rxPaginate/templates/rxPaginate.html
```diff
@@ -32,7 +32,7 @@
             <div>
                 Show
                 <ul>
-                    <li ng-repeat="i in pageTracking.itemSizeList"><button ng-disabled="pageTracking.isItemsPerPage(i)" class="pagination-per-page-button" ng-disabled="i == pageTracking.itemsPerPage" ng-click="updateItemsPerPage(i)">{{ i }}</button> </li>
+                    <li ng-repeat="i in pageTracking.itemSizeList"><button ng-disabled="pageTracking.isItemsPerPage(i)" class="pagination-per-page-button" ng-disabled="i == pageTracking.itemsPerPage" ng-click="pageTracking.setItemsPerPage(i)">{{ i }}</button> </li>
```
parlarjb: This is the key change. Previously we were using a "global" method to set the new items per page. Now it's been moved to a method on the individual pager instance.

megalord:
LGTM

src/rxPaginate/rxPaginate.js
```diff
@@ -452,17 +427,44 @@ angular.module('encore.ui.rxPaginate', ['encore.ui.rxLocalStorage', 'debounce'])
 
     }
 
-    return {
+    var PageTracking = {
+        /**
+        * @property {number} itemsPerPage This is the current setting for the number
+        * of items to display per page.
+        * @property {boolean} [persistItemsPerPage=false] Whether or not a change to this pager's itemsPerPage
```
CITguy: This should be `[persistItemsPerPage=true]` to match the default value.

parlarjb: :facepalm:

CITguy:
LGTM

PR #1089, opened 2015-07-06
docs(CommonQuestions): Move "Common Questions" to its own document

parlarjb:
https://jira.rax.io/browse/FRMW-14


megalord:
LGTM

PR #1098, opened 2015-07-08
E2e docs rx action menu

Droogans:
This also include rxModalAction, since action menus use them so much. They're all linked up together all nice and everything.


CITguy:
LGTM

PR #1099, opened 2015-07-09
fix(rxFeedback): Show url of current page.

megalord:
`rxFeedback` is displaying whatever the url was at the application's bootstrap, which is not necessarily the url at the time the feedback is submitted.  This has no effect on the api call that is made.

To reproduce this bug, go to the home page for the docs, and then click to any other page.  Open up the feedback modal, and you'll see
screen shot 2015-07-09 at 2 14 48 pm

### DIFF
--- a/src/rxFeedback/rxFeedback.js
+++ b/src/rxFeedback/rxFeedback.js
@@ -94,8 +94,11 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
             sendFeedback: '=?onSubmit'
         },
         link: function (scope) {
-            scope.currentUrl = $location.url();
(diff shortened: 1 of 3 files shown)


Droogans:
Can you add a property that will report back the feedback url in rxFeedback.page.js?

megalord:
@Droogans `rxFeedback` is a mixin for `rxModalAction`, so you can use the `subtitle` property to access the url.

Droogans:
Ah yes I completely forgot. Thanks.

parlarjb:
LGTM

CITguy:
LGTM

PR #1103, opened 2015-07-13
Rx form page object refactor

Droogans:
This addresses a couple of issues with the rxForm refactor, from the perspective of the page objects.

1. There are some missing tests for plain old HTML elements. These have been added in the demo "for comparison", and also to test against.

2. Some of the exercises now assert against more specific things (such as text in select elements).

3. ~~The exercises no longer ask for what type of element the component is (these functions have been removed from the page objects entirely as well). If you're using a checkbox/radio/select element, you can interact with it in specific ways to determine it's type. There's no need to ask explicitly.~~ This was added back in. No change here.

4. The `html*` and `rx*` page objects have been unified under a single object that responds to both plain html-type elements and rx-type elements. It's easier to use on in every instance of a form element, instead of forcing a tester to understand the subtle differences between a native form element, and one that has been extended by the framework.

before:

```
var plainHtml = encore.htmlSelect.initialize($('#plainHtml'));
var encoreSelect = encore.rxSelect.initialize($('#rxSelect'));
```

after:

```
var plainHtml = encore.rxSelect.initialize($('#plainHtml'));
var encoreSelect = encore.rxSelect.initialize($('#rxSelect'));
```

### DIFF
--- a/src/rxCheckbox/rxCheckbox.exercise.js
+++ b/src/rxCheckbox/rxCheckbox.exercise.js
@@ -34,12 +34,12 @@ exports.rxCheckbox = function (options) {
             }
         });
 
-        it('should be a checkbox type', function () {
-            expect(component.isCheckbox()).to.eventually.be.true;
-        });
-
         it('should be present', function () {
             expect(component.isPresent()).to.eventually.be.true;
+        });
+
+        it('should be a checkbox', function () {
+            expect(component.isCheckbox()).to.eventually.be.true;
         });
 
         it('should ' + (options.visible ? 'be' : 'not be') + ' visible', function () {
@@ -53,6 +53,32 @@ exports.rxCheckbox = function (options) {
         it('should ' + (options.selected ? 'be' : 'not be') + ' selected', function () {
             expect(component.isSelected()).to.eventually.eq(options.selected);
         });
+
+        if (options.disabled) {
+            it('should not respond to selecting and unselecting', function () {
+                component.isSelected().then(function (selected) {
+                    selected ? component.deselect() : component.select();
+                    expect(component.isSelected()).to.eventually.equal(selected);
+                    // even though it "doesn't respond to selecting and unselecting"
+                    // attempt to put it back anyway in case it did actually respond.
+                    // that way nobody accidentally submits a swapped checkbox after
+                    // running these exercises.
+                    selected ? component.select() : component.deselect();
+                    expect(component.isSelected()).to.eventually.equal(selected);
+                });
+            });
+        } else {
+            it('should respond to selecting and unselecting', function () {
+                component.isSelected().then(function (selected) {
+                    selected ? component.deselect() : component.select();
+                    expect(component.isSelected()).to.eventually.not.equal(selected);
+                    // exercises should never alter the state of a page.
+                    // always put it back when you're done.
+                    selected ? component.select() : component.deselect();
+                    expect(component.isSelected()).to.eventually.equal(selected);
+                });
+            });
+        }
 
         it('should ' + (options.valid ? 'be' : 'not be') + ' valid', function () {
             expect(component.isValid()).to.eventually.eq(options.valid);
--- a/src/rxCheckbox/rxCheckbox.page.js
+++ b/src/rxCheckbox/rxCheckbox.page.js
@@ -5,14 +5,74 @@ var Page = require('astrolabe').Page;
 /**
  * @namespace
  */
-var htmlCheckbox = {
+var rxCheckbox = {
+    eleWrapper: {
+        get: function () {
+            return this.rootElement.element(by.xpath('..'));
+        }
+    },
+
+    eleFakeCheckbox: {
+        get: function () {
+            return this.eleWrapper.$('.fake-checkbox');
+        }
+    },
+
+    /**
+       Useful for situations where input types might change in the future, ensuring that the expected one is being used.
+       @returns {Boolean} Whether or not the element in question is a checkbox.
+     */
+    isCheckbox: {
+        value: function () {
+            return this.rootElement.getAttribute('type').then(function (type) {
+                return type === 'checkbox';
+            });
+        }
+    },
+
     /**
      * @function
      * @returns {Boolean} Whether the checkbox is currently displayed.
      */
     isDisplayed: {
         value: function () {
-            return this.rootElement.isDisplayed();
+            var page = this;
+            return this.eleFakeCheckbox.isPresent().then(function (isFakeCheckbox) {
+                return isFakeCheckbox ? page.eleFakeCheckbox.isDisplayed() : page.rootElement.isDisplayed();
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the checkbox is disabled.
+     */
+    isDisabled: {
+        value: function () {
+            var page = this;
+            return this.eleFakeCheckbox.isPresent().then(function (isFakeCheckbox) {
+                if (isFakeCheckbox) {
+                    return page.eleWrapper.getAttribute('class').then(function (classes) {
+                        return _.contains(classes.split(' '), 'rx-disabled');
+                    });
+                }
+                return page.rootElement.getAttribute('disabled').then(function (disabled) {
+                    return (disabled ? true : false);
+                });
+            });
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the checkbox is present on the page.
+     */
+    isPresent: {
+        value: function () {
+            var page = this;
+            return this.eleFakeCheckbox.isPresent().then(function (isFakeCheckbox) {
+                return isFakeCheckbox || page.rootElement.isPresent();
+            });
         }
     },
 
@@ -29,34 +89,12 @@ var htmlCheckbox = {
     },
 
     /**
-     * @function
-     * @returns {Boolean} Whether or not the checkbox is currently selected
+       @function
+       @returns {Boolean} Whether or not the checkbox is selected.
      */
     isSelected: {
         value: function () {
             return this.rootElement.isSelected();
-        }
-    },
-
-    /**
-     * @function
-     * @returns {Boolean} Whether or not the checkbox is disabled
-     */
-    isDisabled: {
-        value: function () {
-            return this.rootElement.getAttribute('disabled').then(function (disabled) {
-                return (disabled ? true : false);
-            });
-        }
-    },
-
-    /**
-     * @function
-     * @returns {Boolean} Whether or not the checkbox exists on the page
-     */
-    isPresent: {
-        value: function () {
-            return this.rootElement.isPresent();
         }
     },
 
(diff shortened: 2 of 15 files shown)


Droogans:
I just realized I need to reword all of these to include breaking changes.

src/rxCheckbox/rxCheckbox.page.js
```diff
@@ -5,58 +5,90 @@ var Page = require('astrolabe').Page;
...
     isDisplayed: {
         value: function () {
-            return this.rootElement.isDisplayed();
+            var page = this;
+            return this.eleFakeCheckbox.isPresent().then(function (isFakeCheckbox) {
+                if (isFakeCheckbox) {
+                    return page.eleFakeCheckbox.isDisplayed();
+                }
+                return page.rootElement.isDisplayed();
+            });
```
Droogans:
```js
if (isFakeCheckbox) {
    return isFakeCheckbox || page.rootElement.isDisplayed();
}
```
```

File: src/rxCheckbox/rxCheckbox.page.js [hunk 2]
```diff
@@ -5,58 +5,90 @@ var Page = require('astrolabe').Page;
...
+    isPresent: {
         value: function () {
-            return this.rootElement.isSelected();
+            var page = this;
+            return this.eleFakeCheckbox.isPresent().then(function (isFakeCheckbox) {
+                if (isFakeCheckbox) {
+                    return page.eleFakeCheckbox.isPresent();
+                }
+                return page.rootElement.isPresent();
+            });
```
Droogans:
```js
return isFakeCheckbox || page.rootElement.isPresent();
```

src/rxCheckbox/rxCheckbox.exercise.js
```diff
@@ -54,6 +50,32 @@ exports.rxCheckbox = function (options) {
...
+            it('should not respond to selecting and unselecting', function () {
+                component.isSelected().then(function (selected) {
+                    if (selected) {
+                        component.deselect();
+                        expect(component.isSelected()).to.eventually.equal(selected);
+                    } else {
+                        component.select();
+                        expect(component.isSelected()).to.eventually.equal(selected);
+                    }
+                });
```
CITguy: Can you simplify with the `expect` outside of the `if...else` structure?

```javascript
if (selected) {
  component.deselect();
} else {
  component.select();
}
expect(component.isSelected()).to.eventually.equal(selected);
```

src/rxCheckbox/rxCheckbox.exercise.js
```diff
@@ -54,6 +50,32 @@ exports.rxCheckbox = function (options) {
...
+            it('should respond to selecting and unselecting', function () {
+                component.isSelected().then(function (selected) {
+                    if (selected) {
+                        component.deselect();
+                        expect(component.isSelected()).to.eventually.not.equal(selected);
+                    } else {
+                        component.select();
+                        expect(component.isSelected()).to.eventually.not.equal(selected);
+                    }
+                });
```
CITguy: Same comment as above... `expect` outside of `if...else`

src/rxCheckbox/rxCheckbox.page.js
CITguy: agreed

src/rxCheckbox/rxCheckbox.page.js
CITguy: agreed

src/rxCheckbox/rxCheckbox.page.js
```diff
@@ -235,7 +155,9 @@ exports.rxCheckbox = {
     },
 
     /**
-     * @returns {rxCheckbox} Page object representing the _first_ `<input rx-checkbox />` element found on the page
+       Don't use this if you're expecting a regular html checkbox on the page. This only checks for Encore
+       specific rxForm-style checkboxes.
```
CITguy: What should a developer use if they ARE expecting a regular html checkbox?

src/rxRadio/docs/rxRadio.html
```diff
@@ -154,7 +154,7 @@ <h2 class="title">Radio States</h3>
             <input rx-radio
                    id="radInvalidEnabledTwo"
                    value="2"
-                   ng-model="validEnabled"
+                   ng-model="invalidEnabled"
```
CITguy: good catch

src/rxRadio/rxRadio.exercise.js
```diff
@@ -34,10 +34,6 @@ exports.rxRadio = function (options) {
             }
         });
 
-        it('should be a radio type', function () {
-            expect(component.isRadio()).to.eventually.be.true;
-        });
-
```
CITguy: What's the reason for removing this?

For a `<input rx-radio>`, this is needed to test that the `type="radio"` attribute is set on the target element. It may be redundant for plain radio buttons, but I believe it should remain an exercise.

src/rxSelect/docs/rxSelect.midway.js
```diff
@@ -13,47 +9,77 @@ describe('rxSelect', function () {
...
+        selectedText: 'Third'
     }));
 
     describe('(State) Valid NG-Disabled', encore.exercise.rxSelect({
         cssSelector: '#selValidNgDisabled',
         disabled: true,
         visible: true,
-        valid: true
+        valid: true,
+        selectedText: "Disabled by \'ng-disabled\' attribute"
```
CITguy: Do the single quotes need escaping in a double-quoted string?

src/rxSelect/rxSelect.page.js
```diff
@@ -341,6 +272,8 @@ exports.rxSelect = {
     },
 
     /**
+     * Don't use this if you're expecting a regular html select element on the page. This only checks for Encore
+     * specific rxForm-style select elements.
```
CITguy: What should the developer use if they ARE expecting a plain HTML select?

src/rxRadio/rxRadio.page.js
```diff
@@ -162,6 +120,8 @@ exports.rxRadio = {
     },
 
     /**
+     * Don't use this if you're expecting a regular html radio button on the page. This only checks for Encore
+     * specific rxForm-style radio buttons.
```
CITguy: What should the developer use if they ARE expecting a plain radio button?

src/rxCheckbox/rxCheckbox.exercise.js
Droogans: Oh right I meant to make sure it worked this way first and then change it to this later.

Thanks.

src/rxCheckbox/rxCheckbox.page.js
Droogans: `initialize` only. I could spend some time to try and handle both, but it could get messy. I'd have to pick one that would "win" should there be one of each, since as far as I know there is no way to determine which appears first in page order.

Droogans: Also, I don't really like the idea of `main` in form elements at all, anyway.

The `main` functions are included for the sake of consistency. I would never recommend using it for elements like this where there can be more than one.

It's purpose it to make it easier to grab the one and only expected element on a page. As in, rxAccountInfo, rxBreadcrumbs, etc.

src/rxRadio/rxRadio.exercise.js
Droogans: I put the reasoning in the PR details section:

> The exercises no longer ask for what type of element the component is (these functions have been removed from the page objects entirely as well). If you're using a checkbox/radio/select element, you can interact with it in specific ways to determine it's type. There's no need to ask explicitly.

I can look into putting it back. I can see where it's nice to know that you haven't accidentally implemented a series of checkboxes when you meant to do it with radio buttons.

src/rxSelect/docs/rxSelect.midway.js
Droogans: Nope. I put it as single with escaped singles inside, and then changed it to doubles on the outside.

You're looking at the part where I forgot to delete the unnecessary escaping. Good catch.

src/rxSelect/rxSelect.page.js
Droogans: Same as above. I'll look into getting it added back in there.

src/rxRadio/rxRadio.page.js
Droogans: Same as above. I'll look into getting it added back in there.

Droogans:
Ok easy stuff is out of the way.

Working on the `main` stuff now.

Droogans:
Alright I got the `is{Checkbox,Radio,Select}` functions added back in.

src/rxCheckbox/rxCheckbox.page.js
Droogans: @CITguy I don't really know how I should go about addressing this.

Got any ideas?

src/rxSelect/rxSelect.page.js
```diff
@@ -3,92 +3,149 @@ var _ = require('lodash');
...
-    /**
-     * @function
-     * @returns {Boolean} Whether or not the `<option>` is currently selected
-     */
-    isSelected: {
-        value: function () {
-            return this.rootElement.isSelected();
+            return this.rootElement.getTagName().then(function (type) {
+                return type === 'select';
+            });
```
CITguy: Is `isSelect` necessary?

Because `rx-checkbox` and `rx-radio` are meant to be used on an `<input>` element (which can be any number of form controls), it makes sense to verify the correct `type` attribute.  However, `rx-select` is meant to be used on a `<select>` element (which can only be a select control), so I don't see much point in validating the element.

src/rxCheckbox/rxCheckbox.page.js
CITguy: In regards to documentation wording, I was just thinking that if you inform the developer _not_ to use something in a particular scenario, it would be a good idea to provide them with information on what to use _instead_.

I've always wondered why we have `main` in the first place. It doesn't have much use outside of the controlled UI of the framework documentation and it makes more sense to initialize by CSS selector anyway.  Is there any reason to keep `main`?

Droogans: It's handy for stuff where you know there's only going to be one of on the page. Saves some typing.

```js
var breadcrumbs = encore.rxBreadcrumbs.initialize($('.rx-breadcrumbs'));
expect(breadcrumbs.names).to.eql(['Home', 'Info', 'Sub-Info']);
```

becomes

```js
expect(encore.rxBreadcrumbs.main.names).to.eql(['Home', 'Info', 'Sub-Info']);
```

Droogans: That being said, I'll go ahead and delete `main` altogether from these three.

CITguy: agreed, the "only one instance" scenario makes sense. but these cases do not

CITguy:
LGTM

Droogans:
@parlarjb please sign off on the `BREAKING CHANGES` in the commit messages.

It's for end to end tests so the number of impacted users is virtually no one since the rxForm refactor just happened a week or two ago.

parlarjb:
@Droogans I'm ok with the breaking changes :+1:

PR #1107, opened 2015-07-17
feat(rxMetadata): add rxMetadata component

### LINKED ISSUE #885: Create Metadata Directives

### LINKED ISSUE #939: Re-evaluate the usage of description lists <dl> for metadata

CITguy:
* rxMetadata parent element directive
* rxMeta child element directive
* best if used as `rxMetadata > section > rxMeta`

Closes #885 
Closes #939 

JIRA: https://jira.rackspace.com/browse/FRMW-31

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM

### DIFF
--- a/src/rxMetadata/docs/rxMetadata.midway.js
+++ b/src/rxMetadata/docs/rxMetadata.midway.js
@@ -1,0 +1,84 @@
+var exercise = require('../rxMetadata.exercise');
+var _ = require('lodash');
+
+var transformFns = {
+    'Service Level': function (elem) {
+        return elem.getText().then(function (text) {
+            return _.zipObject(['current', 'proposed'], text.split(' → '));
+        });
+    },
+
+    'Service Type': function (elem) {
+        return elem.getText().then(function (text) {
+            return _.zipObject(['current', 'proposed'], text.split(' → '));
+        });
+    },
+
+    'Amount': function (elem) {
+        return elem.getText().then(encore.rxMisc.currencyToPennies);
+    },
+
+    'Date Field': function (elem) {
+        return elem.getText().then(function (text) {
+            return new Date(text);
+        });
+    },
+
+    'Link Field': function (elem) {
+        var promises = [elem.getText(), elem.$('a').getAttribute('href')];
+        return protractor.promise.all(promises).then(function (results) {
+            return { text: results[0], href: results[1] };
+        });
+    }
+}
+
+describe('rxMetadata', function () {
+
+    before(function () {
+        demoPage.go('#/component/rxMetadata');
+    });
+
+    describe('Status', exercise.rxMetadata({
+        present: true,
+        visible: true,
+        transformFns: transformFns,
+
+        terms: {
+            'Field Name': 'Field Value Example',
+            'Another Field Name': 'Another Field Value Example',
+            'Third Field Name': 'The Third Field Value Example',
+            'Fourth Field Name': 'The Fourth Field Value Example',
+            'Short Field Name': 'A long field value given here to show line break style.',
+            'Status': 'Active',
+            'RCN': 'RCN-555-555-555',
+            'Type': 'Cloud',
+            'Service Level': { current: 'Managed', proposed: 'Managed' },
+            'Service Type': { current: 'DevOps', proposed: 'SysOps' },
+            'Amount': 19268,
+            'Phone Number Field': '888 - 888 - 8888',
+            'Date Field': new Date('January 6, 1989'),
+            'Link Field': { text: 'Link', href: browser.baseUrl + '/#' }
+        }
+    }));
+
+    describe('rxMetadata', function () {
+        var metadata;
+
+        before(function () {
+            metadata = encore.rxMetadata.initialize($('rx-metadata'), transformFns);
+        });
+
+        it('should report back null for definitions that are not present', function () {
+            expect(metadata.term('Witty 2015 Pop Culture Reference')).to.eventually.be.null;
+        });
+
+        it('should report back null for definitions that are not displayed', function () {
+            expect(metadata.term('First Hidden')).to.eventually.be.null;
+        });
+
+        it('should report back custom return values for definitions that are not present', function () {
+            expect(metadata.term('Witty 2015 Pop Culture Reference', false)).to.eventually.be.false;
+        });
+
+    });
+});
--- a/src/rxMetadata/rxMetadata.page.js
+++ b/src/rxMetadata/rxMetadata.page.js
@@ -1,0 +1,114 @@
+/*jshint node:true*/
+var Page = require('astrolabe').Page;
+
+/**
+ * @namespace
+ */
+var rxMetadata = {
+
+    /**
+       @function
+       @param {String} metadataTerm - The term to lookup in the metadata component.
+       @returns {*} The web element of the metadata term, unless a `transformFn` is provided, or `null` if not found.
+     */
+    term: {
+        value: function (metadataTerm, fallbackReturnValue) {
+            var page = this;
+            var termElement = $('rx-meta[label="' + metadataTerm  + '"]');
+            return encore.rxMisc.unless(termElement, function (foundTermElement) {
+                var definitionElement = foundTermElement.$('.definition');
+                if (page.transformFns[metadataTerm] !== undefined) {
+                    return page.transformFns[metadataTerm](definitionElement);
+                }
+                return definitionElement.getText();
+            }, fallbackReturnValue);
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the root element is currently displayed.
+     */
+    isDisplayed: {
+        value: function () {
+            return this.rootElement.isDisplayed();
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} Whether or not the root element is present.
+     */
+    isPresent: {
+        value: function () {
+            return this.rootElement.isPresent();
+        }
+    }
+};
+
+/**
+ * @exports encore.rxMetadata
+ */
+exports.rxMetadata = {
+    /**
+     * @function
+     * @param {WebElement} rxMetadataElement - WebElement to be transformed into an rxMetadata page object.
+     * @param {Object} transformFns - An object defining which metadata entries should be transformed, and how.
+     * @returns {rxMetadata} Page object representing the rxMetadata element.
+     * @example
+     * ```js
+     * var myPage = Page.create({
+     *     metadata: {
+     *         get: function () {
+     *             // Every other metadata entry here should just be text.
+     *             // For those entries which need some extra processing, they are defined below
+     *             var transforms = {
+     *                 'Signup Date': function (elem) {
+     *                     return elem.getText().then(function (text) {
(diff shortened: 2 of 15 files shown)


CITguy:
## Three Columns
rxmetadata-local-normal

### With expanded rxCollapse
rxmetadata-local-seemore

## Two Columns
rxmetadata-local-twocolumn

## One Column
rxmetadata-local-onecolumn

glynnis:
@CITguy - @ElementE and I were talking about this, and we think for the 2 column and 1 column layouts (i.e. when columns have decreased in number for smaller screens), the "see more" links shouldn't be present in the middle of the list of metadata, i.e. a "see more" link should only every appear at the very bottom of a column.

Is it possible that the "see more" links could be expanded to reveal the additional metadata keys/values in situations where they're not at the bottom of a column? @ElementE and I don't have a good understanding of how much work this would be on the dev side. @CITguy happy to meet/discuss on vidyo if that's the most efficient way to figure out a way forward.

CITguy:
The only way to restrict content height is to apply a CSS `overflow` to the column (which gives us the option to either hide or scroll the overflow content, neither would probably be desirable).  This is why the columns do not have any height limitations.

The column wrapping/stacking is handled purely by CSS via flexbox properties.  Because of this, the `see more` functionality will have to be added manually to the end of the column.  If you wish it to display at the bottom of a column grouping, we'll have to add a design pattern to the Design Styleguide to always place the `see more` in the last column (always appearing at the bottom of a condensed parent).

I'm not sure if it's possible to implement logic to automatically add or expand `see more` functionality for columns with content that is too tall. This behaviour would be very difficult (if not impossible) to implement because it would require javascript to be aware of how the browser's CSS engine is arranging elements on the page. To my knowledge, there are no browser APIs to provide this functionality.

CITguy:
I will be out next week, so if you have any additional questions/comments/concerns. We may have to have a chat on Vidyo when I return.

src/rxMetadata/rxMetadata.page.js
```diff
@@ -0,0 +1,100 @@
...
+     * @function
+     * @param {WebElement} rxMetadataElement - WebElement to be transformed into an rxMetadata page object.
+     * @returns {rxMetadata} Page object representing the rxMetadata element.
+     */
+    initialize: function (rxMetadataElement) {
+        rxMetadata.rootElement = {
+            get: function () { return rxMetadataElement; }
+        };
+        return Page.create(rxMetadata);
+    },
```
Droogans: From what I gather this expects a single term/definition pair to be passed in. I'm going to modify this a bit so that it accepts an entire block of metadata elements and fetches key/value pairs for you.

Droogans:
@CITguy I rebased on master to get the new goodness in rxMisc page objects for this work. **I force pushed over your branch!**

I did this because I am both a rebel who lives on the edge and a considerate co-worker who wanted to get this done while you were away this week. Very yin/yang stuff there.

Again, **I force pushed over your branch on your remote**, @CITguy.

Droogans:
LGTM.

glynnis:
Design LGTM from @ElementE @lfielder and I. We met with @CITguy to discuss, and concluded that the solution here on the "see more"/"see less" front with flexbox wrapping is to update documentation for "see more"/"see less" to indicate it should only be used on the final column, or not at all. 

Docs update will come in a separate PR. @CITguy will update to remove the "see more/see less" from this demo and we'll be good to go! :clap: :clap: yay for beautiful metadata!

CITguy:
* "see more/see less" removed from demo
* screenshots updated

CITguy:
fixed broken midway

Droogans:
:+1: Anxious for this to hit rx-page-objects.

src/rxMisc/docs/rxMisc.midway.js
```diff
@@ -174,8 +174,10 @@ describe('rxMisc', function () {
             });
 
             it('should return the result of a custom function', function () {
-                var reverseIt = function (text) {
-                    return text.split('').reverse().join('');
+                var reverseIt = function (elem) {
+                    return elem.getText().then(function (text) {
+                        return text.split('').reverse().join('');
```
chriscantu: @CITguy Net picky, but this will throw an error if undefined or null.  I know its just an example though.

chriscantu:
@CITguy LGTM.

src/rxMisc/docs/rxMisc.midway.js
CITguy: @Droogans would know more about this, but I believe that the `getText()` function on protractor should always return a string.

src/rxMetadata/rxMetadata.page.js
```diff
@@ -0,0 +1,114 @@
...
+       @function
+       @param {String} metadataTerm - The term to lookup in the metadata component.
+       @returns {*} The web element of the metadata term, unless a `transformFn` is provided, or `null` if not found.
+     */
+    term: {
+        value: function (metadataTerm, fallbackReturnValue) {
+            var page = this;
+            var termElement = $('rx-meta[label="' + metadataTerm  + '"]');
+            return encore.rxMisc.unless(termElement, function (foundTermElement) {
+                var definitionElement = foundTermElement.$('.definition');
```
parlarjb: I'm sure I'm misisng something, but doesn't the `foundTermElement` value represent `termElement.getText()`? How are we able to do a `.$` on that?

parlarjb: Nevermind, I RTFM https://angular.github.io/protractor/#/api?view=webdriver.WebElement.prototype.getText and it says `innerText of this element, including sub-elements,`

src/rxMetadata/docs/rxMetadata.midway.js
```diff
@@ -0,0 +1,84 @@
...
+        });
+    },
+
+    'Link Field': function (elem) {
+        var promises = [elem.getText(), elem.$('a').getAttribute('href')];
+        return protractor.promise.all(promises).then(function (results) {
+            return { text: results[0], href: results[1] };
+        });
+    }
+}
```
parlarjb: Since every one of these `transformFns` immediately calls `elem.getText()`, what about just passing in that instead of `elem`? Do you think there will be an instance where the actual `elem` will be needed and not the `elem.getText()` promise?

CITguy: @Droogans, thoughts?

CITguy: I would think that it might be possible, but I don't know if it's important enough to block the PR.

parlarjb: I don't want to block the PR on it, it was just a thought

parlarjb:
LGTM

CITguy:
YAY! :tada:

src/rxMetadata/rxMetadata.page.js
Droogans: That's crazy that you say that because it used to be that this `unless` function I wrote (initially in Billing) only needed to use `.getText`, so yes, it was actually pulling in text.

I switched it to the web element when I ported it over to enocre.rxMisc so that you could get hrefs, class names, etc. instead of just text.

src/rxMetadata/docs/rxMetadata.midway.js
Droogans: It used to be text! I thought it was the way to go, it was all working, until I wanted to pull the href from a metadata term and found I'd need the web element instead. So yeah, I really wanted it to be text too, and it was initially, but I felt it's better to force everyone to write an extra five or six lines here or there in order to support the entire web element API, not just text manipulation.

src/rxMisc/docs/rxMisc.midway.js
Droogans: @CITguy @chriscantu the way this works is that the function object `reverseIt` is passed in as a parameter to `rxMisc.unless`. Unless expects 1) a web element, 2) a function to apply against the web element, and 3) some default return value if the web element (from number 1) is not present, not displayed, or has text that matches some value you can define elsewhere.

The idea being that you can say "Hey, if billing termination date says 'N/A', that means that there is no termination date, so just return `null`".

This function `reverseIt` will not execute unless there is a web element that is present, that web element is displayed, and the text of the element doesn't contain the strings `['Not found in account', 'Loading...', 'N/A']` (those are the strings from Billing, for example).

`elem` will never be undefined if used inside of `rxMisc.unless`. If it were, [it'd exit early and return `null` right away](https://github.com/rackerlabs/encore-ui/blob/4cfae33823a6f532e930e109192da76dfdb3f61b/src/rxMisc/rxMisc.page.js#L101-L124).

src/rxMisc/rxMisc.page.js
```diff
@@ -70,7 +70,7 @@ exports.rxMisc = {
        This is useful when applications feature use of `ng-if` to control various messages to the user.
        @function
        @param {WebElement} elem - The web element that may or may not be present, or displayed, or valid.
-       @param {Function} innerFn - Function to call on the element's text should it be present, displayed, and valid.
+       @param {Function} innerFn - Function to call on the element should it be present, displayed, and valid.
```
Droogans: Update this to mark `innerFn` as an optional parameter. It defaults to the element's inner text if left undefined.

PR #1112, opened 2015-07-29
docs(rxCollapse): Add more prescriptive rules around use of see more

glynnis:
For more context, see: https://github.com/rackerlabs/encore-ui/pull/1107#issuecomment-125983838

We should also add this kind of documentation to the Style Guide in the metadata section. Issue for that here: https://jira.rax.io/browse/FRMW-97

### DIFF
--- a/src/rxCollapse/README.md
+++ b/src/rxCollapse/README.md
@@ -2,6 +2,13 @@
 
 The rxCollapse element can be used to display and hide content. It can be configured to show as either expanded or collapsed on page load. A double chevron is used as a toggle to show/hide the inner contents, while keeping the header and border visible.
 
+### See More/See Less
+
+- This pattern was developed for areas displaying metadata that may be short on screen real estate, as a way to hide data on load that is not as important to the user in the context where they are presented.
+- This pattern is not very responsive-friendly, since as browser width decreases, columns will wrap. As columns wrap, the "See More" rxCollapse elements get lost in the new context, which is bad for user experience.
+- To avoid the problem described above, "See More" rxCollapse elements should only be used at the end of the final column present on the page, so that when the columns wrap via flexbox, "See More" is always last and doesn't get lost in between metadata key/value pairs.
+
+
 ### Attributes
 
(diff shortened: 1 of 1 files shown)


parlarjb:
LGTM

src/rxCollapse/README.md
```diff
@@ -2,7 +2,14 @@
 
 The rxCollapse element can be used to display and hide content. It can be configured to show as either expanded or collapsed on page load. A double chevron is used as a toggle to show/hide the inner contents, while keeping the header and border visible.
 
+### See More/See Less
+
+- This pattern was developed for areas that may be short on screen real estate, as a way to hide metadata key/value pairs on load that are not as important to the user in the context where they are presented.
+- This pattern is not very responsive-friendly, since as browser width decreases, columns will wrap. As columns wrap, the "See More" rxCollapse elements get lost in the new context, which is bad for user experience.
+- To avoid the problem described above, "See More" rxCollapse elements should only be used at the end of the final column present on the page, so that when the columns wrap via flexbox, "See More" is always last and doesn't get lost in between metadata key/value pairs.
+
+
```
CITguy: Should we mention that this pattern currently only applies to metadata implementation?

glynnis: I figured that was indicated by the section title, "See More/See Less" -- would something different be more clear?

CITguy: Well, you mention in line 7 that it was `developed for areas that may be short on screen real estate` (not specific to metadata), but in the same line you mention `as a way to hide metadata key/value pairs`.  It seems a bit confusing.  Maybe change the first bullet to:

```
- This pattern was developed for areas that may be short on screen real estate, as a way to hide data on load that is not as important to the user in the context where they are presented.
```

CITguy: If I'm confusing you (which I might), let's get a quick vidyo chat to clarify things.

glynnis: Updated! :+1:

CITguy:
LGTM

PR #1113, opened 2015-07-29
fix(accountUsers): fixing looping switch to admin issue

kyleondata:
We can achieve the same overall goal by sorting the options in the
dropdown by admin instead of using `switchUser`.

JIRA: relates to https://jira.rax.io/browse/EN-1912

- [x] Dev LGTM
- [ ] Dev LGTM
- [ ] QE LGTM

### DIFF
--- a/src/rxApp/rxApp.js
+++ b/src/rxApp/rxApp.js
@@ -438,19 +438,6 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
         link: function (scope) {
             scope.isCloudProduct = false;
 
-            // This function is attached to the scope for the sole purpose of making
-            // it easier to test this functionality. A reorganization of this and/or
-            // the tests is needed in order to pull this off the scope.
-            scope.switchToAdmin = function () {
-                // If the user in the params is not the admin swtich to the admin
-                // this causes the $route.current.params.user to become the admin user
-                var adminUser = _.first(_.where(scope.users, { admin: true }));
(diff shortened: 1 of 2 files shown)


src/rxApp/rxApp.js
```diff
@@ -463,17 +450,23 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
...
+            
             // We use $route.current.params instead of $routeParams because
             // the former is always available, while $routeParams only gets populated
             // after the route has successfully resolved. See the Angular docs on $routeParams
             // for more details.
             var loadUsers = function () {
                 var success = function (account) {
+                    
+                    // Sort the list so admins are at the top of the array
+                    account.users = _.sortBy(account.users, function (user) {
```
halkon: Does `_.sortBy(..., 'admin')` do the right thing? seems overkill to create a function for just one attribute

parlarjb: I agree with Freddy. `_.sortBy` should automatically do the correct pluck if you call it like he showed

parlarjb:
My usual question: Have you tested this change locally against Cloud?

kyleondata:
Yes sir

src/rxApp/rxApp.js
kyleondata: Good point.

kyleondata:
@halkon @parlarjb bits updated.

halkon:
LGTM

parlarjb:
LGTM

PR #1121, opened 2015-08-05
test(rxSelect): Allow for optional `slowClick`

### LINKED ISSUE #1105: rxSelect needs to support non-slowClick for select function.
I just fixed a very odd interaction in an app that included a dropdown value change that would somehow click a tab instead of the desired option in the dropdown.

I fixed it by imitating everything about `rxSelect.dropdown.option.select`, but without the `rxMisc.slowClick` part.

Think of a good way of handling this for users who find themselves not needing the slow click functionality. Perhaps a new call?

It has to reach all the way up to `generateAccessor`, if possible.

Droogans:
Closes #1105

### DIFF
--- a/src/rxSelect/docs/rxSelect.midway.js
+++ b/src/rxSelect/docs/rxSelect.midway.js
@@ -132,7 +132,8 @@ describe('rxSelect', function () {
             });
 
             afterEach(function () {
-                subject.select('I do not like bacon');
+                var slowClick = false;
+                subject.select('I do not like bacon', slowClick);
             });
 
             it('should be valid', function () {
@@ -173,7 +174,8 @@ describe('rxSelect', function () {
             });
 
             it('should remove the select element to the DOM', function () {
-                willHide.select('Hide Next Select Box');
+                var slowClick = false;
+                willHide.select('Hide Next Select Box', slowClick);
                 expect(willBeHidden.isPresent()).to.eventually.be.false;
             });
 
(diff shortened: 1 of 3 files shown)


Droogans:
Pretty sure this is incompatible with the form filling stuff in `rxForm.fill`. It'll think that the `{ option: 'Text', slowClick: false }` part are namespaced page objects under whatever your dropdown is called, and not a literal object to pass to the `rxForm.dropdown` setter.

Droogans:
Nope this won't you'll just need to specify that all interactions with a certain dropdown should not use slow clicking at all at the page object level.

Bad:

```js
myPage.form = {
    state: { option: 'Texas', slowClick: false }
};
```

Good:

```js
var myPage = Page.create({
    state: encore.rxForm.dropdown.generateAccessor($('#state-input'), false);
});

// later...
myPage.form = {
    state: 'Texas'
};
```

src/rxSelect/docs/rxSelect.midway.js
```diff
@@ -132,7 +132,8 @@ describe('rxSelect', function () {
             });
 
             afterEach(function () {
-                subject.select('I do not like bacon');
+                var slowClick = false;
+                subject.select('I do not like bacon', slowClick);
```
parlarjb: Thank you for not passing a bare `false` to the method :heart:

parlarjb:
LGTM

CITguy:
LGTM

PR #1126, opened 2015-08-07
style(rxMetadata): Add data and link example to demo

### LINKED ISSUE #1125: Update metadata demo to include example with data and link
Some instances of metadata usage need to include in-line actions, like on the LBaaS details page. We should include an example of this in the demo, as we recently got a question via chat about what format would be correct for this pattern.

glynnis:
Closes #1125

### Before:
screen shot 2015-08-07 at 1 42 49 pm

### After:
screen shot 2015-08-07 at 1 42 36 pm

 - [X] Dev LGTM
 - [x] Dev LGTM

### DIFF
--- a/src/rxMetadata/docs/rxMetadata.html
+++ b/src/rxMetadata/docs/rxMetadata.html
@@ -26,6 +26,9 @@
         <rx-meta label="Link Field">
           <a href="#">Link</a>
         </rx-meta>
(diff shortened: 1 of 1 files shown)


parlarjb:
LGTM

CITguy:
LGTM

PR #1134, opened 2015-08-12
refact(grunt): Responsive shipit squirrel

Droogans:
This will need some manual testing since I think lodash needs to be more up to date to get this to use https://lodash.com/docs#repeat


TheNando:
LGTM

Droogans:
Alright I tested it it doesn't work unless you mess with the version of lodash, and many packages complain.

I'll roll my own `_.repeat` using `_.fill(Array(len), '_').join('')`. Ugly, but it works.

PR #1137, opened 2015-08-13
docs(VERSIONING): clarification on npm adduser

CITguy:
`npm adduser` (by default) will add the given user credentials to the currently set registry. Given that a user may have the internal npm registry set as their default, it would add the credentials to the incorrect registry.  This will make sure that the public npmjs registry is the target for the additional credentials.

### DIFF
--- a/guides/versioning.md
+++ b/guides/versioning.md
@@ -97,7 +97,7 @@ module.export = {
 To publish `rx-page-objects` to the public npm registry.
 
 0. `npm get registry` should return `https://registry.npmjs.org/`
-1. `npm adduser`
+1. `npm adduser --registry=https://registry.npmjs.org/`
   * Username: `encore-ui`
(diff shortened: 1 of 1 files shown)


TheNando:
LGTM

guides/versioning.md
```diff
@@ -97,7 +97,7 @@ module.export = {
 To publish `rx-page-objects` to the public npm registry.
 
 0. `npm get registry` should return `https://registry.npmjs.org/`
-1. `npm adduser`
+1. `npm adduser --registry=https://registry.npmjs.org/`
```
Droogans: Better safe than sorry.

Droogans:
That branch name though.

CITguy:
branch name was automatically generated via github ui

Droogans:
Yep. The pencil is mightier than the editor.

PR #1142, opened 2015-08-14
feat(rxTags): Add rxTags component.

### LINKED ISSUE #918: Create 'tags' component
Here's the look/feel for tags:

tags

Interaction design for this is based on https://material.angularjs.org/#/demo/material.components.chips, which should be used as a base for building a tags component, since it provides good flexibility with read-only functionality, accommodates hierarchy through use of categories, etc. 

The main thing not provided by Material is a type-ahead drop down, but after consulting with @TheNando it sounds like that won't be too difficult to add and customize. See https://github.com/rackerlabs/encore-ui/issues/909, which should be used by default for this component.

megalord:
Fixes #918 

- [x] Dev LGTM
- [ ] Dev LGTM
- [x] Design LGTM

screen shot 2015-08-14 at 1 56 35 am

@glynnis This does not include the feature to show the list on click.  I started down that path, and it's a bit complex.  I am also going to be out next week, so someone else can have a go or I will look at it again down the road.

### DIFF
--- a/src/rxTags/rxTags.exercise.js
+++ b/src/rxTags/rxTags.exercise.js
@@ -1,0 +1,67 @@
+var rxTags = require('./rxTags.page').rxTags;
+
+/**
+   rxTags exercises.
+   @exports encore.exercise.rxTags
+   @param {Object} [options=] - Test options. Used to build valid tests.
+   @param {rxTags} [options.instance=] - Component to Exercise.
+   @param {string} options.sampleText - A tag that can be added.
+   @example
+   ```js
+   describe('default exercises', encore.exercise.rxTags({
+       instance: encore.rxTags.initialize('.demo rx-tags') // select one of many widgets on page
+   }));
+   ```
+ */
+exports.rxTags = function (options) {
+    if (options === undefined) {
+        options = {};
+    }
+
+    options = _.defaults(options, {
+        sampleText: undefined
+    });
+
+    return function () {
+        var component, tag, numTags;
+
+        before(function () {
+            if (options.instance !== undefined) {
+                component = options.instance;
+            } else {
+                component = rxTags.main;
+            }
+            component.count().then(function (num) {
+                numTags = num;
+            });
+        });
+
+        if (!_.isUndefined(options.sampleText)) {
+            it('adds a tag', function () {
+                tag = component.addTag(options.sampleText);
+                expect(tag.text).to.eventually.equal(options.sampleText);
+                expect(component.count()).to.eventually.equal(numTags + 1);
+            });
+
+            it('removes a tag by clicking the x', function () {
+                tag.remove();
+                expect(tag.exists()).to.eventually.be.false;
+                expect(component.count()).to.eventually.equal(numTags);
+            });
+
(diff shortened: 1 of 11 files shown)


src/rxTags/rxTags.exercise.js
```diff
@@ -0,0 +1,67 @@
...
+    return function () {
+        var component, tag, numTags;
+
+        before(function () {
+            if (options.instance !== undefined) {
+                component = options.instance;
+            } else {
+                component = rxTags.main;
+            }
+            component.count.then(function (num) {
```
Droogans: I know `count` is a property, but it should mirror protractor's `count()` function.

Every other component does this. `:/`

megalord: oops my bad

src/rxTags/templates/rxTags.html
```diff
@@ -0,0 +1,18 @@
+<div class="rx-tags" ng-click="focusInput($event)">
+    <div class="tag"
+        ng-repeat="tag in tags track by tag.text"
+        ng-keydown="removeIfBackspace($event, tag)"
+        tabindex="0">
+        <i class="fa fa-tag"></i>
+        <span class="text">{{tag.text}}</span>
+        <span class="category">({{tag.category}})</span>
+        <i class="fa fa-times" ng-click="remove(tag)"></i>
+    </div>
```
CITguy: Can we make this div an `<rx-tag>` element directive?  I can see developers wanting to display these tags outside of a form.

megalord: How would that differ from using the `disabled` attribute?  Even then, I suppose it's nice not to have to create an `ng-model`.

CITguy: nevermind, I think I misinterpreted what the directive does

CITguy: However, it probably wouldn't be a bad idea to add a disabled state example to the demo.

megalord:
The only test failure is that brittle test for `rxNotify` where a stale element error is thrown.

Droogans:
@CITguy rebuilding the Travis job to see what happens when the visual regression tests see the exact same SHA and the same branch with an existing PR open against it.

My guess is that it'll fail to push it up? I'm not sure what'll happen.

Droogans:
@megalord can you please push up a new commit?

All you need to do is

```bash
git commit --amend
Waiting for ${EDITOR}...
# save, and close, no changes
git push -f [origin|upstream] rx-tags
```

CITguy:
LGTM

glynnis:
Design LGTM. @CITguy would you mind creating a story for the remaining work in a way that will make sense to devs? Here's what's left:

> We also need a version of rxTags that reveals the list on click (like the second example on 
> http://rackerlabs.github.io/encore-ui/#/components/typeahead). Requiring the user to type should be 
> the default, but it would be nice to allow consumers of the framework an option to use the second 
> example.

CITguy:
@glynnis, https://jira.rax.io/browse/FRMW-207

@megalord, please squash

PR #1147, opened 2015-08-17
chore(headings): Deprecate .title classes/subclasses

CITguy:
* added base H1-H6 styles above legacy .title styles
  * allows for backward compatibility as developers update their code
* added namespace for upcoming work on new styleguide pages

JIRA: https://jira.rax.io/browse/FRMW-37

- [x] Dev LGTM
- [ ] Dev LGTM
- [x] Design LGTM

PR #1148, opened 2015-08-18
chore(rxApp): add visual regression

CITguy:
Adding visual regression for standard and custom app definitions of `rxApp`.

### DIFF
--- a/utils/visual-regression/components/rxApp.midway.js
+++ b/utils/visual-regression/components/rxApp.midway.js
@@ -1,0 +1,11 @@
+describe('demo component', function () {
+
+    before(function () {
+        demoPage.go('#/component/rxApp');
+    });
+
+    it('custom app', function () {
+        screenshot.snap(this, $('#custom-rxApp'), { threshold: 1 });
+    });
+
+});


utils/visual-regression/components/rxApp.midway.js
```diff
@@ -0,0 +1,15 @@
+describe('demo component', function () {
+
+    before(function () {
+        demoPage.go('#/component/rxApp');
+    });
+
+    it('standard app', function () {
+        screenshot.snap(this, $('#standard-rxApp'));
```
Droogans: It may make sense to use this instead.

```js
screenshot.snap(this, $('#standard-rxApp'), { threshold: 1 });
```

There is a pretty large area involved here. You'll want to trigger a visual regression for even small changes here.

utils/visual-regression/components/rxApp.midway.js
```diff
@@ -0,0 +1,15 @@
...
+    });
+
+    it('custom app', function () {
+        screenshot.snap(this, $('#custom-rxApp'));
```
Droogans: Same as above here.

```js
screenshot.snap(this, $('#custom-rxApp'), { threshold: 1 });
```

CITguy:
Updated to only include `#custom-rxApp` as per @parlarjb's point of `#standard-rxApp` updating on every nav change. Visual regression will still work on the custom app setup w.r.t. #1147.

parlarjb:
LGTM

PR #1149, opened 2015-08-19
Fix gh pages links

Droogans:
This should make nesting pages under an existing breadcrumb much simpler. Perhaps more of the style guide can become hierarchical in the future.

This also makes testing against the `ghPages` environment simpler. By changing the regular expression to match `/localhost/`, you can simulate what the url will look like when published to Github pages.


Droogans:
Marking urgent from @chriscantu's feedback.

parlarjb:
LGTM

chriscantu:
LGTM

CITguy:
LGTM

PR #1157, opened 2015-08-21
fix(rxAccountUsers): Fix ever increasing user calls

### LINKED ISSUE #1156: rxAccountUsers does not properly release watcher
When the directive rxAccountUsers unloads, it does not properly unregister its rootScope watcher.  This results in stacking calls to the encore account users service.

The call in question is on line https://github.com/rackerlabs/encore-ui/blob/master/src/rxApp/rxApp.js#L503

jj4th:
This is the PR to fix #1156 

- [X] Dev LGTM
- [x] Dev LGTM

### DIFF
--- a/src/rxApp/rxApp.js
+++ b/src/rxApp/rxApp.js
@@ -435,7 +435,7 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
     return {
         restrict: 'E',
         templateUrl: 'templates/rxAccountUsers.html',
-        link: function (scope) {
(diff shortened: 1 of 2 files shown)


CITguy:
@jj4th: can you update the tests for this change?

jj4th:
There are no tests to update for this change :)

CITguy:
Sorry, I meant update via adding a test for this.

jj4th:
Added a unit test :smile_cat:

parlarjb:
LGTM

CITguy:
LGTM

PR #1162, opened 2015-08-27
feat(userVoiceFeedback): User Voice redirection service

halkon:
No JIRA

- [x] Dev LGTM
- [ ] Dev LGTM
- [ ] Design LGTM

This overrides the behavior of `rxFeedback` when a user chooses `Feature Request`, they will be forwarded to the internal **GET Feedback** 

Depending on which application we are on, there will be a category ID for each, and the redirect will happen to that category.

Allowing to override the base URL for **GET Feedback** as well as how the categories get appended on to the URL, in order to allow for change without having to re-release the framework again.

### DIFF
--- a/grunt-tasks/options/connect.js
+++ b/grunt-tasks/options/connect.js
@@ -22,6 +22,17 @@ module.exports = {
         options: {
             middleware: function (cnct) {
                 return [
+                    function (req, res, next) {
+                        if (req.url.indexOf('/encore/feedback') === 0) {
+                            var response = JSON.stringify({ base: 'https://angularjs.org/' });
+                            res.setHeader('Content-Type', 'application/json');
+                            res.setHeader('Content-Length', response.length);
+                            res.statusCode = 200;
+                            res.end(response);
+                            return false;
+                        }
+                        return next();
+                    },
                     config.proxyRequest,
                     config.modRewrite([
                         'login.html /login.html [L]',
@@ -40,7 +51,22 @@ module.exports = {
     keepalive: {
         options: {
             keepalive: true,
-            base: '<%= config.docs %>'
+            middleware: function (cnct) {
+                return [
+                    function (req, res, next) {
+                        if (req.url.indexOf('/encore/feedback') === 0) {
+                            var response = JSON.stringify({ base: 'https://angularjs.org/' });
+                            res.setHeader('Content-Type', 'application/json');
+                            res.setHeader('Content-Length', response.length);
+                            res.statusCode = 200;
+                            res.end(response);
+                            return false;
+                        }
+                        return next();
+                    },
+                    config.mountFolder(cnct, config.docs)
+                ];
+            }
         }
     }
 };
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "grunt-bump": "~0.3.0",
     "grunt-cloudfiles": "~0.3.0",
     "grunt-complexity": "~0.2.0",
-    "grunt-connect-proxy": "0.2.0",
+    "grunt-connect-proxy": "0.1.10",
     "grunt-contrib-clean": "~0.6.0",
     "grunt-contrib-concat": "~0.5.0",
     "grunt-contrib-connect": "~0.9.0",
--- a/src/rxFeedback/README.md
+++ b/src/rxFeedback/README.md
@@ -4,10 +4,21 @@ Component built to gather and send user feedback
 
 ## Default Submission Function
 
-The `rxFeedback` component sends feedback to `/api/encore/feedback`, which routes feedback to `encoreui@lists`. 
-This endpoint also supports a `product` parameter `/api/encore/feedback/:product` for sending feedback to a 
-product-specific mailing list. Adding a custom endpoint is managed in `encore-service-pillar`. Once configured 
+The `rxFeedback` component sends feedback to `/api/encore/feedback`, which routes feedback to `encoreui@lists`.
+For the **Feedback Type** of "Feature Request", we will open up a new window redirecting the user to the **GET Feedback** website, which will now host all internal requests for features.
+
+This endpoint also supports a `product` parameter `/api/encore/feedback/:product` for sending feedback to a
+product-specific mailing list.
+
+Adding a custom endpoint is managed in `encore-service-pillar`. Once configured
 you can override the default endpoint with `rxFeedbackSvc.setEndpoint`.
+
+```javascript
+    // To be put in the run section of your application
+    .run(function (rxFeedbackSvc) {
+        rxFeedbackSvc.setEndpoint('/api/encore/feedback/cloud');
+    });
+```
 
 ## Custom Submission Function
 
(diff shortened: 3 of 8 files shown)


src/rxFeedback/rxFeedback.spec.js
```diff
@@ -70,7 +72,7 @@ describe('rxFeedback', function () {
     });
 
     it('should set the current url of the page on the modal\'s scope', function () {
-        var modalScope = {};
+        var modalScope = { '$watch': sinon.stub() };
```
megalord: For what it's worth, quotes are not needed for a leading `$`.

src/rxFeedback/rxFeedback.spec.js
```diff
@@ -153,6 +155,10 @@ describe('rxFeedback', function () {
         expect(notifySvcMock.add.args[0][0]).to.contain(customFailure);
     });
 
+    it('should attempt to redirect a user when feature requests are chosen', function () {
+
```
megalord: Is this supposed to be empty?

src/rxFeedback/rxFeedback.js
```diff
@@ -86,7 +87,103 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
...
+            }).then(function (response) {
+                // We only want the data after this point
+                return response.data;
+            }, function () {
+                // Connection failed to CDN? return just the default URL as it's base
+                // This overrides the promise failure and trickles down as a succesful execution
+                return $q.when({
+                    base: defaultUserVoiceURL,
+                    categoryPrefix: defaultCategoryPrefix
+                });
```
megalord: This strikes me as a little strange considering the fallbacks below (`var url = mapping.base || defaultUserVoiceURL;`).  If an empty object were returned here, the result would be the same, and now there is one less place in the code that is dependent on the schema of the response.

Also, why is `$q.when()` necessary?  I expect that just returning the option would have the same result.

src/rxFeedback/rxFeedback.spec.js
halkon: bad left-over, thanks

src/rxFeedback/rxFeedback.spec.js
halkon: Thanks :grin:

src/rxFeedback/rxFeedback.js
```diff
@@ -86,7 +87,103 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
...
+        }
+        return url;
+    };
+
+    // Open a new window with the defined route
+    var openWindow = function (scope, route) {
+        scope.loadingUserVoice = true;
+        openPromise = $interval(function () {
+            $window.open(route, '_blank');
+            return route;
```
megalord: Does this return value get used anywhere?

src/rxFeedback/rxFeedback.js
```diff
@@ -96,8 +193,13 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
         link: function (scope) {
             scope.feedbackTypes = feedbackTypes;
 
+            scope.cancelOpen = function () {
+                UserVoiceMapping.cancelOpen(scope);
```
megalord: `cancelOpen()` doesn't take any parameters.

src/rxFeedback/rxFeedback.js
halkon: Good call, when I originally wrote it I didn't have the fallbacks, and you're right no need for `$q.when`.  Will update, thanks!

src/rxFeedback/rxFeedback.js
halkon: Nope, left over from copy pasta.

src/rxFeedback/rxFeedback.js
halkon: Man, I need to do better code clean up :grin: thanks

halkon:
To note I will also work on midways, I just wanted to get the service out there first to be looked at (thanks Jordan!)

megalord:
Perhaps you have also already planned to update the docs, but I think they could use a little love with this PR - just something so that the devs aren't surprised when they see this behavior for the first time.  Also, if there is a process for getting an application added to that json mapping file, that should be documented somewhere.

halkon:
@megalord you're right again, docs and midways are planned in the updates, I have just had this in dev mode for too long and needed dev feedback. 

In regards to updating the json mapping file, it is isolated at the moment at the CDN level and haven't proposed a workflow for updating it, as the closest thing to that is the nav and well, we want to take that out of travis. 

Also worth nothing, there are no environment specific things in this due to the fact that there are no environments in the **GET Feedback** user voice instance.  

The other thing worth noting is that, a lot of this really sniffs of something that ought to be in `encore-ui-svcs` rather than here, however given the nature that people have to upgrade to the latest release of the framework to receive this (or perhaps do some copy pasting of modules to integrate it if they don't want to upgrade) I decided to put it in here.  

It is not generic behavior at all, which is what bothers me.

halkon:
@megalord comments addressed.  Thank you!

megalord:
Sure thing; I'll check back in once the docs and midways are added.  Looks good so far!

Droogans:
@halkon the failing tests are no longer valid ones. They may need to be deleted.

```
  1) rxFeedback feedback types and labels should have the correct label set for descriptionLabel:
     NoSuchElementError: No element found using locator: By.cssSelector(".feedback-description")
      at Array.forEach (native)
  Error
      at Fn.rxFeedback.descriptionLabel.get (utils/rx-page-objects/index.js:1566:64)
      at Context.<anonymous> (src/rxFeedback/docs/rxFeedback.midway.js:69:46)
  From: Task: rxFeedback feedback types and labels should have the correct label set for descriptionLabel
  2) rxFeedback feedback types and labels should have the correct label set for descriptionPlaceholder:
     NoSuchElementError: No element found using locator: by.model("fields.description")
      at Array.forEach (native)
  Error
      at Fn.rxFeedback.descriptionPlaceholder.get (utils/rx-page-objects/index.js:1560:37)
      at Context.<anonymous> (src/rxFeedback/docs/rxFeedback.midway.js:69:46)
  From: Task: rxFeedback feedback types and labels should have the correct label set for descriptionPlaceholder
```

halkon:
@Droogans the midways specific to this are not even fully done.

halkon:
As in:
```javascript
describe('feedback types and labels', function () {
        var typesAndLabels = {
            'Incorrect Data': {
                descriptionLabel: 'Problem Description:',
                descriptionPlaceholder: ['Please be as descriptive as possible ',
                                         'so we can figure it out for you.'].join('')
            },
            'Feature Request': {
                descriptionLabel: 'Feature Description:',
                descriptionPlaceholder: ['Please be as descriptive as possible ',
                                         'so we can make your feature awesome.'].join('')
            },
            'Kudos': {
                descriptionLabel: 'What made you happy?:',
                descriptionPlaceholder: ['We love to hear that you\'re enjoying Encore! ',
                                         'Tell us what you like, and what we can do to ',
                                         'make it even better'].join('')
            }
        };

        it('should have all feedback types', function () {
            var types = [defaultFeedback].concat(_.keys(typesAndLabels));
            expect(successfulFeedback.types).to.eventually.eql(types);
        });

        _.forEach(typesAndLabels, function (typeData, type) {
            it('should switch feedback types', function () {
                successfulFeedback.type = type;
                expect(successfulFeedback.type).to.eventually.equal(type);
            });

            _.forEach(typeData, function (text, property) {
                it('should have the correct label set for ' + property, function () {
                    expect(successfulFeedback[property]).to.eventually.equal(text);
                });
            });

        });

    });```

**Comment by halkon** [comment]
nm
z1eiy

File: package.json [hunk 6]
```diff
@@ -31,7 +31,7 @@
     "grunt-bump": "~0.3.0",
     "grunt-cloudfiles": "~0.3.0",
     "grunt-complexity": "~0.2.0",
-    "grunt-connect-proxy": "0.2.0",
+    "grunt-connect-proxy": "0.1.10",
```
CITguy: Why the downgrade?

halkon: issues with proxying to `https` from `http`

halkon: For reading purposes:

https://github.com/drewzboto/grunt-connect-proxy/issues/79
https://github.com/drewzboto/grunt-connect-proxy/issues/71

halkon:
@megalord added a first pass of the documentation, one of the asks was to see if we could make it so that they could copy/paste some code and not have to upgrade to the latest version of the framework, I'd like to get your thoughts (as well as @parlarjb @CITguy @Droogans @TheNando )

I also realized I have not posted screenshots of this...

halkon:
**Before canceling/opening window**
image
**After canceling/opening window**
image
**Docs:**
image
----- And
image

src/rxFeedback/rxFeedback.js
```diff
@@ -86,7 +87,120 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
 
     return container;
 })
-.directive('rxFeedback', function (feedbackTypes, $location, rxFeedbackSvc, rxScreenshotSvc, rxNotify, Session) {
+.factory('UserVoiceMapping', function ($q, $http, $location, $interval, $window) {
```
megalord: `$q` isn't being used :sweat_smile:

megalord:
LGTM
I'm concerned that the whole copy the code thing could set a bad precedent, and we really need to solve this at a build level.  But without an immediate solution, I suppose this will have to do.

Following up on a prior comment, is the temporary procedure for adding an app to the route mapping just to ask?

halkon:
Update to docs:

image

src/rxFeedback/rxFeedback.js
halkon: Good catch, :grin:

halkon:
@megalord I do not like the copy paste solution, but for now it's just something as they're pushing this feature through. 

Updated docs & code to get rid of the pesky `$q` :laughing:

src/rxFeedback/README.md
```diff
@@ -26,3 +37,74 @@ feedback object with the following definition:
...
 }
 ```
+
+## UserVoice Redirect Integration
+
+### Development
+For development purposes, you may want to include one of the two following configurations depending on which type of project you have:
+
+**Gulp**: `gulp/util/prism.js`
+```javascript
```
CITguy: Maybe add a blurb that the latest version of the generator will include this proxy.

halkon: Done

src/rxFeedback/README.md
```diff
@@ -26,3 +37,76 @@ feedback object with the following definition:
   "description": "(string) user-submitted feedback"
 }
 ```
+
+## UserVoice Redirect Integration
+
+### Development
+For development purposes, you may want to include one of the two following configurations depending on which type of project you have:
+
+*The latest version of the [encore-generator](https://github.com/rackerlabs/generator-encore) will include this proxy*
```
CITguy: no hyphen in `encore generator`

halkon: Fixed

halkon:
Totally rebased/squashed

halkon:
Updated the README with a CDN URL to the "code" they can include

src/rxFeedback/rxFeedback.js
```diff
@@ -86,7 +87,122 @@ angular.module('encore.ui.rxFeedback', ['ngResource'])
 
     return container;
 })
-.directive('rxFeedback', function (feedbackTypes, $location, rxFeedbackSvc, rxScreenshotSvc, rxNotify, Session) {
+.factory('UserVoiceMapping', function ($http, $location, $interval, $window) {
+
+    var defaultUserVoiceURL = 'https://get.feedback.rackspace.com/forums/297396';
```
CITguy: Should this private URL be hardcoded in a public framework?

CITguy: Can we set this as a `.value()` in `encore-ui-svcs`?  Then default to current functionality if it is not defined or blank.

megalord: I'm curious to know if this is even possible.  If we are considering usage without `encore-ui-svcs`, then an attempt to inject the value service would cause an injector error.  The only way I can think of is to put the `UserVoiceMapping` service in `encore-ui-svcs` along with a decorator for `rxFeedback`.  That way, the `rxFeedback` that ships with the framework doesn't rely on it at all.  It's a lot of effort for an idea that I'm not sure matters.  Are we still upholding to the *public* framework idea, and why?

Any other ideas?

halkon: You both have a very great point and it was something that was an afterthought for me after I had posted the PR In the first place. 

The services project is where this should live and the decorator override is where this should be implemented for those that don't want to upgrade the framework. 

However that is the issue, this whole thing has been an encapsulated issue therefore it stayed in the framework,  I am totally fine with this going in the services project.

The main projects that need to update to this are billing, account, cloud, and ticketing. 

If we feel like it's needed I'll move it. 

I also agree I don't want this URL on here, as it's a VPN only URL.  If I get the agreement from you both, we'll do it.  

Thank you guys

glynnis:
FYI, spoke with Ali about this and he'll be taking a stab at tweaking the design. I don't think it'll take him long, but if this needs to get merged first with a separate story for redesign later, that's fine as long as the separate story gets captured in the right place :+1:

cc/ @halkon @kobowley

grunt-tasks/options/connect.js
```diff
@@ -40,7 +51,22 @@ module.exports = {
     keepalive: {
         options: {
             keepalive: true,
-            base: '<%= config.docs %>'
+            middleware: function (cnct) {
+                return [
+                    function (req, res, next) {
```
CITguy: Can you please refactor this function to DRY up the code?

src/rxFeedback/rxFeedback.js
CITguy: This PR has "private business logic" written all over it.  If it can be done in the services project (or some other private directive repo), I would rather place the logic there.  Even though I've been told that this is a temporary stop gap for requested functionality, it will be more and more difficult to remove logic pertaining to this request as time goes by.

Droogans: I agree. Being able to support redirects to arbitrary URLs makes more sense to me. It'd be up to the teams implementing this redirect to supply their own UserVoice URL to the directive.

This makes it easier to test too -- I don't have to fumble with redirects to non-Angular sites that are only accessible behind a firewall :cry:. I can test that it redirects you to AngularJS.org and :smile: instead.

halkon: @Droogans the tests are done and it already redirects to angularjs.org for test purposes.  @CITguy I'll comment here in a bit :grin: but I agree for the most part

halkon:
Closing for now as this is about to get a major re-haul

PR #1163, opened 2015-08-28
chore(docs): create Styles Page: Colors

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-38

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

local-styles-color


demo/assets/demo.less
```diff
@@ -5,8 +7,12 @@
     padding: 10px;
     margin-bottom: 2em;
 
-    // TODO: can we do something about this?
-    h1, h2, h3, h4, h5, h6 {
```
halkon: I am guessing the `TODO` was to perhaps do a `for (...` loop style declaration. 

Only thing I found online for it is:

https://github.com/seven-phases-max/less.curious/blob/master/articles/generic-for.md#multiple-loops-in-same-scope (A mixin someone made)

halkon:
LGTM

demo/assets/demo.less
CITguy: I remember adding this todo, but I cannot remember why. I recall it had something to do with the deprecation of the `.title` classes, but I've slept since then, so I just removed the comment.

halkon: Ah, that would make sense also  :grin: :+1: I was assuming that `TODO` was older than github

src/styles/vars.less
```diff
@@ -1,5 +1,15 @@
...
+ */
 @white: #fff;
+@green: #24b56a;
+@light-green: #c8e5bf;
+@blue: #4881c0;
+@light-blue: #86cafd;
+@disabled-gray: #dededd;
+@cancel-gray: #898989;
+@orange: #db7820;
+@red: #fc0f1d;
```
megalord: Perhaps it's best left for another PR, but we should replace hardcoded values with these variables elsewhere in the codebase.

megalord:
:+1: w/ comment

src/styles/vars.less
CITguy: Yes, likely a separate PR.  Those values that were hard coded within `vars.less` have been updated, though.

demo/demo.js
```diff
@@ -323,6 +328,15 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
             ]
         },
         {
+            title: 'Style',
```
bry4n: We are going to have multiple style pages. pluralize "Style"?

CITguy: I'm going with "Style" considering that is what is in the mockup.

ElementE:
Looks good, Design LGTM

PR #1167, opened 2015-09-02
refactor(rxNotify): remove ability to prevent stack notification dismissal

CITguy:
JIRA:

* [FRMW-32](https://jira.rax.io/browse/FRMW-32): Make all notifications dismiss-able
* [FRMW-113](https://jira.rax.io/browse/FRMW-113): Tech Debt: rxNotify - Move API documentation to API docs

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [ ] QE LGTM

### DIFF
--- a/src/rxNotify/README.md
+++ b/src/rxNotify/README.md
@@ -2,220 +2,114 @@
 
 Service (rxNotify) and Directives (rxNotification and rxNotifications) for displaying status messages on a page.
 
-## Using rxNotification as a State Message
+[API Documentation](ngdocs/index.html#/api/rxNotify.service:rxNotify)
 
-There may be situations where you will need to use the styling/markup of rxNotify's messaging queue in status messages of your own - for example, a modal window which asks if you want to delete an object, with the appropriate warning or error flags. If this is the case, we recommend using the `rx-notification` directive in your views (note difference of `rx-notifications`).
 
-> `<rx-notification type="warning">This is a warning message!</rx-notification>`
+## Displaying an rxNotification as a State Message
 
-For all notification types, please look below under Message options, under `type`.
+There may be situations where you will need to use the styling/markup of
+rxNotify's messaging queue in status messages of your own - for example,
+a modal window which asks if you want to delete an object, with the
+appropriate warning or error flags. If this is the case, we recommend
+using the `rx-notification` directive in your views.  Please note, this
+differs from `rx-notifications` (plural).
 
-Another situation that you might encounter is adding a notification to the default stack but needing to do it via the template.  With the `stack` parameter you're allowed do define a notification and have it get added via `rxNotify.add`:
+```html
+<rx-notification type="warning">
+  This is a warning message!
+</rx-notification>
+```
 
-> `<rx-notification type="error" stack="page">`
+The type attribute can be any type supported by `options.type` for the `add()`
+function in the [rxNotify service](ngdocs/index.html#/api/rxNotify.service:rxNotify).
 
->> `   This is an error message being added to the "page" stack with <strong>Custom</strong> html.`
-
-> `</rx-notification>`
-
-## Adding a New Message Queue via rxNotify
-
-To add a new message to a stack, inject 'rxNotify' into your function and run:
-
-> `rxNotify.add('My Message Text');`
-
-This will add a new message to the default stack ('page') with all default options set. To customize options, pass in an object as the second argument with you specific options set:
-
-> `rxNotify.add('My Message Text', {`
-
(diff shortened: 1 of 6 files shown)


CITguy:
## API Docs rxNotify Component
api-rxnotify

CITguy:
## API Docs rxNotify Service
api-rxnotify-rxnotify-service

CITguy:
## API Docs rxNotifications Service
api-rxnotify-rxpromisenotifications-service

CITguy:
## API Docs rxNotifcation Directive
api-rxnotify-rxnotification-directive

CITguy:
## API Docs rxNotifcations Directive
api-rxnotify-rxnotifications-directive

CITguy:
## Demo rxNotify
demo-rxnotify

glynnis:
@CITguy In the demo under the "Using rxNotification" section, the "Hello World!" examples should have an `x` since https://jira.rax.io/browse/FRMW-32 is about making all instances of rxNotification dismissible, correct? Genuinely asking since I'm uncertain I'm understanding the story correctly.

src/rxNotify/README.md
```diff
@@ -2,220 +2,114 @@
...
-> `<rx-notification type="warning">This is a warning message!</rx-notification>`
+## Displaying an rxNotification as a State Message
 
-For all notification types, please look below under Message options, under `type`.
+There may be situations where you will need to use the styling/markup of
+rxNotify's messaging queue in status messages of your own - for example,
+a modal window which asks if you want to delete an object, with the
+appropriate warning or error flags. If this is the case, we recommend
+using the `rx-notification` directive in your views.  Please note, this
+differs from `rx-notifications` (plural).
```
glynnis: Will rxNotifications within modals be dismissible? This wasn't discussed in the JIRA story (and I think that was an oversight). After speaking with @ElementE, we think rxNotification within modals should NOT be dismissible since there are many cases where rxNotification is used within a modal as the only content. Any instances of rxNotification outside of a modal (besides "loading"/programmatically dismissed rxNotifications) should always be dismissible (since they block content). 

If need be, making rxNotification non-dismissible within a modal can be addressed in a separate PR -- don't want to hold up any work here.

cc/ @lfielder

src/rxNotify/README.md
```diff
@@ -2,220 +2,114 @@
...
+More examples can be found in the [rxNotify API Docs](ngdocs/index.html#/api/rxNotify.service:rxNotify).
 
-------
 
 ## Dismissing a message programatically
 
-Most messages are dismissed either by the user, a route change or using the custom 'dismiss' property.
-
-If you need to dismiss a message programmaticaly, you can run **`rxNotify.dismiss(message)`**, where message is the message object to dismiss.
+Most messages are dismissed either by the user, a route change or using the
```
glynnis: I think we could add something here about UX/UI reasoning for this -- something like:

> All messages except those dismissed programmatically are dismissible by the user. Since notifications cover 
> page content as they stack, the user must be able to dismiss them.

If addressed as part of this PR, you might also add something about why rxNotifications within a modal are not dismissible, since they are sometimes the only content.

CITguy:
@glynnis, there are a few things to note with this component.

0. The `rxNotification` directive will display a dismiss button only if it has been defined with a `stack` attribute.  This is how you add notifications to a stack declaratively via markup. If the `stack` attribute is not provided, it will behave as a static notification and will not display a dismiss button.  
0. The `rxNotifications` directive will automatically add the dismiss button for all notifications tied to a stack.
0.  The `rxNotify` service (and consequently `rxPromiseNotifications`) will add notifications to the stack programmatically and will make use of built in logic in `rxNotifications` (providing dismiss buttons for each notification).

It's not clearly documented, but the "Hello World" examples are static notifications (no `stack` attribute), so they will not display a dismiss button.  

I interpreted the intent of the JIRA story as "**stack** Notifications should always be dismissable" because this is the typical nuisance that users have run into where notifications are blocking data on the page. This is how the PR is coded.

For cases where the notification is used in a modal as the only content (or primary content), developers can still use `rxNotification` without a `stack` to display a static, non-dismissable notification.  If existing code is using a custom stack in the modal view, then they will see dismiss buttons on their modal notifications.  To correct this, teams will need to update to use static notifications instead.

glynnis:
Awesome, thanks for the explanation @CITguy -- very helpful. This has a design LGTM :+1:

parlarjb:
LGTM

bry4n:
LGTM

PR #1172, opened 2015-09-03
feat(docs): FRMWK-50 - Add Collapsible Layout page

bry4n:
JIRA: https://jira.rax.io/browse/FRMW-50

* [x] Dev LGTM
* [x] Dev LGTM
* [x] Design LGTM

Screenshot:
collapsible elements encoreui

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -111,7 +111,7 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                 name: 'Modals'
             }])
         })
-        /* Layout (flexgrid) */
+        /* Layouts */
         .when('/layout/collapsible', {
             templateUrl: 'layout/collapsible.html',
             controller: genericRouteController([{
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -55,12 +55,6 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                 name: 'Layouts'
             }])
         })
-        .when('/styleguide/layouts/collapsible', {
-            templateUrl: 'styleguide/layouts/collapsible.html',
-            controller: genericRouteController([{
-                name: 'Collapsible
-            }])
-        })
         .when('/styleguide/layouts/1', {
             templateUrl: 'styleguide/layout-1.html',
             controller: genericRouteController([{
@@ -115,6 +109,13 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                 name: 'Style Guide', path: '#/styleguide/basics'
             }, {
                 name: 'Modals'
+            }])
+        })
+        /* Layout (flexgrid) */
+        .when('/layout/collapsible', {
+            templateUrl: 'layout/collapsible.html',
+            controller: genericRouteController([{
+                name: 'Collapsible'
             }])
         })
         /* Style Pages */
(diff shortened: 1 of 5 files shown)


demo/layout/collapsible.html
```diff
@@ -0,0 +1,21 @@
+<rx-page title="'Collapsible Elements'" class="styleguide">
+
+    <h3> Collaspible Box </h3>
+    <p>
+        The rxCollapse element can be used to display and hide content. it can be
```
parlarjb: Capital `i` on "it"

bry4n: good catch :+1:

demo/demo.js
```diff
@@ -110,6 +111,13 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                 name: 'Modals'
             }])
         })
+        /* Layout (flexgrid) */
```
CITguy: Please update to `/* Layouts */`

demo/layout/collapsible/see-more-see-less.html
```diff
@@ -0,0 +1,22 @@
...
+	        <dd>Lorem ipsum dolor sit amet</dd>
+	        <dt>ID:</dt>
+	        <dd>1aa2bfa9-de8d-42f7-9f6d-e6437855b36e</dd>
+	        <dt>Region:</dt>
+	        <dd>ORD</dd>
+	        <dt>Created:</dt>
+	        <dd>December 2, 2014 @ 14:28</dd>
+        </rx-collapse>
+    </dl>
+</div>
```
CITguy: Please update to use the new [`rxMetadata`](http://rackerlabs.github.io/encore-ui/#/components/rxMetadata) format.

bry4n: ah forget about new `rxMetadata` format. It was copy & paste from `rxCollapse` directive page.

lfielder:
Should the ```See more/See less``` section be using the new metadata styles?

CITguy:
@lfielder: Yes, I commented yesterday to update to using rxMetadata.

CITguy:
@bry4n, can you please update the screenshot with the rxMetadata update?

CITguy:
Dev LGTM

bry4n:
Screenshot is updated

CITguy:
Thank you!

glynnis:
@bry4n late to the game here, but there's some information on the component page for rxCollapse that I think also belongs here, namely these additional points for the "See More/See Less" section:

```
- This pattern was developed for areas displaying metadata that may be short on screen real estate, as a way to hide data on load that is not as important to the user in the context where they are presented.
- This pattern is not very responsive-friendly, since as browser width decreases, columns will wrap. As columns wrap, the "See More" rxCollapse elements get lost in the new context, which is bad for user experience.
- To avoid the problem described above, "See More" rxCollapse elements should only be used at the end of the final column present on the page, so that when the columns wrap via flexbox, "See More" is always last and doesn't get lost in between metadata key/value pairs.
```

Once those are added, this has a design LGTM :+2:

bry4n:
@glynnis those are added. screenshot is updated. :ok_hand:

glynnis:
Thanks @bry4n! :cake: 

Design: LGTM

halkon:
LGTM

PR #1175, opened 2015-09-04
feat(test): test commit

100stacks:


src/rxApp/rxApp.js
```diff
@@ -417,7 +417,7 @@ angular.module('encore.ui.rxApp', ['encore.ui.rxAppRoutes', 'encore.ui.rxEnviron
         replace: true,
         templateUrl: 'templates/rxAppSearch.html',
         scope: {
-            placeholder: '@?',
+            placeholder: '@?',
```
100stacks: Why the extra space?

PR #1185, opened 2015-09-09
docs(links): correct/remove dead links and documentation

### LINKED ISSUE #1183: Link on Basics page does not work
In the "No Data to Display" section there is a link that points to the wells documentation.  This like does not work.  

I would also point out that the wording is not as user friendly as it could be.  For example, it requires a user to click on the link in order to know what we are suppose to use. In this case its wells.  I as a user may already know how to use wells (give a div a class called ".wells"). If it simply stated such, while also providing a link if needed, it might save users a little time.

http://rackerlabs.github.io/encore-ui/#/styleguide/basics#no-data

CITguy:
Fixes #1183 

* remove old documentation section for collapsible element
* correct old "wells" links to use new href

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -222,10 +222,6 @@ angular.module('demoApp', ['encore.ui', 'ngRoute'])
                             linkText: 'Lists'
                         },
                         {
-                            href: '#/styleguide/basics#collapsible',
-                            linkText: 'Collapsible Element'
-                        },
-                        {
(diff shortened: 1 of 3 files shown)


bry4n:
LGTM

parlarjb:
LGTM

PR #1187, opened 2015-09-09
test(rxLocalStorage): Add a page object for rxLocalStorage

tomreece:
* [x] Dev LGTM
* [x] Dev LGTM

Added a rxLocalStorage page object with basic functionality to be used by tests in other projects. This saves us from having to write driver.executeScript() all over the place.

### DIFF
--- a/src/rxLocalStorage/docs/rxLocalStorage.midway.js
+++ b/src/rxLocalStorage/docs/rxLocalStorage.midway.js
@@ -1,0 +1,81 @@
+var rxLocalStorage = encore.rxLocalStorage;
+
+describe('rxLocalStorage', function () {
+
+    before(function () {
+        demoPage.go('#/components/rxLocalStorage');
+    });
+
+    it('should set and return a string', function () {
+        var aString = 'hello world';
+
+        rxLocalStorage.setItem('aString', aString);
+        expect(rxLocalStorage.getItem('aString')).to.eventually.equal(aString);
+    });
+
+    it('should set and return an int', function () {
+        var anInt = 42;
+
+        rxLocalStorage.setItem('anInt', anInt);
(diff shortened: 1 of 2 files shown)


src/rxLocalStorage/docs/rxLocalStorage.midway.js
```diff
@@ -0,0 +1,77 @@
...
+    it('should remove something that has been set', function () {
+        rxLocalStorage.setItem('something', 42);
+        rxLocalStorage.removeItem('something');
+        expect(rxLocalStorage.exists('something')).to.eventually.be.false;
+    });
+
+    it('should return true if key exists', function () {
+        rxLocalStorage.setItem('cortana', 123);
+        expect(rxLocalStorage.exists('cortana')).to.eventually.be.true;
+    });
```
Droogans: Please test this if something exists `to.eventually.be.false`.

tomreece: I do that on line 69.

CITguy: Technically, line 69 is testing the behaviour of adding and removing a value.  Whereas, I believe @Droogans wants to test if a value never existed in the first place. Something like...

```javascript
it('should return false if key does not exist', function () {
    // 'siri' hasn't been set
    expect(rxLocalStorage.exists('siri')).to.eventually.be.false
});
```

Am I correct in this assumption?

tomreece: Gotcha. I added a test for that scenario. PR updated.

Droogans:
LGTM && @CITguy please merge.

CITguy:
LGTM

PR #1195, opened 2015-09-11
DO NOT MERGE: test(travis): Replace phantomjs-check with Selenium

Droogans:
Apparently [Travis ships with `google-chrome` on all VMs now](http://docs.travis-ci.com/user/build-environment-updates/2015-04-09/#Updates), but that hasn't stopped folks from filing more complaints in the open issue regarding chrome on Travis test runs.

Let's see what happens.


utils/jsErrorCheck.midway.js
```diff
@@ -0,0 +1,111 @@
...
+        'rxSpinner',
+        'rxStatus',
+        'rxStatusColumn',
+        'rxToggle',
+        'rxToggleSwitch',
+        'rxTokenInterceptor',
+        'rxUnauthorizedInterceptor',
+        'tabs',
+        'tooltips',
+        'typeahead'
```
CITguy: I assume that there's no way to automate the creation of this list?

Droogans: I wish. Grabbing all of the names returns a promise, which is given to mocha right away.

```js
element.all(by.repeater('comp in vm.components')).each(function (componentRow) {
    it('should be here but the repeater promise is totally skipped and mocha never sees this');
});
```

You have no option but to use hardcoded strings to get protractor to loop over things like this. That's why the first test is for making sure all the components are there. I think this whole PR needs a re-do next week since Chrome is not working on Travis at all.

Got to figure that out first (firefox cannot see javascript errors in Selenium, so Chrome must be used).

CITguy:
Added "DO NOT MERGE" because there are some security concerns and other items to be hashed out first.

CITguy:
@Droogans, I'm going to close this considering our chat about potential issues with codeship.

PR #1196, opened 2015-09-14
chore(grunt): support for splitting up monolithic js source files

CITguy:
Provides the ability to split up the monolithic component javascript source files. In addition to current file structure, this will support the following hierarchy:

```
src/
  <component>/
    docs/
    templates/ 
    README.md
    <component>.less
    <component>.module.js # defines angular module for component
    scripts/
      <item>.js
      <item>.exercise.js
      <item>.page.js
      <item>.spec.js
```

### LGTMS
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/grunt-tasks/build.js
+++ b/grunt-tasks/build.js
@@ -1,8 +1,25 @@
 module.exports = function (grunt) {
     grunt.registerTask('build', 'Create build files', function () {
-        grunt.task.run(['clean:build', 'modules', 'concat:dist', 'concat:distTpls', 'concat:tmpLess',
-            'concat:tmpLessResp', 'less:encore', 'less:encoreResp', 'less:styleguide', 'copy:demomarkdown',
-            'copy:demohtml', 'copy:demoassets', 'copy:demopolyfills', 'imagemin', 'concat:rxPageObjects',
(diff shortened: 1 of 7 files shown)


CITguy:
* `src/<component>/<component>.less` still applies to the component as a whole. (no change necessary)
* `src/<component>/scripts`
  * new directory that will contain individual source files (directives, services, and filters) along with their specs, page objects, and exercises
  * _I don't believe there should be any `*.exercise.js` files in the root of the component directory._ (cc: @Droogans)
* `<component>.module.js`
  * acts as a sort of "manifest" file that will define:
    * the angular module (e.g. `encore.ui.rxForm`) and its dependencies
    * any simple values/constants required for the component to function
  * The rename is required because many components will have a `scripts/<item>.js` that would have the same name as the `src/<component>/<component>.js`. So it would be difficult to differentiate if your editor does not provide the full path of the file.

parlarjb:
LGTM, thanks for doing this!

CITguy:
@Droogans, do you see any issues with this directory structure?

Droogans:
LGTM @CITguy

bry4n:
LGTM

PR #1199, opened 2015-09-14
fix(bugs): manual save + rxButton

CITguy:
* update manual save example to use better markup based on current CSS (per conversation with @glynnis)
  * old markup using deprecated `title` class on inline `<span>` 
  * browser doesn't apply margins to `inline` elements
  * now using `<h3>`
* replace rxButton font-sizes with static, `px` values (per [JIRA 208](https://jira.rax.io/browse/FRMW-208))

### LGTMS
- [ ] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/demo/styleguide/forms/manual-saving.html
+++ b/demo/styleguide/forms/manual-saving.html
@@ -3,18 +3,18 @@
         <rx-notification ng-if="form.$dirty" type="info" stack="page">
             Changes made are not automatically saved. Click "Save Now" to save changes.
         </rx-notification>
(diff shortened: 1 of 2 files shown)


CITguy:
@megalord, FYI: this may affect you and the apps you support (based on the markup change to the manual form save demo).

megalord:
Oh yes, it definitely will.  Thank you for the heads up!

glynnis:
LGTM

megalord:
:+1:

PR #1229, opened 2015-09-23
fix(rxFloatingHeader): Pass local protractor tests.

Droogans:
This isn't an issue on Travis, but when running tests locally, there
is an issue where the floating header will be off by one pixel on
higher resolution displays. This fix adds a tolerance around the
location of the floating header during protractor tests to be
+/- 2 pixels.


src/rxFloatingHeader/docs/rxFloatingHeader.midway.js
```diff
@@ -56,22 +56,24 @@ describe('rxFloatingHeader', function () {
...
-            rxFloatingHeader.scrollToElement(middleRow);
-            expect(rxFloatingHeader.compareYLocations(filterHeader, middleRowY)).to.eventually.be.true;
+            encore.rxFloatingHeader.scrollToElement(middleRow);
+            expect(encore.rxFloatingHeader.compareYLocations(filterHeader, middleRowY)).to.eventually.be.true;
         });
 
         it('should have the right distance between the float header and the middle of the table', function () {
-            expect(rxFloatingHeader.compareYLocations(titlesHeader, middleRowY + filterHeight)).to.eventually.be.true;
+            var offset = middleRowY + filterHeight;
+            expect(encore.rxFloatingHeader.compareYLocations(titlesHeader, offset)).to.eventually.be.true;
```
CITguy: Should this test be checking within the lower/upper as well?

CITguy: @Droogans: I ask because I'm still getting an error for this test.

Droogans: Wasn't failing on my machine but I 100% believe you when you say it does on yours.

All of these tests will be updated to use tolerances.

CITguy:
@Droogans, I'm closing this PR because I've got an updated one at #1243

PR #1266, opened 2015-10-02
feat(docs): FRMW-198 - Migrate rxStatus docs

bry4n:
JIRA: https://jira.rax.io/browse/FRMW-198

* [x] Dev LGTM
* [x] Dev LGTM


src/rxStatus/rxStatus.js
```diff
@@ -4,7 +4,165 @@
  * @description
  * # rxStatus Component
  *
- * [TBD]
+ * This service is provided as a compliment to the `rxNotify`.  It abstracts out
+ * some of the raw functionality provided by `rxNotify` to make the addition and
+ * removal of single messages easier.
+ *
+ * ## Preparation
```
ssalinas24: @bry4n I confirmed with @CITguy, and I believe the content from here down is supposed to be in the service section. Down around line 198.

CITguy: Correct. Documentation pertains to the service, not the component.

bry4n: Yeah that makes sense. updated.

100stacks:
LGTM

CITguy:
LGTM

Issue #1270, opened 2015-10-02
rxTable page objects


jj4th:
There is a need for a set of general page objects for accessing tables/table data.  I have created rxTable to fill this need.

I expect there will be a LOT of discussion on this, and it's fairly documented, so I will let the PR speak for itself for now.

jj4th:
Oh yeah, I have not yet constructed demos or midways for this.  They will be forthcoming.

glynnis:
@jj4th Is this still relevant? Looks like the PR was closed, so I'm closing this issue, but re-open if this is still needed.

PR #1271, opened 2015-10-02
feat(rxTable): Page objects for tables

jj4th:
#1270

### DIFF
--- a/src/rxTable/rxTable.page.js
+++ b/src/rxTable/rxTable.page.js
@@ -1,0 +1,264 @@
+/*jshint node:true*/
+var _ = require('lodash');
+var Page = require('astrolabe').Page;
+
+/**
+ * @namespace
+ */
+var rxTable = {
+    /**
+     * @returns {rxBatchActions} - The rxBatchActions page object for the table (see encore.rxBatchActions)
+     */
+    batchActions: {
+        get: function () {
+            if (!this.batchActionsObject) {
+                this.batchActionsObject = exports.rxBatchActionMenu.initialize(this.rootElement.$('rx-batch-actions'));
+            }
+            return this.batchActionsObject;
+        }
+    },
+
+    /**
+     * @returns {rxBulkSelect} - The rxBulkSelect page object for the table (see encore.rxBulkSelect)
+     */
+    bulkSelect: {
+        get: function () {
+            if (!this.bulkSelectObject) {
+                this.bulkSelectObject = exports.rxBulkSelect.initialize(this.rootElement, this.rowFromElement);
+            }
+            return this.bulkSelectObject;
+        }
+    },
+
+    /**
+     * @returns {WebElement[]} - Returns the elements for all the data columns on the table.
+     */
+    columnElements: {
+        get: function () {
+            var selectors = [];
+            _.each(this.config.columns, function (column, index) {
+                selectors.push('thead > tr > th.column-title:nth-of-type(' + (index + 1) + ')');
+            });
+            return this.rootElement.$$(selectors.join(', '));
+        }
+    },
+
+    /**
+     * @returns {WebElement[]} - Returns the elements for all the column headers on the table.
+     */
+    headerElements: {
+        get: function () {
+            var selectors = [];
+            _.each(this.config.columns, function (column, index) {
+                var selector = 'thead > tr > th.column-title:nth-of-type(' + (index + 1) + ')';
+                if (column.sortable) {
+                    selector += ' span.ng-scope:first-of-type';
+                }
+                selectors.push(selector);
+            });
+            return this.rootElement.$$(selectors.join(', '));
+        }
+    },
+
+    /**
+     * @returns {WebElement[]} - Returns the elements for all the data rows on the table.
+     */
+    rowElements: {
+        get: function () {
+            return this.rootElement.$$('tbody > tr');
+        }
+    },
+
+    /**
+     * @returns {WebElement[]} - Returns the elements for all the footer rows on the table.
+     */
+    footerRowElements: {
+        get: function () {
+            return this.rootElement.$$('tfoot > tr');
+        }
+    },
+
+    /**
+     * @returns {WebElement} - The actual column headers present on the table.
+     */
+    searchElement: {
+        get: function () {
+            return this.rootElement.$('rx-search-box');
+        }
+    },
+
+    /**
+     * @function
+     * @param {Integer} columnIndex - The zero-index of a column in the table.
+     * @returns {rxTableColumn} - The rxTableColumn page object for the column (see rxTableColumn).
+     */
+    column: {
+        value: function (columnIndex) {
+            return exports.rxTableColumn.initialize(this, columnIndex);
+        }
+    },
+
+    /**
+     * @returns {Text[]} - The actual column headers present on the table.
+     */
+    headers: {
+        get: function () {
+            return this.headerElements.getText();
+        }
+    },
+
+    /**
+     * @function
+     * @param {Integer} rowIndex - The zero-index of a row in the table.
+     * @returns {rxTableRow} - The rxTableRow page object for the row (see rxTableRow).
+     */
+    row: {
+        value: function (rowIndex) {
+            return exports.rxTableRow.initialize(this, rowIndex);
+        }
+    },
+
+    /**
+     * @function
+     * @returns {rxSearchBox} - The rxSearchBox page object associated with the table.
+     */
+    searchBox: {
+        get: function () {
+            if (!this.rxSearchBoxObject) {
+                this.rxSearchBoxObject = exports.rxSearchBox.initialize(this.searchElement);
+            }
+            return this.rxSearchBoxObject;
+        }
+    },
+
+    /**
+     * @function
+     * @param {String} filterText - The text which will be applied to the searchBox element.
+     */
+    filter: {
+        value: function (filterText) {
+            this.searchBox.term = filterText;
+        }
+    },
+
+    /**
+     * @function
+     * @returns {Boolean} - Whether or not the table is present in the page.
+     */
+    isPresent: {
+        value: function () {
+            return this.rootElement.isPresent();
+        }
+    },
+
+    /**
+     * @function
+     * @returns {rxPagination} - The pagination page object for the table
+     */
+    pagination: {
+        get: function () {
+            return exports.rxPaginate.initialize(this.rootElement.$('.rx-paginate'));
+        }
+    }
+};
+
+/**
+ * @function
+ * @param {WebElement} rxTableElement - WebElement to be transformed into an rxTable page object.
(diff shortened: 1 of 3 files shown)


src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+function main (config) {
+    var elem = $$('rx-page table').first();
+    return initialize(elem, config);
+}
+
+/**
+ * @exports rxTable
+ */
+module.exports = {
+    columnTypes: exports.rxTableColumn.types,
```
jj4th: Travis CI doesn't like this.

jj4th:
Yeah that visual diff makes no sense.

CITguy:
@jj4th, are the visual regression screenshots what is holding this up?

CITguy:
@jj4th, @Droogans: how are we feeling about this PR?

Droogans:
@CITguy I'm interested in seeing if this is being backported from cloud.

It'd be nice to see it in action over there before pulling it into a component directory `src/rxTable` that has no other resources besides page objects + tests.

src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+    return initialize(elem, config);
+}
+
+/**
+ * @exports rxTable
+ */
+exports.rxTable = Page.create({
+    columnTypes: { get: function () { return exports.rxTableColumn.types; }},
+    initialize: { get: function () { return initialize; }},
+    main: { get: function () { return main; }}
```
Droogans: `main` should be dropped.

src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * @function
+     * @returns {rxSearchBox} - The rxSearchBox page object associated with the table.
+     */
+    searchBox: {
+        get: function () {
+            if (!this.rxSearchBoxObject) {
+                this.rxSearchBoxObject = exports.rxSearchBox.initialize(this.searchElement);
+            }
+            return this.rxSearchBoxObject;
```
Droogans: Why wrap this getter in a conditional? It looks like a memoization pattern.

src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+    },
+
+    /**
+     * @returns {WebElement} - The actual column headers present on the table.
+     */
+    searchElement: {
+        get: function () {
+            return this.rootElement.$('rx-search-box');
+        }
+    },
```
Droogans: Move all web elements to the top of the page object definition, and please prefix them accordingly.

src/rxTable/rxTableColumn.page.js
```diff
@@ -0,0 +1,324 @@
+var _ = require('lodash');
+var Page = require('astrolabe').Page;
+var moment = require('moment');
+// The date format comes from the Encore UI framework, see:
+// http://rackerlabs.github.io/encore-ui/#/styleguide/basics#date-formatting
+var dateFormat = 'MMM D, YYYY @ HH:mm (UTCZZ)';
```
Droogans: This is good but likely belongs someplace more easily consumable. I think rxAge or rxMisc would be appropriate.

src/rxTable/rxTableColumn.page.js
```diff
@@ -0,0 +1,324 @@
...
+    return elem.element(by.xpath('./span')).getAttribute('tooltip');
+}
+
+/**
+ * @function
+ * @returns {String} An empty string (cog columns have no data)
+ */
+function rxCogData() {
+    return '';
+}
```
Droogans: All of these are all hoisted function expressions. Please only use function definitions. Also, `rxCogData` is a function that returns a blank string. Should that really be a function?

src/rxTable/rxTableColumn.page.js
```diff
@@ -0,0 +1,324 @@
...
+    types: {
+        checkbox: { data: rxColumnCheckboxData },
+        date: { data: rxColumnDateData, sort: rxColumnDateSort },
+        relativeDate: { data: rxColumnRelativeDateData, sort: rxColumnRelativeDateSort },
+        string: { data: rxColumnStringData, sort: rxColumnStringSort },
+        ipAddress: { data: rxColumnIpData, sort: rxColumnIpSort },
+        nameId: { data: rxColumnNameIdData, sort: rxColumnStringSort },
+        status: { data: rxColumnStatusData },
+        number: { data: rxColumnStringData, sort: rxColumnNumberSort },
+        actionCog: { data: rxCogData }
```
Droogans: Ah, I see now why `rxCogData` is a function. Nevermind.

Droogans:
Midways are needed. Examples in the doc strings are great, but without midways it's hard to see the typical use cases in action.

src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+     * @returns {rxBulkSelect} - The rxBulkSelect page object for the table (see encore.rxBulkSelect)
+     */
+    bulkSelect: {
+        get: function () {
+            if (!this.bulkSelectObject) {
+                this.bulkSelectObject = exports.rxBulkSelect.initialize(this.rootElement, this.rowFromElement);
+            }
+            return this.bulkSelectObject;
+        }
+    },
```
Droogans: These two properties also ask if they've been called once before returning a "cached" version (though that's not even what is happening). These should return `exports.rxComponent.initialize(this.rootElement, ...)` directly.

src/rxTable/rxTable.page.js
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * @returns {WebElement[]} - Returns the elements for all the column headers on the table.
+     */
+    headerElements: {
+        get: function () {
+            var selectors = [];
+            _.each(this.config.columns, function (column, index) {
+                var selector = 'thead > tr > th.column-title:nth-of-type(' + (index + 1) + ')';
+                if (column.sortable) {
+                    selector += ' span.ng-scope:first-of-type';
```
Droogans: I think it's worth altering the html template of sortable columns that to rely on a css selector that is this weak.

src/rxTable/rxTableColumn.page.js
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/**
+ * @function
+ * @param {Object} elem A single cell's WebElement within the table column
+ * @returns {String} The Encore standard UTC date string for Encore standard relative date value of the cell
+ */
+function rxColumnRelativeDateData(elem) {
+    return elem.element(by.xpath('ancestor-or-self::td/time')).getAttribute('datetime').then(function (text) {
+        return moment(text, dateFormat).utc().format(dateFormat);
```
Droogans: `moment` has deprecated date parsing -- it expects that step to already have happened by `new Date()`, which is expects by default.

I'm pretty sure that you can pass in the `datetime` attribute to `new Date` without any trouble.

src/rxTable/rxTableColumn.page.js
```diff
@@ -0,0 +1,324 @@
...
+    });
+}
+
+/**
+ * @function
+ * @param {Object} elem A single cell's WebElement within the table column
+ * @returns {String} The Encore standard UTC date string for Encore standard relative date value of the cell
+ */
+function rxColumnRelativeDateData(elem) {
+    return elem.element(by.xpath('ancestor-or-self::td/time')).getAttribute('datetime').then(function (text) {
```
Droogans: This isn't just `text`. It's a `datetimeText`.

src/rxTable/rxTableColumn.page.js
jj4th: Is there any other reason than you not liking function declarations?  They are at the top of the file so there is no hoisting confusion (which, btw, function expressions are also hoisted).  This way allows them to be first-pass evaluated, which is more efficient... though granted efficiency is not a major concern here.

Droogans: I see these as global variables, and I try to never use them or allow them into apps unless there's a really good reason to.

src/rxTable/rxTable.page.js
jj4th: Yeah this is just memoization as you identified earlier... I'm not married to it.

src/rxTable/rxTable.page.js
jj4th: Yeah this is just memoization... I'm not married to it.

src/rxTable/rxTable.page.js
jj4th: Do we have a pluralization pattern for that prefix?  `eleSearchBox` is fine, but `eleFooterRows` kind of... meh... I guess it works.

src/rxTable/rxTable.page.js
jj4th: Yup, this was done before that change.

src/rxTable/rxTableColumn.page.js
jj4th: Yup, holdover from the cloud implementation.  I will move it.

src/rxTable/rxTable.page.js
Droogans: Ok it should be changed because if I recall correctly assigning one instance of rxBulkSelect to a static variable `this.bulkSelectObject` will throw a stale element reference error if you were to hide/show it and attempt to refer to `this.bulkSelectObject` again.

src/rxTable/rxTable.page.js
jj4th: I could get behind that.  Wasn't really appropriate when it was just in cloud.

src/rxTable/rxTable.page.js
Droogans: I use `tbl` for anything that is a list of elements, or a table of elements.

src/rxTable/rxTableColumn.page.js
jj4th: Actually we are insisting here on a specific format string, which is not deprecated.  This was intentional.

src/rxTable/rxTableColumn.page.js
jj4th: Just so I'm certain about what you're asking here, you want me to rename the function expression parameter?  I use text in every other function in this file.

jj4th:
@droogans responded to comments, changes not yet made.

src/rxTable/rxTableColumn.page.js
Droogans: No, it's fine. I just noticed it, nothing critical.

src/rxTable/rxTableColumn.page.js
Droogans: :+1: I see now, thanks.

jj4th:
I don't have time right now to work on this, so I'm going to table (hah) it for the time being.

Issue #978, opened 2015-05-21
Move utility functions to `rxMisc`


Droogans:
`rxForm` is a catch-all of sorts. There is a function that lives there in the page objects that converts the string output from Angular's currency filter and turns it into an integer for pennies that match the value of the currency string.

This is only in `rxForm` because `rxMisc` didn't exist at the time. Move it there.

### Encore 1.x 
- [x] Alias `rxForm.currencyToPennies` to point to `rxMisc.currencyToPennies`.
- [x] Alias `rxForm.slowClick` to point to `rxMisc.slowClick`.
- [x] Alias all `rxForm.*.generateAccessor` to point to `rx*.generateAccessor`
- [x] Mark `rxForm.currencyToPennies` as deprecated.
- [x] Mark `rxForm.slowClick` as deprecated.
- [x] Mark `rxFloatingHeader.*` as deprecated.

### Encore 2.0
- [ ] Delete `rxForm.currencyToPennies`.
- [ ] Delete `rxForm.slowClick`.
- [ ] Delete `rxForm.form`
- [ ] Delete `rxForm.*.generateAccessor`
- [ ] Delete `rxFloatingHeader.*`

CITguy:
@Droogans, how much of this is still valid?

Droogans:
Once the aliases go away we're set.

anur2213:
Associated Jira story: https://jira.rax.io/browse/FRMW-518

PR #1272, opened 2015-10-05
chore(rxFloatingHeader): add Y tolerance to midways

CITguy:
Expanded upon #1229, to apply Y value tolerance to all checks.

> This isn't an issue on Travis, but when running tests locally, there is an issue where the floating header will be off by one pixel on higher resolution displays. This fix adds a tolerance around the location of the floating header during protractor tests to be +/- 2 pixels.

- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -25,6 +25,7 @@ var config = {
     },
 
     onPrepare: function () {
+        browser.driver.manage().window().setSize(1366, 768); // laptop
         expect = require('chai').use(require('chai-as-promised')).expect;
         demoPage = require('./utils/demo.page.js');
         encore = require('./utils/rx-page-objects/index');
--- a/src/rxFloatingHeader/docs/rxFloatingHeader.midway.js
+++ b/src/rxFloatingHeader/docs/rxFloatingHeader.midway.js
@@ -1,11 +1,15 @@
-var rxFloatingHeader = encore.rxFloatingHeader;
-
 describe('rxFloatingHeader', function () {
     var table, tr, middleRow, middleRowY, initialY;
+    var yTolerance = 2; // +/- pixel tolerance for Y value comparison
+    var lower, upper;
+    var actual;
 
-    describe('One header row table', function () {
+    before(function () {
+        demoPage.go('#/components/rxFloatingHeader');
+    });
+
+    describe('Single-row header table', function () {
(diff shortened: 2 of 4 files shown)


bry4n:
LGTM

parlarjb:
What about building the tolerance stuff into the page object? i.e. something like

```
actual = encore.rxFloatingHeader.transformLocation(tr, 'y', { yTolerance: 2 });
expect(actual).to.eventually.be.within(actual.lower, actual.upper);
```

This way, if a product is using our page objects to test their own use of `rxFloatingHeader`, the API will make explicit that this sort of this can/should be done.

parlarjb:
(I just noticed that the `expect(actual)` wouldn't work anymore above, but you get the general idea)

parlarjb:
```
y = encore.rxFloatingHeader.transformLocation(tr, 'y', { yTolerance: 2 });
expect(y.actual).to.eventually.be.within(y.lower, y.upper);
```
maybe this instead

CITguy:
I see what you're saying, it would make sense to add this to the default exercise options if it needs changed/overridden for edge cases. I'll update this.

CITguy:
So, after looking at the page object, there are a few things to note:

1. All existing functions in `encore.rxFloatingHeader` will be moved to `encore.rxMisc` (with aliases to support backward compatibility), because they are utility functions that are not specific to floating headers. I will do this in the current PR.
2. The rxFloatingHeader page object needs refactored to act more like a page object (`encore.rxFloatingHeader.initialize()`) rather than a namespace for utility functions. (Captured in [FRMW-256](https://jira.rax.io/browse/FRMW-256))
3. Refactoring the location functions to use tolerances is not a simple undertaking, outside of the scope of this PR, and should be performed in a separate story (Captured in [FRMW-255](https://jira.rax.io/browse/FRMW-255))

parlarjb:
LGTM

Droogans:
Updated https://github.com/rackerlabs/encore-ui/issues/978 to reflect these deprecation notices.

@parlarjb I don't like the idea of passing in an optional argument that alters the return value. I think it makes sense to mark your tolerance values in the test if you need a tolerance (so far, it hasn't affected phone-ui, so there's that).

LGTM.

PR #1281, opened 2015-10-06
test(rxMisc): #1279 Round before parseInt in currencyToPennies

### LINKED ISSUE #1279: rxMisc.currencyToPennies can be off by a penny
Try calling `rxMisc.currencyToPennies("$150.14")`

You'll get 15013.

150.14 * 100 = 15013.999999999998
parseInt(15013.999999999998) = 15013

tomreece:
Attempt to fix #1279.

The rxMisc.page.js currencyToPennies function will sometimes `parseInt(15013.9999999999998)` when the expected result is 15014.

I removed the `it('should lose precision when converting a fraction of a penny to an int')` test case per @Droogans suggestion.

### DIFF
--- a/src/rxMisc/docs/rxMisc.midway.js
+++ b/src/rxMisc/docs/rxMisc.midway.js
@@ -116,10 +116,6 @@ describe('rxMisc', function () {
                 expect(fn('$0.01')).to.equal(1);
             });
 
-            it('should lose precision when converting a fraction of a penny to an int', function () {
-                expect(fn('$0.019')).to.equal(1);
-            });
-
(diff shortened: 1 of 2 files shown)


Droogans:
Needs a test case for `$150.14`.

tomreece:
@Droogans Added a test case for `$150.14`

Droogans:
Hey I noticed that the jsdoc examples need to be updated. The description may be out of date as well.

https://github.com/tomreece/encore-ui/blob/4439c30e0732636a16812c55633b16e41760401f/src/rxMisc/rxMisc.page.js#L32-L37

PR #1283, opened 2015-10-06
test(rxMisc): Remove a test and update currencyToPennies ngdoc

tomreece:
I forgot to update the ngdoc for currencyToPennies in my last commit. Oops!

Also removed an irrelevant test since we're now saying currencyToPennies should not be passed fractions of a penny.


CITguy:
LGTM

tomreece:
I should have had this in the first pull request. I got in a hurry at the end of day yesterday. Bad Tom.

PR #1290, opened 2015-10-07
fix(rxTags): midway revert and workaround

CITguy:
* revert template back to `ng-keydown` (b/c Chrome doesn't work with `ng-keypress`)
* use protractor chord (SHIFT + BACK_SPACE) to workaround weirdness
* without this correction, rxTags will NOT work on Chrome

### LGTMS

- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/rxTags/rxTags.page.js
+++ b/src/rxTags/rxTags.page.js
@@ -98,8 +98,19 @@ var rxTags = {
 
     sendBackspace: {
         value: function () {
-            this.txtNewTag.click();
-            this.txtNewTag.sendKeys(protractor.Key.BACK_SPACE);
+            /*
+             * Protractor isn't properly trapping the `BACK_SPACE = navigate back`
+             * functionality so we have to use SHIFT + BACK_SPACE as a workaround.
+             *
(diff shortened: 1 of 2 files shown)


parlarjb:
LGTM

src/rxTags/rxTags.page.js
```diff
@@ -98,8 +98,11 @@ var rxTags = {
 
     sendBackspace: {
         value: function () {
-            this.txtNewTag.click();
-            this.txtNewTag.sendKeys(protractor.Key.BACK_SPACE);
+            var chordBackspace = protractor.Key.chord(
+                protractor.Key.SHIFT,
+                protractor.Key.BACK_SPACE
+            );
```
parlarjb: Could you leave a comment about this?

halkon:
LGTM
angel_ok

Droogans:
Ah, yes. Changing front end code for tests that break user's experience. I've *never* done it before :suspect:.

Yet another backfire from the tail wagging the dog. LGTM.

PR #1295, opened 2015-10-07
fix(package): pin to protractor 2.4.0

CITguy:
* Node 0.10
* Protractor 2.4.0


CITguy:
Gonna try bumping travis to node 0.12 with 2.5.0 and see what happens.

CITguy:
Choosing to bump to Node 0.12 instead of pinning to protractor 2.4.0

Issue #1234, opened 2015-09-25
rxMisc.currencyToPennies should accept ElementFinder, ElementArrayFinder types


ccarteronline:
It should also accept arrays of text, too.

```js
encore.rxMisc.currencyToPennies(['$1,234', '$2,345.01']);
```

Droogans:
Speaking with @jj4th we decided to drop the element finder params since we'd return promises for some types of calls and immediate values for others.

This should only support arrays of strings that look like currencies.

tomreece:
@ccarteronline @Droogans What should `encore.rxMisc.currencyToPennies(['$1,234', '$2,345.01']);` return?

`[123400, 234501]`? Or should it sum them or something?

tomreece:
We decided to make it return an array of penny values. Sum it yourself! :P

CITguy:
Closed via associated PR.

PR #1298, opened 2015-10-08
test(rxMisc): #1234 Modify currencyToPennies to accept an array too

tomreece:
For #1234 

The rxMisc.page.currencyToPennies function can now accept an array of currencies. It still works with a single currency so it's backwards compatible.

### LGTMS

- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/rxMisc/docs/rxMisc.midway.js
+++ b/src/rxMisc/docs/rxMisc.midway.js
@@ -131,6 +131,11 @@ describe('rxMisc', function () {
             it('should not incur any binary rounding errors', function () {
                 expect(fn('$150.14')).to.equal(15014);
             });
+
+            it('should convert an array of currencies to an array of pennies', function () {
+                expect(fn(['$0.01', '$100 CAN', '($100 AUS)', '$1.10', '$150.14']))
+                    .to.eql([1, 10000, -10000, 110, 15014]);
+            });
         });
 
         describe('unless', function () {
--- a/src/rxMisc/rxMisc.page.js
+++ b/src/rxMisc/rxMisc.page.js
@@ -22,27 +22,38 @@ exports.rxMisc = {
 
     /**
      * @description
-     * Transform `currencyString` (USD) to an integer representing pennies. Built to reverse
-     * Angular's 'currency' filter. Do not pass in fractions of a penny.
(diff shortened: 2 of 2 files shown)


src/rxMisc/rxMisc.page.js
```diff
@@ -22,27 +22,38 @@ exports.rxMisc = {
...
      * @example
      * ```js
-     * encore.rxMisc.currencyToPennies('$0.01')      ==  1
-     * encore.rxMisc.currencyToPennies('$100 CAN')   ==  10000
+     * encore.rxMisc.currencyToPennies('$0.01') == 1
+     * encore.rxMisc.currencyToPennies('$100 CAN') == 10000
      * encore.rxMisc.currencyToPennies('($100 AUS)') == -10000
-     * encore.rxMisc.currencyToPennies('$1.10')      ==  110
+     * encore.rxMisc.currencyToPennies('$1.10') == 110
+     * encore.rxMisc.currencyToPennies(['$0.01', '($100 AUS)', '$150.14']) == [1, -10000, 15014]
```
Droogans: Thanks for getting this.

Keep in mind that line 28 says the param is type `string`, you can write that as `string|array<string>` to show at a glance support for the new input!

tomreece: Ah crap. Missed that. Will update.

tomreece: Updated.

CITguy:
LGTM

src/rxMisc/docs/rxMisc.midway.js
```diff
@@ -131,6 +131,11 @@ describe('rxMisc', function () {
             it('should not incur any binary rounding errors', function () {
                 expect(fn('$150.14')).to.equal(15014);
             });
+
+            it('should convert an array of currencies to an array of pennies', function () {
+                expect(fn(['$0.01', '$100 CAN', '($100 AUS)', '$1.10', '$150.14']))
```
bry4n: It would be cool to use different name for `fn` function because I had to look up to see where is `fn` function coming from.

Maybe we could update them to use `currencyFn` or something in future.

bry4n:
LGTM

PR #1305, opened 2015-10-13
fix(z_index): FRMW-66-Bug-fix-rxFloatingHeader

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-66 - BUG: z-index for rxSelectFilter is behind rxFloatingHeader

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM 

Added a `z-index` criteria for `rxFloatingHeader`

### DIFF
--- a/src/rxFloatingHeader/rxFloatingHeader.less
+++ b/src/rxFloatingHeader/rxFloatingHeader.less
@@ -28,3 +28,9 @@ th.filter-header {
 tr.rx-floating-header:not(.rx-table-filter-row) {
     z-index: 2;
 }
+
+// Added to accommodate dropdown boxes to display "on top" of rx-floating-header 
+// components 
+thead .rx-floating-header:first-child {
+    z-index: 3;
+}


100stacks:
screen shot 2015-10-13 at 12 10 07 pm

src/rxFloatingHeader/rxFloatingHeader.less
```diff
@@ -28,3 +28,9 @@ th.filter-header {
 tr.rx-floating-header:not(.rx-table-filter-row) {
```
ssalinas24: @glynnis I sat with james to trouble shoot this item. Line 28: targets both tr elements in the markup because they both have the .rx-floating-header class, and neither of them have the .rx-table-filter-row. So they were both getting the Z-index: 2 and ending up on the same layer. The additional code on line 34: only targets the first child of .rx-floating-header within the thead. This was done since both elements are inside of the thead. Does this seem like an appropriate fix, or would you suggest anything different?

CITguy: Seems like sound logic for me. :+1:

CITguy:
LGTM, nice work!

parlarjb:
LGTM

glynnis:
LGTM

PR #1309, opened 2015-10-14
fix(bug): FRMW-130-Bug-unbreakable-long-values

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-130 - Long, Unbreakable Values Break rxFloatingHeader

~~-[ ] Dev LGTM~~
~~- [ ] Dev LGTM~~
~~- [ ] Design LGTM~~

Get feedback on proper way to visually represent and handle unbreakable long lines.

### DIFF
--- a/src/rxApp/common.less
+++ b/src/rxApp/common.less
@@ -126,6 +126,9 @@ table {
             .chevron-mixin(-1);
         }
     }
+    td.force-word-break {
+        word-break: break-all;
+    }
     .fa-times.delete-x:hover {
         color: @warnRedHover;
         cursor: pointer;
--- a/src/rxPaginate/docs/rxPaginate.js
+++ b/src/rxPaginate/docs/rxPaginate.js
@@ -4,7 +4,8 @@ angular.module('demoApp')
                                         PageTracking, rxSortUtil, SelectFilter) {
     $scope.pager = PageTracking.createInstance({ itemsPerPage: 3 });
 
-    var os = ['Ubuntu 12.04', 'Red Hat Enterprise Linux 6.4', 'CentOS 6.4', 'Ubuntu 13.04'];
+    var os = ['Ubuntu 12.04', 'Red Hat Enterprise Linux 6.4', 'CentOS 6.4', 'Ubuntu 13.04',
+              'superlonglinevalue1superlonglinevalue2superlonglinevalue3superlonglinevalue4superlonglinevalue5superlonglinevalue6superlonglinevalue7superlonglinevalue8superlonglinevalue9'];
     var makeServers = function (serverCount) {
         var servers = [];
         for (var i = 1; i < serverCount + 1; i++) {


100stacks:
Normal values in `rxPaginate` tables:

screen shot 2015-10-14 at 11 17 11 am

100stacks:
Long values in `rxPaginate`:

100stacks:
Also, long values break `rxFloatingHeader` from the `rxPaginate` table:

screen shot 2015-10-14 at 11 20 13 am

bry4n:
That's scary to have really long OS name :)

The fix may be necessary, but I doubt the OS name will be really long

100stacks:
@glynnis @ssalinas24 can I get your feedback on how to visually represent long values.

ssalinas24:
@100stacks I can try to take a look at it with you.

100stacks:
@bry4n, okay I agree. :+1:   I can across this when looking at long data values in `rxPaginate` and wanted to capture this behavior with drop downs.

bry4n:
@100stacks just updated my comment because I realized you have concern about that issue in general, not for just demo.

CITguy:
Well, crap! Wasn't expecting the values to bleed into rxSelectFilter. Definitely need some design input on this one.

CITguy:
IMO, I'd:

* `word-break: break-all;` on the selectable item (by default) in rxMultiSelect
* `word-break: break-all;` on table cells

100stacks:
@CITguy okay.  I will hold off on this issue, and breakout:
> * `word-break: break-all;` on the selectable item (by default) in rxMultiSelect

https://jira.rax.io/browse/FRMW-292

glynnis:
I talked with @100stacks about this, and we're going to take a stab at applying `word-break: break-all;` to table cells, followed by rigorous experimenting/smoke testing to see whether it does anything we don't expect. 

I don't think it would have any negative effects, especially since these long values seem to be an edge case, but we'll want to review it with extra scrutiny since many products use these table styles. Might even be worth checking out a few projects and running it against local Framework with this change just to try out real-world examples.

100stacks:
@glynnis sounds good.  @CITguy we might want to move this to our next sprint, since it will require a broader test suite.

src/rxApp/common.less
```diff
@@ -68,6 +68,7 @@ table {
     padding: 0;
     margin: 0;
     border-collapse: collapse;
+    word-break: break-all;
```
100stacks: This was moved to backlog for further discussions.  Added this line to all `table` elements in `common.less` as a proposed modification.  
As a side note, I noticed that unbreakable long values would break `nav` items, and possibly other items in *EncoreUI*.

100stacks: Unbreakable long values in navigation:

screen shot 2015-10-16 at 11 46 47 am

CITguy: Since we have control over nav items via the `encore-ui-nav` repo, I don't foresee this as being a problem.  Unbreakable values are more troublesome when they exist in received API data (the nav currently doesn't populate via API data) which is most obvious in:

* page title/subtitle
* modal title/subtitle
* table data
* metadata values (already handled via `rx-meta.force-break` css class)
* notifications

/cc @glynnis is there anything I might be missing?

glynnis: @CITguy Can't think of anything else at the moment, even after clicking through staging to see if anything jumped out at me. Sounds good :+1: 

I'd agree re: the fact that we have control over what goes in the nav/not needing to worry about long lines there.

100stacks: Okay @glynnis @CITguy did you want me to remove the `DO NOT MERGE` label and push forward for review/merge?

CITguy: Consider the following.

## No Word-Break
Below is the table without word-break styles applied (for comparison).

table-no-break


## Blanket Word-Break
This strategy will affect natural word-break functionality already present in the browser, so you'll be affecting cells that shouldn't need the style applied.

### LBaaS Table (blanket style present, no floating headers)
table-word-break-normal
* natural word-break isn't interrupted
* forced word-break prevents data being pushed off screen


### LBaaS Table (blanket style present, floating headers present)
table-word-break-floating
* floating headers broken
* natural word-break interrupted
* forced word-break prevents data being pushed off screen

## TD-specific CSS Class
I would highly recommend the use of a TD-specific CSS class to apply to table cells (`td.force-break`) so that it can be applied without affecting the natural word-break of other cells in the table.

### LBaaS Table (td style present, no floating headers)
td-word-break-normal
* natural word-break isn't interrupted
* forced word-break prevents data being pushed off screen

### LBaaS Table (td style present, floating headers present)
td-word-break-floating
* floating headers still work
* natural word-break isn't interrupted
* forced word-break prevents data being pushed off screen

CITguy: @glynnis: Thoughts on above?

glynnis: @CITguy ah, good catch. This looks good.

Just to be sure I understand, in summary, we'd supply a class and ask projects to only apply `.force-break` on `td`s they know might contain long values?

CITguy: @glynnis: Correct, it would work the same way it's done with rxMetadata.

glynnis: @CITguy Cool, yeah, a separate class sounds good.

CITguy: @100stacks, please update styling to apply only to `td.force-word-break` instead of the table as a whole.

```less
td.force-word-break {
    word-break: break-all;
}
```

100stacks:
@CITguy @glynnis incorporated changes from your discussions. :+1:

CITguy:
@100stacks, please post screenshot for Design

100stacks:
@CITguy please notice the `force-word-break` on the `<td>` for the main column `server.os`

```html
<tbody>
            <tr ng-repeat="server in pagedServers.items">
                <td>
                    {{server.name}}
                </td>
                <td class="force-word-break">{{server.os}}</td>
            </tr>
</tbody>
```

100stacks:
FF35 - the `OS` `<td>` extends to right

screen shot 2015-10-26 at 5 57 44 pm

100stacks:
Chrome 45 - appears to display the content appropriately.

screen shot 2015-10-26 at 6 16 42 pm

CITguy:
These browser inconsistencies are making me wonder if changes in this PR would just make symptoms worse, and I would rather regroup to get a better long-term solution rather than patching browser functionality. @glynnis, @100stacks: would you guys be open to just closing this PR and regrouping at a later date?

100stacks:
@CITguy @glynnis makes sense, as you clearly detailed the potential unintended consequences of this current approach.  I agree we should close, and revisit later...we can include a reference to this PR thread for background on the new approach.

PR #1332, opened 2015-10-23
chore(rxfeedback): Updated text, and removed well

halkon:
Updated text and removed modal

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

screen shot 2015-10-23 at 4 44 07 pm

### DIFF
--- a/src/rxFeedback/templates/feedbackForm.html
+++ b/src/rxFeedback/templates/feedbackForm.html
@@ -19,16 +19,11 @@
               ng-init="fields.type = feedbackTypes[0]"
               required="true"></select>
         </rx-input>
-        <rx-help-text ng-if="state === 'redirect'">You'll be redirected to a new window.</rx-help-text>
+        <rx-help-text ng-if="state === 'redirect'">Click continue to be redirected to the 'Feedback Forum' page to submit your feedback.</rx-help-text>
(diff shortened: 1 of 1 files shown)


halkon:
@glynnis @lfielder was the one who gave me this mockup

julianbonilla:
LGTM

CITguy:
LGTM

ssalinas24:
I pulled pr/1332 down to my local and the following is what I see. I don't see the continue button anywhere. I selected Feature Request from drop down and typed in the box. 
As for expected results:
- Is it supposed to have a Continue button like in your initial image?
- Is the feedback/comment box supposed to be there?

chore_rxfeedback

halkon:
@ssalinas24 there's a caveat when downloading this and testing it, you have to install `encore-ui-svcs` include it in the `demo/index.html` and add the angular module dependency `encore.svcs.feedback` to `demo/demo.app` 

That allows the modification to show up.

The reasoning for this is so that we don't have business logic in here and it's "easier" to make modifications to the behavior of the feedback logic. 

It's not straight forward to test locally, sorry about that, however the screen shot was taken from me doing that locally.

ssalinas24:
LGTM

Issue #1338, opened 2015-10-27
rxSearchBox api guide dead links


dman777:
In [rxSearchBox api documentation](http://rackerlabs.github.io/encore-ui/ngdocs/index.html#/api/rxSearchBox.directive:rxSearchBox) the [demo links](https://developer.rackspace.com/) in 

>  Though it is described as a search box, you can also use it for filtering capabilities (as seen by the placeholder text in the "Customized" demo).
Styling
You can style the <rx-search-box> element via custom CSS classes the same way you would any HTML element. See the customized search box in the demo for an example.

point to https://developer.rackspace.com/

PR #1339, opened 2015-10-27
fix(links): Fix demo links

bry4n:
For Issue https://github.com/rackerlabs/encore-ui/issues/1338

Fix all demo links (`/#/components/...` -> `/encore-ui/#/components/...`)

* [x] Dev LGTM
* [x] Dev LGTM

### DIFF
--- a/grunt-tasks/options/connect.js
+++ b/grunt-tasks/options/connect.js
@@ -16,6 +16,14 @@ module.exports = {
             rewrite: {
                 '/api/identity': '/v2.0/'
             }
+        },
+        {
+            context: '/encore-ui',
+            host: 'localhost',
+            port: 9001,
+            rewrite: {
+                '/encore-ui': '/'
+            }
         }
     ],
     dist: {
--- a/src/rxButton/rxButton.js
+++ b/src/rxButton/rxButton.js
@@ -8,9 +8,9 @@
  *  for `<button>` elements in scenarios where the button has multiple states.
  *
  * ## Styling
- * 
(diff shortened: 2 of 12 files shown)


100stacks:
@bry4n when we add `/encore-ui/` it won't point to the demo pages on local builds.

So...
`http://localhost:9001/#/components/rxModalAction` 
works, and...
`http://localhost:9001/encore-ui/#/components/rxModalAction` 
will not.

It appears the deployed environment requires `/encore-ui/`, while local environment doesn't.  Any thoughts on how to handle both?

CITguy:
Sadly, these links will not work locally without the following proxy added to `grunt-tasks/options/connect.js`

```javascript
{
    context: '/encore-ui',
    host: 'localhost',
    port: 9001,
    rewrite: {
        '/encore-ui', '/'
    }
}
```

@bry4n, please add the above proxy to the `proxies` object in `grunt-tasks/options/connect.js`.

CITguy:
Additionally, please update the following links as well.

* https://github.com/rackerlabs/encore-ui/blob/master/src/typeahead/typeahead.js#L21
* https://github.com/rackerlabs/encore-ui/blob/master/src/rxButton/rxButton.js#L13

bry4n:
@CITguy @100stacks added new proxy to connect config and updated two API docs.

100stacks:
@bry4n looks great! LGTM :+1:

CITguy:
LGTM

PR #1344, opened 2015-10-28
feat(test): FRMW-425 add rxToggleSwitch failed async midway

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-425

- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

---
Added midway tests for `rxToggleSwitch` failed asynchronous operation

### DIFF
--- a/src/rxToggleSwitch/docs/rxToggleSwitch.midway.js
+++ b/src/rxToggleSwitch/docs/rxToggleSwitch.midway.js
@@ -17,8 +17,13 @@ describe('rxToggleSwitch', function () {
         instance: encore.rxToggleSwitch.initialize($('.demo-post-hook'))
     }));
 
+    describe('failed asynchronous operation', encore.exercise.rxToggleSwitch({
+        instance: encore.rxToggleSwitch.initialize($('.demo-failed-async')),
+        enabledAtEnd: false
+    }));
+
     describe('disabled', encore.exercise.rxToggleSwitch({
-        cssSelector: '.demo-disabled',
+        instance: encore.rxToggleSwitch.initialize($('.demo-disabled')),
         disabled: true
     }));
 
--- a/src/rxToggleSwitch/rxToggleSwitch.exercise.js
+++ b/src/rxToggleSwitch/rxToggleSwitch.exercise.js
@@ -8,6 +8,10 @@ var rxToggleSwitch = require('./rxToggleSwitch.page').rxToggleSwitch;
    @param {rxToggleSwitch} [options.instance=] - Component to exercise.
    @param {string} [options.cssSelector=] - DEPRECATED: Fallback selector string to initialize widget with.
    @param {boolean} [options.disabled=false] - Determines if the switch can be toggled
+   @param {boolean} [options.enabledAtStart=null] Beginning state of toggle switch.  The value will be detected 
+   automatically if not given.
+   @param {boolean} [options.enabledAtEnd=null] End state of toggle switch.  The value will be detected automatically 
+   if not given.
    @example
    ```js
    describe('default exercises', encore.exercise.rxToggleSwitch({
@@ -21,12 +25,15 @@ exports.rxToggleSwitch = function (options) {
     }
 
     options = _.defaults(options, {
-        disabled: false
+        disabled: false,
+        enabledAtStart: null, // begins 'OFF'
+        enabledAtEnd: null // ends 'ON'
     });
 
     return function () {
         var component;
-        var initialState;
+        var enabledAtStart;
+        var enabledAtEnd;
 
         var getText = function (isEnabled) {
             return isEnabled ? 'ON' : 'OFF';
@@ -51,7 +58,11 @@ exports.rxToggleSwitch = function (options) {
             }
 
             component.isEnabled().then(function (isEnabled) {
-                initialState = isEnabled;
+                // use option if available, otherwise use detected state
+                enabledAtStart = _.isNull(options.enabledAtStart) ? isEnabled : options.enabledAtStart;
+
+                // use option if available, otherwise use inverse of enabledAtStart
+                enabledAtEnd = _.isNull(options.enabledAtEnd) ? !enabledAtStart : options.enabledAtEnd;
             });
         });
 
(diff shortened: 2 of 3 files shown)


bry4n:
LGTM

src/rxToggleSwitch/rxToggleSwitch.exercise.js
```diff
@@ -69,12 +70,18 @@ exports.rxToggleSwitch = function (options) {
...
+                    expect(component.isEnabled()).to.eventually.equal(initialState);
+                    expect(component.text).to.eventually.equal(getText(initialState));
+                });
+            } else {
+                it('changes to ' + getText(!initialState) + ' when clicked', function () {
+                    toggle();
+                    expect(component.isEnabled()).to.eventually.equal(!initialState);
+                    expect(component.text).to.eventually.equal(getText(!initialState));
+                });
+            }
```
CITguy: These tests should not be in the exercises. Not all rxToggleSwitch elements are going to have async functionality.

Please move to `src/rxToggleSwitch/docs/rxToggleSwitch.midway.js`

src/rxToggleSwitch/rxToggleSwitch.spec.js
```diff
@@ -116,4 +122,14 @@ describe('rxToggleSwitch', function () {
...
+
+    it('should attempt then fail to activate asychronous operation', function () {
+        directiveScope = failedAsyncEL.scope();
+        expect(scope.model).to.be.true;
+        directiveScope.update();
+        timeout.flush(1);
+        scope.$apply();
+        expect(scope.model).to.be.false;
+        sinon.assert.calledOnce(scope.countMe);
+    });
```
CITguy: This is not testing any failed async functionality. If it's function were to fail, the model would still be `true`. Please update.

src/rxToggleSwitch/rxToggleSwitch.exercise.js
```diff
@@ -21,7 +21,8 @@ exports.rxToggleSwitch = function (options) {
     }
 
     options = _.defaults(options, {
-        disabled: false
```
100stacks: @CITguy for https://github.com/rackerlabs/encore-ui/pull/1344/files#r43451323
Should/can  `options` account for exercise exceptions?  It appears that this is happening when it checks for the `options.disabled` state.

CITguy: A `disabled` state is a *visual* state of the toggle switch, which is one of the things that midways/exercises test.  How the toggle switch arrives at that state is irrelevant to midways/exercises. All we care about is that it arrives at the expected state.

src/rxToggleSwitch/rxToggleSwitch.exercise.js
```diff
@@ -69,12 +70,18 @@ exports.rxToggleSwitch = function (options) {
             it('is in the ' + getText(initialState) + ' state', function () {
                 expect(component.text).to.eventually.equal(getText(initialState));
             });
-
-            it('changes to ' + getText(!initialState) + ' when clicked', function () {
```
100stacks: @CITguy when I revert this file to the original test suite, this test as written fails when testing for failed asynchronous action on these lines:
> expect(component.isEnabled()).to.eventually.equal(!initialState);
> expect(component.text).to.eventually.equal(getText(!initialState));

since this test is expecting the opposite/toggled state with `!initialState`.  

Any recommendations on how to handle? Thanks.

100stacks: Please see initial `rxToggleSwitch` on `master`:

screen shot 2015-10-30 at 5 38 07 pm

100stacks:
`rxToggleSwitch` when I include a call to the failed asynchronous demo:

screen shot 2015-10-30 at 5 39 35 pm

src/rxToggleSwitch/rxToggleSwitch.spec.js
100stacks: @CITguy comment implemented :+1:

100stacks:
@CITguy changes updated :+1:

src/rxToggleSwitch/rxToggleSwitch.exercise.js
```diff
@@ -8,6 +8,8 @@ var rxToggleSwitch = require('./rxToggleSwitch.page').rxToggleSwitch;
    @param {rxToggleSwitch} [options.instance=] - Component to exercise.
    @param {string} [options.cssSelector=] - DEPRECATED: Fallback selector string to initialize widget with.
    @param {boolean} [options.disabled=false] - Determines if the switch can be toggled
+   @param {boolean} [options.beginState=null] Beginning state of toggle switch
+   @param {boolean} [options.endState=null] End state of toggle switch
```
CITguy: Just thought of this. Add some verbiage here that the value will be automatically detected if not given.

CITguy:
@Droogans or @jj4th: can either of you please check this for QE approval?

100stacks:
@CITguy okay comments implemented :+1:

CITguy:
LGTM

src/rxToggleSwitch/docs/rxToggleSwitch.midway.js
```diff
@@ -17,8 +17,13 @@ describe('rxToggleSwitch', function () {
         instance: encore.rxToggleSwitch.initialize($('.demo-post-hook'))
     }));
 
+    describe('failed asynchronous operation', encore.exercise.rxToggleSwitch({
+        instance: encore.rxToggleSwitch.initialize($('.demo-failed-async')),
+        endState: false
+    }));
```
Droogans: `endState: false` is not exactly memorable.

What are you trying to convey with this option? The end state....is enabled at the end?

Droogans: `enabledAtStart: true` is all that is really needed here.

The `endState` business is all computed anyhow.

Just make sure you always leave it at the same position that you received it in. Exercises that have side effects are not good.

CITguy: The original exercise functionality (if not disabled) is to detect the initial state of the toggle switch and test that it transitions to the opposite state and back again (On -> Off -> On, Off -> On -> Off).

However, in the case of toggle switches that may not change state based on the functionality of its `post-hook` functionality, the initial state and its post-hook state may be the same value (Off -> Off, On -> On). An example being when a promise is rejected within the `post-hook` function.

As such, the exercises were updated to allow the default functionality, but allow the developer to override the initial or end state of the toggle switch if necessary.

Droogans: So you're expecting testers to predict when a toggle switch will fail, and pass that expected failure into the exercise as an option?

CITguy: They don't need to predict it. The exercise will still automatically check state, but in the case that you need to override either state, it's available.

Droogans: @100stacks @CITguy what do you think about adding a verb to the option, then? `enabledAtStart: true`, `enabledAtEnd: true` makes it obvious that there's a manual over ride happening of the typical behavior, or am I still not understanding the intent of this option?

CITguy: Sure, adding a verb makes sense to remove ambiguity.

100stacks:
@Droogans @CITguy okay I updated:
- `beginState` -> `enabledAtStart`
- `endState` -> `enabledAtEnd`

Please let me know if have other recommendations. :+1:

Droogans:
@jj4th please merge if you're OK with this.

jj4th:
@Droogans I looked it over and didn't see anything overwhelmingly alarming here.

PR #1357, opened 2015-11-05
chore(travis): testing if we can install FF41 on travis

CITguy:
DO NOT MERGE

Seeing if we can install Firefox 41 on travis for testing purposes.


CITguy:
Test successful. Moving on.

PR #1358, opened 2015-11-05
refactor(various): refactor for future compatibility

CITguy:
Backporting refactor work added in `ng1.3` branch. Should still be ng1.2 compatible.

* readability updated for `watch` config
* corrected rxBreadcrumbs demo/midway
* refactor rxOptionTable template to have `ng-switch-when` on wrapper elements
* refactor rxOptionTable to be compatible with ng1.3
* refactor rxBulkSelect to be compatible with ng1.3
* refactor rxSearchBox to be compatbile with ng1.3

### LGTMS
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/grunt-tasks/options/watch.js
+++ b/grunt-tasks/options/watch.js
@@ -3,36 +3,78 @@ module.exports = {
         livereload: 1337
     },
     scripts: {
-        files: ['src/**/*.js', '!src/*/*.spec.js', '!src/**/*.midway.js',
-                '!src/**/*.page.js', '!src/**/*.exercise.js'],
-        tasks: ['build', 'karma:watch:run']
+        files: [
+            'src/**/*.js',
+            '!src/**/*.spec.js',
+            '!src/**/*.midway.js',
+            '!src/**/*.page.js',
+            '!src/**/*.exercise.js'
+        ],
+        tasks: [
+            'build',
+            'karma:watch:run'
+        ]
     },
     specs: {
-        files: ['src/**/*.spec.js'],
-        tasks: ['karma:watch:run', 'copy:coverage'],
+        files: [
+            'src/**/*.spec.js'
+        ],
+        tasks: [
+            'karma:watch:run',
+            'copy:coverage'
+        ],
         options: {
             livereload: false
         }
     },
     componentHtml: {
-        files: ['src/**/templates/*.html'],
-        tasks: ['html2js', 'build', 'karma:watch:run']
+        files: [
+            'src/**/templates/*.html'
+        ],
+        tasks: [
+            'html2js',
+            'build',
+            'karma:watch:run'
+        ]
     },
     componentLess: {
-        files: ['src/**/*.less'],
-        tasks: ['concat:tmpLess', 'less']
+        files: [
+            'src/**/*.less'
+        ],
+        tasks: [
+            'concat:tmpLess',
+            'less'
+        ]
     },
     componentImages: {
-        files: ['src/**/images/*'],
-        tasks: ['imagemin']
+        files: [
+            'src/**/images/*'
+        ],
+        tasks: [
+            'imagemin'
+        ]
     },
     demoSite: {
-        files: ['src/**/docs/*.html', 'src/**/*.md', 'demo/**/*', '!demo/bower_components/**/*'],
-        tasks: ['build']
+        files: [
+            'src/**/docs/*.html',
+            'src/**/*.md',
+            'demo/**/*',
+            '!demo/bower_components/**/*'
+        ],
+        tasks: [
+            'build'
+        ]
     },
     rxPageObjects: {
-        files: ['src/**/*.page.js', 'src/**/*.exercise.js'],
-        tasks: ['concat:rxPageObjects', 'concat:rxPageObjectsExercises',
-                'jsdoc2md:rxPageObjects', 'shell:rxPageObjectsDemoDocs']
+        files: [
+            'src/**/*.page.js',
+            'src/**/*.exercise.js'
+        ],
+        tasks: [
+            'concat:rxPageObjects',
+            'concat:rxPageObjectsExercises',
+            'jsdoc2md:rxPageObjects',
+            'shell:rxPageObjectsDemoDocs'
+        ]
     }
 };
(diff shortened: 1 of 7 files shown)


src/rxBreadcrumbs/docs/rxBreadcrumbs.js
```diff
@@ -2,7 +2,7 @@
 angular.module('demoApp')
 .controller('rxBreadcrumbsCtrl', function ($scope, rxBreadcrumbsSvc) {
     rxBreadcrumbsSvc.set([{
-        path: '/',
+        path: '/#/components',
```
bry4n: :+1: I recently checked the live demo.  the `/` path took me to https://developer.rackspace.com/

grunt-tasks/options/watch.js
```diff
@@ -3,36 +3,78 @@ module.exports = {
         livereload: 1337
     },
     scripts: {
-        files: ['src/**/*.js', '!src/*/*.spec.js', '!src/**/*.midway.js',
-                '!src/**/*.page.js', '!src/**/*.exercise.js'],
-        tasks: ['build', 'karma:watch:run']
+        files: [
+            'src/**/*.js',
+            '!src/**/*.spec.js',
```
CITguy: This entry was also updated to have two asterisks for first wildcard.

`!src/*/*.spec.js` -> `!src/**/*.spec.js`

src/rxBreadcrumbs/docs/rxBreadcrumbs.midway.js
```diff
@@ -110,17 +110,19 @@ describe('rxBreadcrumbs', function () {
...
 
             middle.visit();
-            expect(browser.getCurrentUrl()).to.eventually.equal(homeHref);
+            expect(browser.getCurrentUrl()).to.eventually.equal(componentsHref);
         });
-        // Note that after this test, we are now at the /#/overview page
 
+        after(function () {
+            demoPage.go('#/components/rxBreadcrumbs');
+        });
```
CITguy: These changes will be necessary with ng1.3, but are ng1.2 compatible.

src/rxOptionTable/templates/rxOptionTable.html
```diff
@@ -20,9 +20,8 @@
       <td class="option-table-input">
         <div class="fillWrapper">
           <label ng-switch="type">
-            <div class="alignWrapper">
+            <div class="alignWrapper" ng-switch-when="radio">
               <input rx-radio
-                 ng-switch-when="radio"
```
CITguy: ng1.3 will get a little confused when you have the `ng-switch-when` on the `rx-radio` element. Also...
screen shot 2015-11-05 at 3 01 12 pm

src/rxBulkSelect/scripts/rxBatchActions.js
```diff
@@ -61,7 +61,7 @@ angular.module('encore.ui.rxBulkSelect')
             };
             rxBulkSelectCtrl.registerForNumSelected(numSelectedChange);
 
-            if (!_.isUndefined(rxFloatingHeaderCtrl)) {
```
CITguy: This will fail in ng1.3 for some reason. Not entirely sure why. But it's backward compatible with ng1.2 anyway.

src/rxBreadcrumbs/docs/rxBreadcrumbs.js
CITguy: yup :cry: :panda_face:

src/rxBulkSelect/scripts/rxBatchActions.js
bry4n: have you debugged `rxFloatingHeaderCtrl`? 

https://github.com/rackerlabs/encore-ui/blob/backport-ng13-updates/src/rxBulkSelect/scripts/rxBatchActions.js#L44

I wonder if it sets different value in ng1.3

CITguy: ### Angular 1.2 vs 1.3
If an optional controller is not present, Angular 1.2 will return `undefined`, but Angular 1.3 will return `null`.  The old way of using `!_.isUndefined()` worked great for 1.2, but will return a false positive for 1.3.

To correct, we would have to check that the value is an object, but `typeof null` is `"object"` so we can't use a vanilla `typeof` check.

#### comparing values of existing controllers
* ng1.2 returns constructor functions
* ng1.3 returns native objects

screen shot 2015-11-05 at 4 58 26 pm

#### comparing values of missing optional controllers
* ng1.2 returns `undefined`
* ng1.3 returns `null`

screen shot 2015-11-05 at 5 01 33 pm

### Maybe `angular.isObject()`
We cannot use `angular.isObject()` because their logic ([ng1.2.29](https://github.com/angular/angular.js/blob/v1.2.29/src/Angular.js#L445-L458), [ng1.3.20](https://github.com/angular/angular.js/blob/v1.3.20/src/Angular.js#L445-L461)) will not return true for a "function" typeof value (aka. Constructor in screenshots above). 

screen shot 2015-11-05 at 5 22 13 pm

### Lodash to the rescue!
Fortunately, lodash does a proper object check for `isObject()`. The only `typeof` values that register as an objects are `function` and `object` - [lodash source](https://github.com/lodash/lodash/blob/2.4.1/lodash.js#L143-L151).  

Using `_.isObject()` will return false for `undefined` and `null` but true for `object` and `function`. This is the desired behavior to prevent further execution of code.

bry4n: awesome!! yeah, that's pretty clear to see why object returns different value.

:heart: lodash

bry4n:
LGTM

src/rxBulkSelect/scripts/rxBatchActions.js
100stacks: @CITguy @bry4n this is a very informative thread on ng1.2.29 vs ng1.3.20 nuances!  I can definitely see the potential random behaviors without Lodash.

100stacks:
Great protips on ng1.3.x migration while balancing ng1.2.x support...LGTM :+1:

src/rxBreadcrumbs/docs/rxBreadcrumbs.midway.js
CITguy: These changes were pulled in from #1195 as well as implemented in `ng1.3` branch.

CITguy:
@jj4th: do you have any comments on the midways?

jj4th:
Doesn't seem like there's a lot to this PR.  Midway wise, you've updated the breadcrumbs midways, and that should be all that's needed.

CITguy:
@jj4th, thank you.

PR #1365, opened 2015-11-09
chore(npm): fix broken package

CITguy:
Yet another random NPM module updated upstream and caused Travis builds to break.

Correct broken package and shrinkwrap dependencies to prevent this from happening again.

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "grunt-contrib-connect": "~0.9.0",
     "grunt-contrib-copy": "~0.7.0",
     "grunt-contrib-cssmin": "~0.11.0",
-    "grunt-contrib-imagemin": "^0.9.2",
+    "grunt-contrib-imagemin": "^1.0.0",
     "grunt-contrib-jshint": "~0.10.0",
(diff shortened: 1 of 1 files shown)


bry4n:
LGTM

100stacks:
LGTM

CITguy:
Updated shrinkwrap with `devDependencies`.

CITguy:
Updated to correct broken package. We cannot shrinkwrap npm dependencies as long as we are using TravisCI.

100stacks:
@CITguy @bry4n looks like it's the same npm package dependency causing issue: https://www.npmjs.com/package/grunt-contrib-imagemin ...just updated a few hours ago.

CITguy:
@100stacks, yeah it updated to `1.0.0`, but we were using `^0.9.2` which installs `0.9.4` (last updated in March), so the fault lies in an upstream dependency of `grunt-contrib-imagemin`.

100stacks:
@CITguy,  definitely sounds like a lot to keep track of upstream dependencies.

PR #1371, opened 2015-11-10
style(progressBar): FRMW-327 Update Progressbar To Color 2.0

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-327

- [ ] Dev LGTM
- [ ] Design LGTM


src/progressbar/progressbar.less
```diff
@@ -74,12 +74,12 @@
 
 .progress-bar-secondary {
```
ssalinas24: @glynnis I don't see these .progress-bar-secondary  or .progress-bar-tertiary used anywhere in the framework. However, I tested in the browser to see what they would look like. Are the two middle bars in this screenshot correct?

progress-bar_secondary-and-tertiary

src/progressbar/progressbar.less
```diff
@@ -88,7 +88,7 @@
   //
   // Creates two color stops, start and end, by specifying a color and position for each color stop.
   // Color stops are not available in IE9 and below.
-  .horizontal(@start-color: #555; @end-color: #333; @start-percent: 0%; @end-percent: 100%) {
```
ssalinas24: @glynnis I don't see these .horizontal, .vertical, or .vertical-three-colors used anywhere. You happen to know what they are for?

CITguy:
LGTM

bry4n:
LGTM

src/styles/vars.less
```diff
@@ -121,7 +121,7 @@
 @actionGreenHover: saturate(@actionGreen, 15%);
 @infoOrange: #dd7800;
 @infoOrangeHover: saturate(@infoOrange, 15%);
-@infoBlue: #4580c2;
+@infoBlue: @blue-700;
```
CITguy: :x: This is a global variable capable of having a large effect if changed. I'd suggest changing any instance of `@infoBlue` in `progressbar.less` to `@blue-700` instead and revert this change.

ssalinas24:
@CITguy I have included changes based on your feedback. I have removed the changes to `@infoBlue`.

CITguy:
LGTM

src/progressbar/progressbar.less
glynnis: I think these are artifacts from the progress bar design/code being brought in from an external library. I believe we left this in because we wanted to provide a few options for color of progress bars, but hadn't yet decided or exposed them in the documentation.

We should consult with @ElementE about this to see how he wants to define these other colors and what documentation we might need around that.

PR #1374, opened 2015-11-10
style(rxTags): FRMW-328 Update rxTags To Color 2.0

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-328

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

Images have been updated to reflect the most recent commit.

#### Before:
Showing the Previous Colors
01_before_rxtags

#### Before:
(Includes tota11y) Color Contrast Ratios
01_before_rxtags-with-color-contrast

#### After:
I used the recommended colors from Zhen, but had to change some of the matches to get a good color ratio contrast as shown.
02_after_rxtags

#### Before:
03_before_focus

#### After:
04_after_focus

#### Before:
05_before_close

##### On Load
08_tag-on-load

##### Hover State
08_tag-hover

##### Focused State
08_tag-focused

##### Focused State and Hovering Again on the "X" button no longer produces the red icon
08_tag-focused-with-hover

### DIFF
--- a/src/rxTags/rxTags.less
+++ b/src/rxTags/rxTags.less
@@ -37,13 +37,13 @@ rx-tags {
     line-height: @rxTag-height;
     border-radius: @rxTag-height / 2;
     vertical-align: top;
-    background-color: #e0e0e0;
+    background-color: @gray-300;
     &:hover {
       cursor: default;
     }
 
     &:focus {
-      background-color: #007bff;
+      background-color: @blue-700;
       outline: none;
       .fa,
       .text,
@@ -59,27 +59,30 @@ rx-tags {
     .text {
       font-size: 12px;
       margin: 0 7px 0 10px;
-      color: #727272;
+      color: @gray-800;
     }
 
     .category {
       font-size: 10px;
       margin-right: 20px;
-      color: #878787;
+      color: @gray-800;
     }
 
     .fa-tag {
       margin-left: @rxTag-height / 2;
-      color: #999999;
+      color: @gray-600;
     }
 
     .fa-times {
       margin-right: @rxTag-height / 2;
-      color: #b8b8b8;
+      color: @gray-600;
       &:hover {
         cursor: pointer;
-        color: @warnRedHover;
+        color: @gray-800;
       }
     }
+  }
+  .tag:focus .fa-times:hover {
+    color: @white;
   }
 }


CITguy:
FYI

### Selected Before
screen shot 2015-11-13 at 11 07 04 am

### Selected After
screen shot 2015-11-13 at 11 06 58 am

CITguy:
LGTM

src/rxTags/rxTags.less
```diff
@@ -59,23 +59,23 @@ rx-tags {
...
+      color: @gray-600;
     }
 
     .fa-times {
       margin-right: @rxTag-height / 2;
-      color: #b8b8b8;
+      color: @gray-600;
       &:hover {
         cursor: pointer;
         color: @warnRedHover;
```
glynnis: Should this be updated?

CITguy: Good catch. I would suggest updating to 2.0 color variable.

ssalinas24: @glynnis and @CITguy Sure thing. I will update it.

ssalinas24:
@glynnis @CITguy I started taking a second look at this and I was wondering the following. Are these tags supposed to look different when in standard and when they are disabled states? No previous code seems to be doing anything different to them.

ssalinas24:
@glynnis Here's a situation that I came across. The Red "X" on hover while the tag is currently in focus. Should the red stay and convert to something other than `@red-700` or should the "X" perhaps just be white when focused and hovered in this situation?

### Hover While In Focus
#### Before:
07_before_focus-with-hover-close

#### After:
07_after_focus-with-hover-close-problem

CITguy:
@ssalinas24: When disabled:

* the outline around the entire group is removed
* the "X" is removed
* you cannot "select" a tag (blue background)

100stacks:
@ssalinas24 @glynnis this is probably already covered...documenting color choices for various components to meet recommended visual assistive guidelines, so that consumers of the framework understand color choice tradeoffs.

LGTM :+1:

glynnis:
@ssalinas24 I agree that for the selected state, we should use a white X because of the contrast issue. Another thing we should run by @ElementE -- I wonder if using red for Xs is something we can move away from, given the general direction we want to go with using color more selectively. Let's discuss with him.

ElementE:
Pending changes from Steven/myself

glynnis:
@ssalinas24 Can you update screen shots to reflect the changes you made?

ssalinas24:
@glynnis I did. I updated the very first comment box. They are the screens to the bottom of it.

glynnis:
Design LGTM

PR #1389, opened 2015-11-14
fix(bug): FRMW-451 currencyToPennies handle negative values

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-451

- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

---
Explicitly check for opening and closing parenthesis for negative numbers in `currencyToPennies`.

### DIFF
--- a/src/rxMisc/scripts/rxMisc.page.js
+++ b/src/rxMisc/scripts/rxMisc.page.js
@@ -25,6 +25,8 @@ exports.rxMisc = {
      * Transform `currencyStringOrArray` (USD) to an integer representing pennies or an array of integers representing
      * pennies. Built to reverse Angular's 'currency' filter. Do not pass in fractions of a penny because it will be
      * rounded with Math.round() which doesn't use bankers' rounding for negative numbers.
+     * NOTE: AngularJS representation of negative currency changes in   
+     * [AngularJS v1.4.4](https://code.angularjs.org/1.4.4/docs/api/ng/filter/currency)
      * @param {string|array<string>} currencyStringOrArray - Raw text or an array of raw texts as output by Angular's
      * `currency` filter.
      *
@@ -42,7 +44,7 @@ exports.rxMisc = {
             var resFloat = parseFloat(currencyString.split(' ')[0].replace(/[,$()]/g, '').trim());
 
             // Negative number
-            if (currencyString.indexOf('(') > -1 && currencyString.indexOf(')') > -1) {
+            if (_.first(currencyString) === '(' && _.last(currencyString) === ')') {
                 resFloat = -resFloat;
             }
 


CITguy:
Code looks good, but I'd like to see some tests to ensure that this solves what it claims it's solving.

100stacks:
@CITguy okay, I'll add some tests for this.

src/rxMisc/rxMisc.page.js
```diff
@@ -42,7 +42,7 @@ exports.rxMisc = {
             var resFloat = parseFloat(currencyString.split(' ')[0].replace(/[,$()]/g, '').trim());
```
CITguy: that line also removes `$` and `,` from the string

src/rxMisc/rxMisc.page.js
```diff
@@ -42,7 +42,7 @@ exports.rxMisc = {
...
 
             // Negative number
-            if (currencyString.indexOf('(') > -1 && currencyString.indexOf(')') > -1) {
+            if (_.first(currencyString) === '(' && _.last(currencyString) === ')') {
```
100stacks: Okay, no problem.  Our new expression checks `currencyString` rather than `resFloat`, so if `currencyString` = `$(100.00)` then it would fail.  This is only an issue if the `(` appears after the currency symbol.

Droogans: Check the angular currency filter. It is designed to only work using that format.

https://code.angularjs.org/1.3.1/docs/api/ng/filter/currency

@CITguy you might want to pay attention to the 1.3.1 docs (linked above) versus the [same docs for 1.4.7](https://code.angularjs.org/1.4.7/docs/api/ng/filter/currency).

### 1.3.1

-1234.56 becomes `($1,234.56)`

### 1.4.7

-1234.56 becomes ` -$1,234.56`

100stacks:
@Droogans, appreciate that.  Ok so the test for `($1,234.56)` will change after ng1.4.3 https://code.angularjs.org/1.4.3/docs/api/ng/filter/currency -- this appears to be last release that supports `($xxx)` negative numbers.  

@CITguy...I can annotate this in the `currencyToPennies` function comments.

100stacks:
@CITguy, `src/rxMisc/docs/rxMisc.midway.js` includes a negative currency test:
https://github.com/rackerlabs/encore-ui/blob/master/src/rxMisc/docs/rxMisc.midway.js#L123-L125 

@Droogans mentioned his original case for this was possibly a billing issue, where the field may have included two numbers.

src/rxMisc/scripts/rxMisc.page.js
```diff
@@ -25,6 +25,8 @@ exports.rxMisc = {
      * Transform `currencyStringOrArray` (USD) to an integer representing pennies or an array of integers representing
      * pennies. Built to reverse Angular's 'currency' filter. Do not pass in fractions of a penny because it will be
      * rounded with Math.round() which doesn't use bankers' rounding for negative numbers.
+     * NOTE: AngularJS representation of negative currency changes in   
+     * [AngularJS v1.4.4](https://code.angularjs.org/1.4.4/docs/api/ng/filter/currency)
```
CITguy: Is this note necessary considering that `currencyToPennies` is not something that uses the Angular `currency` filter?

CITguy:
@Droogans, @100stacks: ng1.4 issue noted. Thanks for the heads up.

LGTM

bry4n:
LGTM

PR #1359, opened 2015-11-05
style(colors): FRMW-322 Buttons/Actions

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-322

- [x] Dev LGTM
- [ ] Dev LGTM
- [x] Design LGTM

@glynnis  I have updated these screenshots to reflect my changes from my recent commit.

**Buttons**
Before:
01_before-01

After:
01_after-01

**Button Types**
Before:
01_before-02

After: 
01_after-02
@glynnis I think the red from before was a richer Red.

**Button With Animation**
Before:
01_before-03

After:
01_after-03

**Button Groups**
Before:
02_before

After:
02_after

**Action Links**
Before:
03_before

After:
03_after
@glynnis The Orange and Reds are hard to distinguish. Should we try converting the red back to its previous #cc0000?

**Action Link Colors**
Before:
04_before

After:
04_after

**Spinners with Action Links**
Before:
05_before

After:
05_after_contrast-trouble
@glynnis This is following the disabled gray that Eric and I establish for the disabled buttons, but is has a worse contrast on white background. Should we try a different gray for this scenario?

PR #1397, opened 2015-11-17
style(colors): FRMW-322 Buttons/Actions

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-322

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

Re-submitting this to run the visual regression. This is the same PR as previously submitted https://github.com/rackerlabs/encore-ui/pull/1359 One Dev and Design were already LGTM. Just need one more Dev LGTM.

### Buttons
Before:
01_before-01

After:
01_after-01

### Button Types
Before:
01_before-02

After: 
01_after-02
@glynnis I think the red from before was a richer Red.

### Button With Animation
Before:
01_before-03

After:
01_after-03

### Button Groups
Before:
02_before

After:
02_after

### Action Links
Before:
03_before

After:
03_after

### Action Link Colors
Before:
04_before

After:
04_after

### Spinners with Action Links
Before:
06_before

After:
06_after2

@glynnis This is the screenshot of the disabled link at `@gray-600`

### Delete-X Hover
This was a change I made since it was in common.less, however I realized that this was crossing into the Tables story.  I kept it in since it was a minor change. This is the before and after for it.

Before:
07_before-delete

After:
07_after-delete

### DIFF
--- a/src/rxApp/common.less
+++ b/src/rxApp/common.less
@@ -128,7 +128,7 @@ table {
     }
   }
   .fa-times.delete-x:hover {
-    color: @warnRedHover;
+    color: saturate(@red-700, 15%);
     cursor: pointer;
   }
 }
@@ -261,37 +261,37 @@ h6 {
 
 .msg-action,
 .msg-action a {
-  color: @actionGreen;
+  color: @green-700;
   &:hover,
   &:focus {
-    color: @actionGreen;
+    color: saturate(@green-700, 15%);
   }
 }
 
 .msg-warn,
 .msg-warn a {
-  color: @warnRed;
+  color: @red-700;
   &:hover,
   &:focus {
-    color: @warnRedHover;
+    color: saturate(@red-700, 15%);
   }
 }
 
 .msg-info,
 .msg-info a {
-  color: @infoOrange;
+  color: @orange-700;
   &:hover,
   &:focus {
-    color: @infoOrangeHover;
+    color: saturate(@orange-700, 15%);
   }
 }
 
 .msg-info-blue,
 .msg-info-blue a {
-  color: @infoBlue;
+  color: @blue-700;
   &:hover,
   &:focus {
-    color: @infoBlueHover;
+    color: saturate(@blue-700, 15%);
   }
 }
 
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -26,7 +26,7 @@
 // ## Oranges
 @orange-100: #ffecb3;
 @orange-500: #ffc107;
-@orange-700: #ff8f00;
+@orange-700: #d14904;
 @orange-accent: #ffd740;
 
 // ## Purples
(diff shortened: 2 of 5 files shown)


src/styles/vars.less
```diff
@@ -80,19 +80,19 @@
...
-@buttonNegativeBg: #cc0000;
+@buttonNegativeBg: @red-700;
 @buttonNegativeBgHover: lighten(@buttonNegativeBg, 10%);
-@buttonCancelBg: @cancel-gray;
+@buttonCancelBg: @gray-700;
 @buttonCancelBgHover: lighten(@buttonCancelBg, 10%);
-@buttonDisabledBg: #cfd2d4;
-@buttonSpinnerBg: @buttonColor;
+@buttonDisabledBg: @gray-300;
+@buttonSpinnerBg: @buttonColor; //Is this correct for ellipse Background?
```
ssalinas24: Actually I have to look into this color real quick.

glynnis:
@ssalinas24 The color you used for spinners seems a bit dark -- is that the same color for disabled links/buttons/etc.? When the spinner displays in that context, it's to prevent users from clicking on those links (as an API call is being made), so it should match any other disabled elements we use (even if the contrast ratio for that is a bit under par).

ssalinas24:
@glynnis I had previously used `@gray-300` but it was a worse contrast then what was already there. This time around I did `@gray-700` since i thought that's what we used to replace subdued title stuff right?

CITguy:
The colors of the links on the right side of the screenshot below (found in visual regression updates) are too close to distinguish for deuteranomaly color blindness (like myself). The top and bottom links look brown and the middle looks red. Should the top and bottom colors be next to each other, I *cannot* tell the difference.

screen shot 2015-11-18 at 1 18 56 pm

glynnis:
@CITguy Popping in to corroborate what @ssalinas24 mentioned -- right now there is a lot of work happening around revising designs for page actions and action links, and some of the discussion has been about whether colors or icons are truly helping the user, or if they are overused to the point of not having much meaning anymore.

Once there's a decision on that front (about whether colors will remain part of the action links pattern), we'll re-address the issues you mentioned here. I believe we'll have a decision on that front fairly soon, so I'm comfortable moving forward with this PR with the confidence that it will be addressed.

src/styles/vars.less
```diff
@@ -119,13 +119,13 @@
 @menuLinkLightBlue: #97b1d0;
 @menuLinkOrange: #ffa61b;
 @menuRed: #f37a7c;
-@warnRed: #eb2124;
+@warnRed: @red-700;
```
CITguy: This variable is used in:

* rxApp
* rxCharacterCount
* rxTags

Additionally, I'm positive that these variables are being consumed by other applications.

### Suggestion
I'd suggest:

0. revert these changes to reduce unwanted updates across other components
0. change the implementation in the specific Less files for the desired components
0. deprecate this variable (considering that the proper color variables are documented in "Color Usage" in new color docs)

src/styles/vars.less
```diff
@@ -119,13 +119,13 @@
...
 @warnRedHover: saturate(@warnRed, 15%);
-@actionGreen: #3ab661;
+@actionGreen: @green-700;
```
CITguy: This variable is used in:

* rxApp
* rxForm

Same suggestion as above.

src/styles/vars.less
```diff
@@ -119,13 +119,13 @@
...
 @actionGreenHover: saturate(@actionGreen, 15%);
-@infoOrange: #dd7800;
+@infoOrange: @orange-700;
```
CITguy: This variable is used in:

* rxApp

Same suggestion as above

src/styles/vars.less
```diff
@@ -119,13 +119,13 @@
...
 @infoOrangeHover: saturate(@infoOrange, 15%);
-@infoBlue: #4580c2;
+@infoBlue: @blue-700;
```
CITguy: This variable is used in:

* rxApp
* rxPaginate

Same suggestion as above

glynnis:
This has an LGTM from @ElementE

ssalinas24:
@CITguy I will update the changes that you have pointed out with the variable colors.

src/rxApp/common.less
```diff
@@ -128,7 +128,7 @@ table {
     }
   }
   .fa-times.delete-x:hover {
-    color: @warnRedHover;
+    color: saturate(@red-700, 15%);
```
ssalinas24: @CITguy This was a change I made because it was in common.less that i was looking at, however I realized that this was crossing into the Tables story. I kept it in since it was a minor change. The rest of Table changes can be tackled by the table story. I attached the screenshots in the initial comment with the other screenshots.

ssalinas24:
@CITguy I have reverted the @actionGreen and other general variables that you had mentioned to their initial hex colors. I then updated the common.less file for the links to have 2.0 color variables.

CITguy:
LGTM

PR #1400, opened 2015-11-18
style(tooltips): FRMW-339 Update Tooltips to Color 2.0

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-339

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### Before
01_before

### After
02_after

### DIFF
--- a/src/tooltips/tooltips.less
+++ b/src/tooltips/tooltips.less
@@ -7,8 +7,8 @@
 @zindexTooltip: 1070;
 @tooltipMaxWidth: 200px;
 @tooltipHeaderColor: @white;
-@tooltipColor: #f1f1f1;
-@tooltipBg: #878787;
+@tooltipColor: @white;
+@tooltipBg: @gray-700;
 @tooltipOpacity: 1;
 
 //** Tooltip arrow width


src/tooltips/tooltips.less
```diff
@@ -7,8 +7,8 @@
 @zindexTooltip: 1070;
```
ssalinas24: @CITguy Should I move these top few lines into the vars.less file?

CITguy:
FYI: there's actually more contrast between the tooltip bg and the tabs in the "Before"

*before* - more contrast between tooltip gray and tab blue
screen shot 2015-11-18 at 4 39 43 pm

*after* - less contrast between tooltip gray and tab blue. I can barely tell where the tooltip ends and the tab begins.
screen shot 2015-11-18 at 4 40 01 pm

I'm sure this is something that is out of scope of this PR, but maybe adding a white border to the tooltip might be enough to visually separate the two colors.

CITguy:
Code-wise. LGTM

glynnis:
@CITguy FWIW, the tab colors are changing in another PR, so maybe it's worth merging this as-is and checking the contrast of the two once both components are updated. If there's still a contrast problem then, we can open a new story/PR. 

Can you make a note to give the contrast a once-over once tabs and tooltips are merged? I'm working on tabs now but blocked by @ElementE for the moment because we need to add some darker blues to the palette.

@ssalinas24 Not entirely sure just from the screen shots, but it looks like the buttons here are using the old blues. Are you up to date with master? If so, there might be some mark-up on the tooltip demo that needs to be corrected.

glynnis:
Oh. Der. Just realized that the reason the buttons have old colors is because that PR is still open :)

CITguy:
@glynnis, JIRA created (https://jira.rax.io/browse/FRMW-511) to capture request.

100stacks:
LGTM

PR #1423, opened 2015-11-20
style(rxNotify): FRMW-329 Update rxNotify To Color 2.0

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-329

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

## Before
01_before

02_before

## After
01_after-01

01_after-02

02_after


glynnis:
@ssalinas24 Could you post some screen shots here? It looks like the only changes picked up by visual regression were for the "info" style of rxNotify.

ssalinas24:
@glynnis Yep coming right up.

ssalinas24:
@glynnis I just updated the comment with screenshots. However, I just realized that I may have to update the white spinner/loader on the blue notification.

CITguy:
Should the links be different colors? The BEFORE links were different colors from the main text to visually communicate that they were clickable links. The AFTER does not have this.

CITguy:
Additionally, should the warning notifications be red or orange? The AFTER has both the warning and error with the same colors.

ssalinas24:
@CITguy You are correct, and I question that I had. @glynnis and @ElementE Do the links need to be different colors, or can we just have an underline below it?

ssalinas24:
@CITguy They are actually two different colors, but comes back to our problem of the Orange and Red being too similar to one another.

glynnis:
@ssalinas24 Yes, you may have to write a property for the spinner in rxNotify context once my rxSpinner changes get merged.

Re: @CITguy's comment about distinguishing links, I would worry about making them a different color, but I do think they need to be underlined. Here's an example of what that can look like from Rdio that I think works really nicely:
screen shot 2015-11-23 at 2 30 33 pm

Also worth noting is the fact that I think the icon color should match the text, so I'd change your icon colors to white for the orange and red notifications. I'm also wondering if you need to use the darkest variables for those, since this is a background rather than text, so the contrast might shake out differently and be less of an issue. Did you try any of the lighter red or orange colors as backgrounds? Did they pass the contrast check?

CITguy:
Can we get SVG files for those icons? It would make it much easier to style.

CITguy:
@ssalinas24, please rebase at your earliest convenience

ssalinas24:
@glynnis and @CITguy I spoke with @ElementE and we decided to change the orange to `@orange-100`. Please let us know if these solutions work for the both of you. I need to look at the icon image paths  and correct now with the merge update. I will also look into the SVG solution.

grunt-tasks/options/imagemin.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
         files: [{
             expand: true,
             cwd: 'src/',
-            src: '*/images/*',
+            src: '**/images/*',
```
ssalinas24: @CITguy I added the change we spoke about for the `/images` folder.

ssalinas24:
@CITguy I was able to use the SVG images as background images. They scale with no pixelation and look awesome. I tested in Chrome and Firefox just to be sure, and everything is fine. I still have the previous .png files in the folder as well. Lets see if these SVG files pass the tests and build.

src/components/rxNotify/images/error-stop.svg
```diff
@@ -0,0 +1,46 @@
...
+</style>
+<g>
+	<path class="st0" d="M10.3,4.4c-3.1,0-5.6,2.5-5.6,5.6s2.5,5.6,5.6,5.6c3.1,0,5.6-2.5,5.6-5.6S13.4,4.4,10.3,4.4z M13.8,10
+		c0,0.6-0.1,1.1-0.4,1.6L8.7,6.9c0.5-0.3,1-0.4,1.6-0.4C12.2,6.5,13.8,8.1,13.8,10z M6.8,10c0-0.6,0.1-1.1,0.4-1.6l4.7,4.7
+		c-0.5,0.3-1,0.4-1.6,0.4C8.4,13.5,6.8,11.9,6.8,10z"/>
+	<path class="st0" d="M19.8,5.4l-4.9-5C14.6,0.2,14.4,0,14,0H6.6C6.2,0,5.9,0.2,5.7,0.4l-5.2,5C0.3,5.6,0,6,0,6.3v7.5
+		c0,0.3,0.3,0.6,0.5,0.9l5.1,5c0.2,0.2,0.6,0.4,1,0.4H14c0.3,0,0.6-0.1,0.9-0.4l4.8-5c0.2-0.2,0.2-0.6,0.2-0.9V6.3
+		C20,6,20,5.6,19.8,5.4z M19,13.8L14,19H6.6L1,13.8V6.3L6.6,1H14l5,5.3V13.8z"/>
+</g>
+<g>
```
CITguy: Please remove lines 16-45

ssalinas24:
@CITguy It looks like the SVG files are safe to proceed with, correct? If so, can I remove the PNG files that were previously used?

CITguy:
@ssalinas24: Yes, good to go with SVG files :+1:

100stacks:
LGTM

glynnis:
LGTM! Going to count that as a Design LGTM since @ElementE is out and it sounds like he's more or less already signed off on this one. :+1:

src/components/rxNotify/rxNotify.less
```diff
@@ -68,14 +74,20 @@
     height: 100%;
     padding: 0 15px;
     content: "\00D7";
-    .opacity(.5);
-    color: @black;
+    .opacity(.75);
+    color: @gray-950;
```
ssalinas24: @glynnis and @ElementE I noticed that the "X" close/dismiss icon was set to black and 50%. They were still not passing the tota11y contrast ratio. I tried `@gray-600` for the on load state and `@gray-800` for hover, since we are using that on modals. However, they didn't work against these rxNotify background colors. I was able to stick with the same `@gray-950` used on the text and an opacity of 75%. On hover, they go 100% `@gray-950`. This passes the contrast ratio. Let me know if you see any problem with this.

01_after-01

01_after-02

src/components/rxNotify/rxNotify.less
```diff
@@ -68,14 +74,20 @@
...
     &:hover,
     &:focus {
-      .opacity(0.8);
+      .opacity(1);
       text-decoration: none;
     }
   }
+  .notification-error {
```
ssalinas24: I also added this section so that the "X" icon would match the text on the right. The previous "X" was left black and not passing the color contrast ratio when on the red notification Error div. On hover, it just stays white as well since the code comes later/after the :hover code.

01_after-02

glynnis:
Xs look good to me :+1:

src/components/rxNotify/rxNotify.less
```diff
@@ -68,14 +74,20 @@
...
     padding: 0 15px;
     content: "\00D7";
-    .opacity(.5);
-    color: @black;
+    .opacity(.75);
+    color: @gray-950;
     &:hover,
     &:focus {
-      .opacity(0.8);
+      .opacity(1);
```
CITguy: If we're going with full opacity, this mixin is no longer needed. Please remove.

CITguy:
LGTM

PR #1427, opened 2015-11-24
style(tabs): Update tabs to use 2.0 colors

glynnis:
https://jira.rax.io/browse/FRMW-338

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### Before
screen shot 2015-11-24 at 2 00 42 pm
screen shot 2015-11-24 at 2 00 55 pm

### After
screen shot 2015-11-24 at 2 10 08 pm

### DIFF
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -107,10 +107,6 @@
 @tableHeaderBorder: #6e6e6e;
 @subtableBorder: #adadad;
 
-// ## Tabs and Account Info
-@tabBorder: #4480c1;
-
-
 // # Text colors
 
 // ## General
--- a/src/components/tabs/tabs.less
+++ b/src/components/tabs/tabs.less
@@ -2,7 +2,7 @@
 
 .nav-tabs {
   display: flex;
-  border-bottom: 9px solid @tabBorder;
+  border-bottom: 9px solid @blue-700;
   clear: both;
 
   li {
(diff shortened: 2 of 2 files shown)


src/styles/vars.less
```diff
@@ -107,10 +107,6 @@
 @tableHeaderBorder: #6e6e6e;
 @subtableBorder: #adadad;
 
-// ## Tabs and Account Info
-@tabBorder: #4480c1;
```
glynnis: Took this out since the last reference to it in the Framework was just removed. I feel pretty confident that projects aren't relying on this heavily, but @CITguy could you be sure when this goes out to have a big warning that this variable will no longer be in `vars.less` so projects understand to update any places they may be using it? It can be replaced with `@blue-700` if they're on a version of the Framework that includes the new palette.

ssalinas24:
LGTM

100stacks:
LGTM

ElementE:
LGTM

src/styles/vars.less
CITguy: Can do :+1:

CITguy:
LGTM - :heart: the new colors

src/styles/vars.less
CITguy: It didn't occur to me before I merged this PR, but I'd like to reduce the opportunities for introducing breaking changes. Variables are included in breaking change updates. As such, I've submitted #1437 to readd this information back to the file for compatibility reasons.

glynnis: Der. Thanks for catching this!

PR #1428, opened 2015-11-24
style(modals): FRMW-332 Updating Modals to Color 2.0

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-332

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

I only found modals used in the following LESS files.
* vars.less
* common.less
* rxModalAction.less
* rxApp.less

and in the following components shown below.

## BEFORE
01_before

### Modals
http://localhost:9001/#/styleguide/modals
NOTE: Keep in mind that the current rxNotify PR will take care of the notification colors.
01_after

### rxModalAction
http://localhost:9001/#/components/rxModalAction
NOTE: Form colors will be corrected in rxForm
02_after-rxmodalaction

### Wells
http://localhost:9001/#/styles/wells
03_after-wells

### rxActionMenu
http://localhost:9001/#/components/rxActionMenu
04_after-rxactionmenu

### rxFeedback
http://localhost:9001/#/components/rxFeedback
05_after-rxfeedback

### rxBulkSelect
http://localhost:9001/#/components/rxBulkSelect
NOTE: Table colors will be corrected when they are addressed in the Tables color story.
06_after-rxbulkselect

### Submit Feedback
07_after-submit-feedback

### DIFF
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -99,7 +99,7 @@
 // # Borders
 
 // ## General
-@pageDivider: #e6e6e5;
+@pageDivider: #e6e6e5; // Deprecated
 @appHelpBorder: #424242;
 
 // ## Tables
@@ -141,8 +141,8 @@
 @subduedText: #bbbbbb;
 @subduedTextHover: darken(@subduedText, 15%);
 @subduedTitle: #989998;
-@closeText: #e5e5e5;
-@closeTextHover: darken(@closeText, 10%);
+@closeText: #e5e5e5; // Deprecated
+@closeTextHover: darken(@closeText, 10%); // Deprecated
 
 // ## Tables
 @tableHeaderText: @white;
--- a/src/components/rxApp/rxApp.less
+++ b/src/components/rxApp/rxApp.less
@@ -446,7 +446,7 @@
     .rx-feedback .modal-link {
       background: url(images/icon-feedback.png) no-repeat 20px calc(~"50% + 1px");
       vertical-align: middle;
-      color: @menuLinkOrange;
+      color: @orange-500;
       padding: 0 21px 0 41px;
       float: right;
     }
--- a/src/components/rxModalAction/rxModalAction.less
+++ b/src/components/rxModalAction/rxModalAction.less
@@ -62,7 +62,7 @@
 
 .modal-header {
   padding: @modalPadding;
-  border-bottom: 1px solid @pageDivider;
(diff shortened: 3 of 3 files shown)


src/styles/vars.less
```diff
@@ -141,8 +141,8 @@
 @subduedText: #bbbbbb;
 @subduedTextHover: darken(@subduedText, 15%);
 @subduedTitle: #989998;
-@closeText: #e5e5e5;
-@closeTextHover: darken(@closeText, 10%);
+@closeText: @gray-200;
+@closeTextHover: @gray-800;
```
ssalinas24: @glynnis Should we keep these `@closeText` variables? They refer to the close button "X" icon. Or should we just replace them in the rxModalAction.less file?

ssalinas24: @glynnis `@pageDivider` is also used for the border. Should we keep it and use it in other places? or just replace it with the color variable?

glynnis: I'd replace them. There's also the option of making them component specific by changing the variable name to `@rxModal-close-text`, but I'm guessing the variable isn't referred to so often that keeping something in `vars.less` adds much value.

glynnis: Same for `@pageDivider` -- I'd replace it, unless you see that it's used widely in the Framework.

glynnis:
I know that the X to close isn't actually text, but keeping with higher contrast for better accessibility, could we darken it a bit? On my non-mac/non-retina monitors it's not as contrasty. Maybe somewhere around `@gray-400`?

ssalinas24:
@glynnis the new "After" image above shows `@gray-600` on the close button. Do we agree that this is not critical text and is fine as is?

glynnis:
I agree that it's not critical text, also it's been vastly improved over previous colors :) I feel good about `@gray-600`.

Guessing that the warning message here will update with the icon color once the rxNotify PR gets merged?

LGTM

ssalinas24:
@glynnis I looked through the other demo modal usage and found the components that i have shown in my initial comment for this PR. Let me know if you think I have missed anything. Looks like the color refactor for Forms and Tables will take care of the red contrast labels shown in my images.

glynnis:
LGTM :+1:

CITguy:
LGTM

parlarjb:
LGTM

PR #1432, opened 2015-11-30
style(rxAccountInfo): Update rxAccountInfo to use 2.0 colors

glynnis:
https://jira.rax.io/browse/FRMW-330

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### Before
screen shot 2015-11-30 at 2 27 35 pm
User dropdown in Encore Cloud

screen shot 2015-11-30 at 2 27 57 pm

screen shot 2015-11-30 at 2 28 05 pm

### After
screen shot 2015-11-30 at 2 26 33 pm

screen shot 2015-11-30 at 2 28 21 pm

screen shot 2015-11-30 at 2 28 30 pm

### DIFF
--- a/src/components/rxAccountInfo/rxAccountInfo.less
+++ b/src/components/rxAccountInfo/rxAccountInfo.less
@@ -10,10 +10,10 @@
     margin-right: 5em;
   }
   .account-info-label {
-    color: #777777;
+    color: @gray-700;
   }
   .account-info-data {
-    color: #777777;
+    color: @gray-700;
     font-size: 26px;
     img {
       height: 35px;
@@ -28,10 +28,10 @@
 .rx-page {
   .account-info-banner {
     clear: both;
-    background: #d1d1d1;
+    background: @white;
     width: 100%;
     border-top: 1px solid;
-    border-color: #b9b9b9;
+    border-color: @gray-500;
   }
 
   ul.account-info-text {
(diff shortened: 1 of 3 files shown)


src/components/rxAccountInfo/rxAccountInfo.less
```diff
@@ -41,37 +41,28 @@
...
           }
-          &:after { // arrow color
-            color: #6c6c6c;
-          }
         }
         select {
           font-size: 12px;
-          background: #959595; // select box background
-          border: 1px solid #6c6c6c;
-          color: @white;
```
glynnis: These line deletions are so that the user dropdown is drawing on the standard styles for dropdowns from rxForm, with fewer customizations.

ElementE:
Design: LGTM

glynnis:
Had to run local Framework against local Cloud to be able to see the user dropdown, so I created a story to add one of those to the demo, which is here: https://jira.rax.io/browse/FRMW-535

100stacks:
@glynnis just curious, I noticed these in the screenshots above and in the demo pages.  Why do we use them?

screen shot 2015-11-30 at 3 36 02 pm

glynnis:
@100stacks I think that's to represent the placement/size of badges. Since account badges are in a private repo they were not something we could expose/use in Framework docs. Or at least I think that's why? It was a while ago that this demo page was originally built and my memory isn't the best.

100stacks:
@glynnis appreciate the background on the badges.  That makes sense considering the framework is open-source. :+1: 

LGTM

src/components/rxAccountInfo/rxAccountInfo.less
CITguy: These line deletions are ok, but the user select is using old dropdown styles and should be replaced with rxSelect.

Please do the following:

0. Add `rx-select` to the `<select>` in `src/components/rxApp/templates/rxAccountUsers.html`
  1. `<select rx-select ...>`
  1. This uses the correct, modern select dropdown styling.
0. Remove [lines 223-225](https://github.com/rackerlabs/encore-ui/blob/master/src/components/rxForm/rxForm.less#L223-L225) of `rxForm.less` (they are no longer necessary)

glynnis: Done :ok_hand:

CITguy:
LGTM

PR #1433, opened 2015-11-30
chore(rxButton): FRMW-436-removed bottom margin

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-436

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

---
Added `margin-bottom: 0px;` to `rxInput` -> `.button` to remove the `10px` bottom margin inherited from `rxButton`.

### DIFF
--- a/src/components/rxForm/rxForm.less
+++ b/src/components/rxForm/rxForm.less
@@ -442,6 +442,10 @@ rx-field-name {
           justify-content: flex-start;
           align-items: stretch;
 
+         .button {
+            margin-bottom: 0px;
+          }
+
           & > input,
           & > textarea {
             flex: 1 1 auto; // as dynamic flex item


100stacks:
### Before: Firefox

screen shot 2015-11-30 at 3 03 38 pm

### After: Firefox

screen shot 2015-11-30 at 3 04 28 pm

100stacks:
### Before: Chrome

screen shot 2015-11-30 at 3 02 54 pm

### After: Chrome

screen shot 2015-11-30 at 3 02 05 pm

src/components/rxButton/rxButton.less
```diff
@@ -10,7 +10,6 @@
   border: 0;
   font-size: 14px;
   padding: 7px 13px;
-  margin-bottom: 10px;
```
ssalinas24: @100stacks This one change is causing some bad alignments. I think the JIRA story calls for it to be targeted within rx-input, because removing it causes the following to happen:

### Problems in Demo
screen shot 2015-11-30 at 5 59 03 pm

### Problem in rxSpinner
screen shot 2015-11-30 at 5 58 56 pm

### Problem with Buttons Above Tables
screen shot 2015-11-30 at 5 56 14 pm

ssalinas24: @100stacks and @glynnis I think it would be best to put line 13 back in place. Then only target the `.button` when inside of rx-input. Which would probably call for an update in rxForm.less. Some where within lines 438 - 470. Actually probably right before 470 bracket. 
 ```
rx-input {
          // as a flex container
          display: flex;
          flex-flow: row nowrap;
          justify-content: flex-start;
          align-items: stretch;

          & > input,
          & > textarea {
            flex: 1 1 auto; // as dynamic flex item

            min-height: @rxForm-input-min-height;

            &, &.ng-valid {
              .validInput();
            }

            &.ng-invalid.ng-dirty {
              .invalidInput();

              // DISABLED STYLING
              // (these should trump all other styles)
              &[disabled] {
                .disabledInput();
              }
            }

            // DISABLED STYLING
            // (these should trump all other styles)
            &[disabled] {
              .disabledInput();
            }
          }
```

you would probably add 
```
.button {
  margin-bottom: 0px;
}
```

@glynnis Am I correct? Am i missing anything that would be added here like `button` tag or something?

100stacks: @ssalinas24 appreciate the feedback.  Comments incorporated. :+1: 

@ssalinas24 @glynnis it looks like this doesn't quite solve the `10px` bottom margin problem, so will keep at it.

---
screen shot 2015-12-01 at 10 16 25 am

100stacks: @ssalinas24 appreciate the feedback and the second set of :eyes: 
@ssalinas24 @glynnis looks like this should work. :+1:

ssalinas24:
LGTM

Looks like it will be safe and only target `.button` when inside of `<rx-input>`.

ssalinas24:
@100stacks You may want to update your initial comment to reflect the final solution.

100stacks:
@ssalinas24 initial comment updated. :+1:

CITguy:
LGTM

parlarjb:
LGTM

PR #1450, opened 2015-12-11
style(vars): FRMW-545 Rename Variables with Proper Component Prefix

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-545

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -149,26 +149,24 @@
 /*
  * Buttons
  */
-@optionHighlightBg: @blue-500;
-@buttonColor: @white;
-@buttonColorDisabled: @gray-600;
-@buttonDefaultBg: @blue-700;
-@buttonDefaultBgHover: lighten(@buttonDefaultBg, 10%);
-@buttonPositiveBg: @green-700;
-@buttonPositiveBgHover: lighten(@buttonPositiveBg, 6%);
-@buttonNegativeBg: @red-700;
-@buttonNegativeBgHover: lighten(@buttonNegativeBg, 10%);
-@buttonCancelBg: @gray-700;
-@buttonCancelBgHover: lighten(@buttonCancelBg, 10%);
-@buttonDisabledBg: @gray-300;
-@buttonSpinnerBg: @white;
+@rxButton-default-text-color: @white;
+@rxButton-default-background-color: @blue-700; // TODO replace misused text colors and others in rxButton
+@rxButton-default-hover-background-color: lighten(@rxButton-default-background-color, 10%); // TODO replace misused text colors and others in rxButton
+@rxButton-positive-background-color: @green-700;
+@rxButton-positive-hover-background-color: lighten(@rxButton-positive-background-color, 6%);
+@rxButton-negative-background-color: @red-700;
+@rxButton-negative-hover-background-color: lighten(@rxButton-negative-background-color, 10%);
+@rxButton-cancel-background-color: @gray-700;
+@rxButton-cancel-hover-background-color: lighten(@rxButton-cancel-background-color, 10%);
+@rxButton-disabled-text-color: @gray-600;
+@rxButton-disabled-background-color: @gray-300; // TODO replace misused text colors
+@rxButton-spinner-background-color: @white;
 
 /*
  * Button Groups (rxButton)
  */
-// TODO: rename with proper component prefix
-@buttonGroupBorder: 2px solid #eaeaea;
-@buttonGroupBorderRadius: 4px;
+@rxButton-button-group-border: 2px solid @gray-200;
+@rxButton-button-group-border-radius: 4px;
 
 /*
  * rxForm
@@ -310,22 +308,25 @@
 /*
  * progressbar
  */
-// TODO: rename with proper component prefix
 //** Background color of the whole progress component
-@progress-bg: @gray-50;
-//** Default progress bar color
-@progress-bar-bg: @blue-100;
-//** Secondary progress bar color
-@progress-bar-secondary-bg: @blue-700;
-//** Tertiary progress bar color
-@progress-bar-tertiary-bg: @green-700;
-@font-size-base: @appFontSize;
-@font-size-large: ceil((@font-size-base * 1.25));
-@font-size-small: ceil((@font-size-base * 0.85));
+@progressbar-background-color: @gray-50;
+//** Default progress bar colors
+@progressbar-default-text-color: @gray-900;
+@progressbar-default-background-color: @blue-100;
+@progressbar-border-color: @gray-400;
+//** Secondary progress bar colors
+@progressbar-secondary-text-color: @white;
+@progressbar-secondary-background-color: @blue-700;
+//** Tertiary progress bar colors
+@progressbar-tertiary-text-color: @white;
+@progressbar-tertiary-background-color: @green-700;
+//** Font sizing for progress bar
+@progressbar-font-size-base: @appFontSize;
+@progressbar-font-size: 12px;
 //** Unit-less `line-height` for use in components like buttons.
-@line-height-base: 1.428571429; // 20/14
+@progressbar-line-height-base: 1.428571429; // 20/14
 //** Computed "line-height" (`font-size` * `line-height`) for use with `margin`, `padding`, etc.
-@line-height-computed: floor((@font-size-base * @line-height-base)); // ~20px
+@progressbar-line-height-computed: floor((@progressbar-font-size-base * @progressbar-line-height-base)); // ~20px
 
 /*
  * rxBreadcrumbs
@@ -334,10 +335,10 @@
 @rxBreadcrumbs-arrow-width: 9px;
 
 /*
- * Modals *TODO* Rename variables based on component name
+ * Modals
  */
-@modalPadding: 30px;
-@modalBodyPadding: 50px;
+@rxModalAction-padding: 30px;
+@rxModalAction-body-padding: 50px;
 
 
 /* ======================================== *\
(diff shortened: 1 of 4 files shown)


ssalinas24:
FYI, I didn't include Screenshots because only the variable names changed, and not the colors.

glynnis:
LGTM :clap:  :clap:  :clap:  :clap:  :clap:  :clap:  :clap:  :clap:  :clap:  :clap:

src/styles/vars.less
```diff
@@ -241,22 +240,26 @@
...
-@progress-bar-secondary-bg: @blue-700;
-//** Tertiary progress bar color
-@progress-bar-tertiary-bg: @green-700;
-@font-size-base: @appFontSize;
-@font-size-large: ceil((@font-size-base * 1.25));
-@font-size-small: ceil((@font-size-base * 0.85));
+@progressbar-background-color: @gray-50;
+//** Default progress bar colors
+@progressbar-text-color: @gray-900;
+@progressbar-background-default-color: @blue-100;
```
CITguy: I'd rename this to `@progressbar-default-background-color` to follow same naming convention for the secondary and tertiary colors.

src/styles/vars.less
```diff
@@ -241,22 +240,26 @@
...
-//** Secondary progress bar color
-@progress-bar-secondary-bg: @blue-700;
-//** Tertiary progress bar color
-@progress-bar-tertiary-bg: @green-700;
-@font-size-base: @appFontSize;
-@font-size-large: ceil((@font-size-base * 1.25));
-@font-size-small: ceil((@font-size-base * 0.85));
+@progressbar-background-color: @gray-50;
+//** Default progress bar colors
+@progressbar-text-color: @gray-900;
```
CITguy: I'd rename this to `@progressbar-default-text-color` to follow same naming convention for secondary and tertiary colors.

src/styles/vars.less
```diff
@@ -241,22 +240,26 @@
...
-@progress-bg: @gray-50;
-//** Default progress bar color
-@progress-bar-bg: @blue-100;
-//** Secondary progress bar color
-@progress-bar-secondary-bg: @blue-700;
-//** Tertiary progress bar color
-@progress-bar-tertiary-bg: @green-700;
-@font-size-base: @appFontSize;
-@font-size-large: ceil((@font-size-base * 1.25));
-@font-size-small: ceil((@font-size-base * 0.85));
```
CITguy: Please place definitions for the old variables in the bottom "Deprecated" section and assign them to their new variables.

```less
@progress-bg: @progressbar-background-color;
@progress-bar-bg: @progressbar-background-default-color;
@progress-bar-secondary-bg: @progressbar-secondary-background-color;
@progress-bar-tertiary-bg: @progressbar-tertiary-background-color;
...
```
This supports backward compatibility and documents which variable to use instead.

src/styles/vars.less
```diff
@@ -241,22 +240,26 @@
...
+@progressbar-border-color: @gray-400;
+//** Secondary progress bar colors
+@progressbar-secondary-text-color: @white;
+@progressbar-secondary-background-color: @blue-700;
+//** Tertiary progress bar colors
+@progressbar-tertiary-text-color: @white;
+@progressbar-tertiary-background-color: @green-700;
+//** Font sizing for progress bar
+@progressbar-font-size-base: @appFontSize;
+@progressbar-font-size-large: ceil((@progressbar-font-size-base * 1.25));
```
CITguy: Is this variable necessary? I'm not seeing the large font being used.

src/styles/vars.less
```diff
@@ -241,22 +240,26 @@
...
+@progressbar-font-size-small: ceil((@progressbar-font-size-base * 0.85));
```
CITguy: If large font not used, I'd suggest:
* rename to `@progressbar-font-size`
* change value to an explicit value (based on the output of `ceil()`)

src/styles/vars.less
```diff
@@ -112,9 +112,8 @@
 /*
```
glynnis: Just looked at this again because of some work I'm doing in another story -- do lines 90-110 need to be updated as well?

ssalinas24: @glynnis Sure thing, I can include them within this PR. I will update.

ssalinas24:
@CITguy Thanks for the feedback. I will focus on making the changes today.

src/styles/vars.less
ssalinas24: @CITguy and @glynnis I don't see this variable being used right now either. I assumed this was a variable created to make the framework more flexible for users in the future, and to provide them an option for adjusting the font size. I see it as a cool idea to offer an option. What do you guys think, keep it in?

src/styles/vars.less
ssalinas24: @CITguy and @glynnis or do we go with Ryan's idea in this later comment about consolidating to one variable called `@progressbar-font-size` and letting the developers just adjust the `ceil()` em size?

glynnis: @ssalinas24 Right now I don't think we've come across any requests for smaller font sizes/progress bars. I would opt for a single font-size var, which doesn't rely on a calculation or ems, since ems can change the size of text depending on what context something appears in (and we don't always know what context a component will be used in). Could you set this to a vanilla pixel value?

That said, I think having some varied sizes for progress bars and their fonts is a good idea. Could you capture that in a story for the Framework backlog?

cc/ @CITguy

src/styles/vars.less
```diff
@@ -95,26 +95,24 @@
...
+@rxButton-default-hover-background-color: lighten(@rxButton-default-background-color, 10%); // TODO replace misused text colors and others in rxButton
+@rxButton-positive-background-color: @green-700;
+@rxButton-positive-hover-background-color: lighten(@rxButton-positive-background-color, 6%);
+@rxButton-negative-background-color: @red-700;
+@rxButton-negative-hover-background-color: lighten(@rxButton-negative-background-color, 10%);
+@rxButton-cancel-background-color: @gray-700;
+@rxButton-cancel-hover-background-color: lighten(@rxButton-cancel-background-color, 10%);
+@rxButton-disabled-text-color: @gray-600;
+@rxButton-disabled-background-color: @gray-300; // TODO replace misused text colors
+@rxButton-spinner-background-color: @white;
```
ssalinas24: @glynnis I updated the button vars that you pointed out. Listed on this line and above.

src/styles/vars.less
```diff
@@ -357,3 +358,41 @@
 @inputColorError: @red; // replaced by @rxForm-error-text-color
 @inputBackgroundError: @inputColorError; // replaced by @rxForm-error-accent-color
 @rxCheckbox-color-selected: @white; // not used, replaced by @rxForm-input-background-color
+
+// Buttons
```
ssalinas24: @CITguy  I updated deprecated versions of the buttons, button groups, progressbar, and modals below.

ssalinas24:
@glynnis and @CITguy I have reviewed the feedback and implemented changes. Please proceed with review.

src/styles/vars.less
```diff
@@ -357,3 +358,41 @@
...
+// deprecated
+@optionHighlightBg: @blue-500;
```
ssalinas24: @glynnis and @CITguy Do you happen to know what this variable is for? It's not used any where. So I took it out of the 2.0 vars section, but left it in this deprecated section. Let me know if i need to add it back into the 2.0 section.

src/styles/vars.less
CITguy: I'd normalize to a single `@progressbar-font-size` rather than have a `*-base` and `*-large`.

src/styles/vars.less
CITguy: I'm not opposed to different sizes, but for now, we should only focus on a single size until updated in a future story.

src/styles/vars.less
CITguy: That variable doesn't ring a bell. I'm fine with keeping in deprecated section.

glynnis: Same here :+1:

ssalinas24:
@CITguy All feedback was addressed and no changes needed. This just needs the Dev Reviews.

src/styles/vars.less
```diff
@@ -256,22 +254,25 @@
...
+@progressbar-border-color: @gray-400;
+//** Secondary progress bar colors
+@progressbar-secondary-text-color: @white;
+@progressbar-secondary-background-color: @blue-700;
+//** Tertiary progress bar colors
+@progressbar-tertiary-text-color: @white;
+@progressbar-tertiary-background-color: @green-700;
+//** Font sizing for progress bar
+@progressbar-font-size-base: @appFontSize;
+@progressbar-font-size: ceil((@progressbar-font-size-base * 0.85));
```
CITguy: What is the computed result of the `ceil()` function? I'd rather have an explicit value than a computed one.

glynnis: Related: https://github.com/rackerlabs/encore-ui/pull/1450#discussion_r47961904

ssalinas24: @CITguy `@appFontSize: 13px;` is in the vars, then the calculation is run on in this line. Then in the chrome inspector `.progress-bar` is getting a final result of `font-size: 12px;`. Which is fine. So without the ceil i guess the calculation would be 13*0.85 = 11.05.

ssalinas24: @CITguy Nevermind, i see the comment glynnis referred to. I will update.

ssalinas24:
@CITguy  Font Size was updated and rebased. This should pass once the tests clear.

CITguy:
LGTM

100stacks:
@ssalinas24 LGTM after rebase and tests pass :+1:

PR #1454, opened 2015-12-14
style(rxForm): Update color for rxMultiSelect + rxToggleSwitch

glynnis:
https://jira.rax.io/browse/FRMW-538

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

# rxToggleSwitch
### Contrast before
screen shot 2015-12-14 at 4 23 33 pm

### Contrast after
screen shot 2015-12-14 at 4 26 46 pm

### Off styles (before)
screen shot 2015-12-14 at 4 44 57 pm

### Off styles (after)
screen shot 2015-12-14 at 4 44 49 pm

### On styles (before)
screen shot 2015-12-14 at 4 46 01 pm

### On styles (after)
screen shot 2015-12-14 at 4 46 06 pm

### Disabled styles (before)
screen shot 2015-12-14 at 4 46 53 pm

### Disabled styles (after)
[Outdated screenshot.](https://cloud.githubusercontent.com/assets/1529366/11796430/e72a5be6-a281-11e5-8bb7-78026b894878.png)

# rxMultiSelect
### Unchecked item on hover (before)
screen shot 2015-12-14 at 3 51 37 pm

### Unchecked item on hover (after)
screen shot 2015-12-14 at 4 02 03 pm

### Checked item on hover (before)
screen shot 2015-12-14 at 3 51 44 pm

### Checked item on hover (after)
screen shot 2015-12-14 at 4 02 09 pm

### Checked item before and after (no change)
screen shot 2015-12-14 at 4 02 14 pm

### DIFF
--- a/src/components/rxToggleSwitch/rxToggleSwitch.less
+++ b/src/components/rxToggleSwitch/rxToggleSwitch.less
@@ -1,30 +1,15 @@
-/*
- * rxToggleSwitch
- */
-@textColor: @white;
-@offStateColor: #b3b3b3;
-@borderColor: darken(@offStateColor, 10%);
-@onStateColor: #00ac31;
-@knobColor: @white;
-@knobBorderColor: darken(@knobColor, 10%);
-@knobSize: 1.2em;
-@edgeSpacing: @knobSize * 0.9;
-@lineHeight: @knobSize * 1.25;
-@blurDist: @edgeSpacing;
-@blurSpread: (@edgeSpacing/3) * -1;
-
 .rx-toggle-switch {
   font-size: 0.8em;
-  line-height: @lineHeight;
+  line-height: @rxToggleSwitch-line-height;
 
   display: inline-block;
 
-  border: 1px @borderColor solid;
+  border: 1px @rxToggleSwitch-border-color solid;
   border-radius: 9px;
-  background-color: @offStateColor;
-  color: @textColor;
+  background-color: @rxToggleSwitch-off-state-color;
+  color: @rxToggleSwitch-text-color;
 
-  padding: 0 @edgeSpacing 0 @edgeSpacing;
+  padding: 0 @rxToggleSwitch-padding 0 @rxToggleSwitch-padding;
 
   width: 5em;
 
@@ -32,24 +17,25 @@
   position: relative;
   text-align: right;
 
-  .box-shadow(inset 1px 2px @blurDist @blurSpread rgba(0,0,0,0.50));
+  .box-shadow(inset 1px 2px @rxToggleSwitch-blur-dist @rxToggleSwitch-blur-spread rgba(0,0,0,0.50));
   .user-select(none);
   .box-sizing(border-box);
 
   &[disabled] {
-    background-color: desaturate(lighten(@offStateColor, 7%), 80%);
-    color: #666666;
+    background-color: @disabled-background-color;
+    color: @disabled-text-color;
     &.on {
-      background-color: desaturate(lighten(@onStateColor, 33%), 80%);
+      background-color: desaturate(lighten(@rxToggleSwitch-on-state-color, 40%), 30%);
+      color: darken(@disabled-text-color, 10%);
     }
     cursor: not-allowed;
     .knob {
-      background-color: #adadad;
-      border: 1px #999999 solid;
+      background-color: @disabled-background-color;
+      border: 1px @disabled-text-color solid;
     }
   }
   &.on {
-    background-color: @onStateColor;
+    background-color: @rxToggleSwitch-on-state-color;
     text-align: left;
     .knob {
       left: 100%;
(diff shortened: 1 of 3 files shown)


CITguy:
@glynnis, can you attach before/after screenshots of the "Toggle Switches" field from the _rxForm_ Demo? It's the only place at the moment that provides all possible current states of the switch. 

I'm loving the new rxMultiSelect colors :heart_eyes:

src/components/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -1,30 +1,15 @@
...
-  border: 1px @borderColor solid;
+  border: 1px @rxToggleSwitch-border-color solid;
   border-radius: 9px;
-  background-color: @offStateColor;
-  color: @textColor;
+  background-color: @rxToggleSwitch-off-state-color;
+  color: @rxToggleSwitch-text-color;
 
-  padding: 0 @edgeSpacing 0 @edgeSpacing;
+  padding: 0 @rxToggleSwitch-edge-spacing 0 @rxToggleSwitch-edge-spacing;
```
ssalinas24: @glynnis would it make more sense to include "padding" in the name, instead of "edge spacing"?

glynnis:
@CITguy good call. Here's before/after from rxForm page:

### Before
screen shot 2015-12-15 at 11 44 37 am

### After
screen shot 2015-12-15 at 11 44 51 am


@ssalinas24 @ElementE Do you think we should tweak the disabled colors so they're (a) more legible, and (b) not so different in appearance between on/off?

ElementE:
@glynnis - Agreed. They are really dark - maybe we try an opacity decrease for the bg? Legibility is really compromised here.

ssalinas24:
@glynnis We should reference the disabled button and use the same colors as much as possible.

src/components/rxToggleSwitch/rxToggleSwitch.less
glynnis: Good call. Updated :+1:

glynnis:
Revised disabled styles using same colors as disabled buttons:
screen shot 2015-12-15 at 1 59 52 pm

@ssalinas24 @ElementE Thoughts?

ElementE:
Much better! LGTM

100stacks:
@glynnis, I think the bottom two buttons (Setting 3 and Setting 4) are suppose to indicate `asynchronous operation is in progress` in https://github.com/rackerlabs/encore-ui/pull/1454#issuecomment-164878841.

In 1.0 color palette:
* Setting 3 is indicating an asynchronous OFF > ON operation with the _green-gray_ color transitioning to the ON state _green_ color.
* Setting 4 is indicating an asynchronous ON > OFF operation with the _darker gray_ transitioning to the OFF state _light grey_ color.  

I think the transition color helps the user know an asynchronous operation is in progress.  Is there another 2.0 color palette combo that could help convey this status, as the grey colors may not be as noticeable for the user?

glynnis:
@100stacks So what you brought up kind of made my head spin, but @ssalinas24 helped me figure it out :smile: 

Here's what we discussed/determined:
- The states that are represented as "Setting 1" etc. don't include a separate style for "async" states. They are actually just:
 - Setting 1 = ON
 - Setting 2 = OFF
 - Setting 3 = ON but disabled (i.e. can't be changed)
 - Setting 4 = OFF but disabled (i.e. can't be changed)

The only reason this is relevant for the "async" state is that while I call is being made, the switch goes to the new state (ON or OFF), but remains disabled (and gets disabled styles) until the call completes to prevent the user from interrupting the call by clicking again. 

This brings up an interesting point about how adding styles/a state for "an async call is being made" might be useful for UX, but that requires more dev work than just changing some styles. To complicate matters further, @ssalinas24 and I found some "in the wild" switches where spinners are being used to indicate that a call is being made:
![screen shot 2015-12-16 at 2 15 38 pm] (https://cloud.githubusercontent.com/assets/1529366/11852695/194c1d3e-a3ff-11e5-919b-551a08dafb18.png) (in Billing)

All very interesting to consider! And all outside the scope of a story to update colors :smile: So for now, I'm going to add a story on the UX board to look more carefully at "loading/pending/async" states for switches and better documenting them, and revert some of the color changes to the "disabled ON" state so that it still has some green to it.

Thanks for chiming in on this @100stacks! It produced some interesting conversation and revealed a need for better docs, even if we'll have to address that separately.

Code changes and color modification coming soon :+1:

100stacks:
@glynnis @ssalinas24, appreciate the clarification! :smile:  After talking to Steven and reading your https://github.com/rackerlabs/encore-ui/pull/1454#issuecomment-165232504 above, I agree with your comments concerning visual representation of "loading/pending/async" states.

I like the idea of the 'in the wild' spinner example as the animation provides visual feedback close to the action requested, although it will require the UX team to consider use cases across all Encore.

LGTM :+1:

glynnis:
Okay, updated! Here's the latest:
screen shot 2015-12-16 at 4 36 44 pm

CITguy:
LGTM

PR #1459, opened 2015-12-18
feat(rxDatePicker): Add rxDatePicker Molecule

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-95

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM

*screenshots to follow*

### DIFF
--- a/src/molecules/rxDatePicker/rxDatePicker.less
+++ b/src/molecules/rxDatePicker/rxDatePicker.less
@@ -1,0 +1,186 @@
+@import (reference) 'vars';
+
+rx-date-picker {
+  #vars.rxDatePicker();
+  box-sizing: border-box;
+  display: block;
+  margin: 0;
+  padding: 0;
+  width: @control-width;
+
+  .flexCentered() {
+    display: flex;
+    flex-flow: row wrap;
+    align-items: center;
+    justify-content: center;
+  }//.flexCentered()
+
+  &,
+  &.ng-valid {
+    .control {
+      .flexCentered();
+      background-color: @control-background-color;
+      border: 1px solid @control-border-color;
+      box-sizing: border-box;
+      color: @control-text-color;
+      cursor: @control-cursor;
+      flex-flow: row nowrap;
+      height: @control-height;
+      padding: @control-padding-x @control-padding-y;
+
+      .displayValue {
+        flex: 1;
+      }
+
+      .icon {
+        color: @control-icon-color;
+        flex: 0 0 @control-icon-width;
+        margin-left: @control-gutter;
+      }
+    }
+  }
+
+  &.ng-invalid.ng-dirty {
+    .control {
+      border: 2px solid @invalid-control-border-color;
+      // MUST differ with valid padding by 1px to account for thicker border
+      padding: (@control-padding-x - 1) (@control-padding-y - 1);
+
+      .icon {
+        color: @invalid-control-icon-color;
+      }
+    }
+  }
+
+  // Disabled styles should trump all other styles
+  // !important is required to achieve this
+  &[disabled] {
+    .control {
+      background-color: @disabled-control-background-color !important;
+      border: 1px solid @disabled-control-border-color !important;
+      color: @disabled-control-text-color !important;
+      cursor: @disabled-control-cursor !important;
+
+      .icon {
+        color: @disabled-control-icon-color !important;
+      }
+    }
+  }
+
+  .popup {
+    background-color: @popup-background-color;
+    border-color: @popup-border-color;
+    border-style: solid;
+    border-width: 5px 1px 1px; // top border handled by <nav>
+    border-top-color:  @popup-border-top-color;
+    margin: 0;
+    padding: 0;
+    position: absolute;
+    width: @popup-width;
+    z-index: @popup-z-index;
+
+    .circle {
+      .border-radius(100%);
+      .flexCentered();
+      border: 1px solid transparent;
+      box-sizing: border-box;
+      flex: 0 0 100%;
+      height: 100%;
+      width: 100%;
+    }//.circle
+
+    .day,
+    header h6 {
+      .flexCentered();
+      flex: 0 0 @day-width;
+      height: @day-width;
+    }
+
+    .day {
+      background-color: transparent;
+      color: @day-text-color;
+      cursor: pointer;
+
+      &.outOfMonth {
+        background-color: @out-of-month-day-background-color;
+        cursor: default;
+      }
+
+      .circle {
+        background-color: transparent;
+      }
+    }//.day
+
+    nav {
+      align-items: stretch;
+      border-bottom: 2px solid @popup-border-color;
+      color: @nav-text-color;
+      cursor: default;
+      display: flex;
+      flex-flow: row nowrap;
+      font-size: @nav-font-size;
+      font-weight: bold;
+      height: @nav-height;
+      justify-content: center;
+
+      span {
+        .flexCentered();
+        box-sizing: border-box;
+
+        &.arrow {
+          color: @nav-arrow-icon-color;
+          cursor: pointer;
+          flex: 0 0 @nav-arrow-width;
+
+          &:hover {
+            color: @nav-arrow-hover-icon-color;
+          }
+        }
+
+        &.currentMonth {
+          flex: 1;
+        }
+      }//span
+    }//nav
+
+    .calendar {
+      .flexCentered();
+      flex-flow: row wrap;
+
+      header {
+        .flexCentered();
+        background-color: @header-background-color;
+        border-bottom: 1px solid @header-border-bottom-color;
+        color: @header-text-color;
+        font-size: @header-font-size;
+        font-weight: bold;
+        height: @header-height;
+        flex-flow: row nowrap;
+        flex: 0 0 100%;
+      }//header
+
+      .day {
+        &.today {
+          .circle {
+            border-color: @today-border-color;
+          }
+        }//.today
+
+        &.selected {
+          .circle {
+            background-color: @selected-day-background-color;
+            border-color: transparent;
+            color: @selected-day-text-color;
+          }
+        }//.selected
+
+        &:not(.selected):hover {
+          .circle {
+            background-color: @day-hover-background-color;
+            color: @day-hover-text-color;
+          }
+        }//not(.selected):hover
+      }//.day
+    }//.calendar
+  }//.calendar-selector
+}//rx-date-picker
--- a/src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
+++ b/src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
@@ -1,0 +1,345 @@
+describe('rxDatePicker', function () {
+    var scope, isoScope, compile, rendered, renderDirective;
+    var clock;
+    var actual, expected;
+    var template;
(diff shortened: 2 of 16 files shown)


CITguy:
## Demo App Page
rxdatepicker-docs

## Date Picker in Action
rxdatepicker-empty-string

## API Docs
screen shot 2015-12-17 at 10 37 36 pm

src/styles/vars.less
```diff
@@ -91,6 +91,58 @@
 /* ======================================== *\
    COMPONENT VARIABLES
 \* ======================================== */
+#vars {
```
CITguy: This is an experiment in an attempt to namespace LESS variables. So far, the only downside to this is that the consuming LESS file can only import one namespace at a time, so you wouldn't be able to import both `rxDatePicker` and `rxForm` (if it were present). However, this doesn't seem to be an issue so far because the namespaced variables are only using local or global variables (colors and `@control-border-color`).

src/molecules/rxDatePicker/rxDatePicker.less
```diff
@@ -0,0 +1,193 @@
+@import (reference) 'vars';
+
+rx-date-picker {
+  #vars.rxDatePicker();
```
CITguy: By "extending" the `#vars.rxDatePicker` namespace, the `rx-date-picker` block inherits the LESS variables defined within the namespace.

megalord:
I'd like to  advocate that this be refactored to _fully_ use `ngModelCtrl`.  One of the key things missing here is the use of `$setViewValue()`, which means that users of this component cannot utilize the `$parsers` pipeline to perform validation.

You have set the `$modelValue` to be a string in the format `'YYYY-MM-DD'`, so I figure the `$viewValue` should probably be a moment object.  You can make your conversion to the string a `$parser`, and that way other `$parsers` can utilize the moment object's handy comparison methods.

src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
```diff
@@ -0,0 +1,345 @@
...
+            describe('selectDate()', function () {
+                describe('when passed Feb 14, 2016', function () {
+                    var valentines = moment('2016-02-14', isoFormat);
+
+                    it('should set selected to formatted value', function () {
+                        isoScope.selectDate(valentines);
+                        expect(isoScope.selected).to.eq('2016-02-14');
+                    });
+
+                    it('should set calendarVisible to fasle', function () {
```
100stacks: No biggie...misspelled `false`.

src/styles/vars.less
```diff
@@ -91,6 +91,57 @@
...
+  /*
+   * rxDatePicker
+   */
+  .rxDatePicker {
+    @calendar-background-color: @white;
+    @calendar-border-color-top: @blue-700;
+    @calendar-border-color: @gray-400;
+    @calendar-width: (7 * @calendar-day-size);
+
+    @calendar-nav-arrow-color: @gray-700;
```
ssalinas24: @CITguy  I believe the verdict on the header arrows from @glynnis and @ElementE is going to be to remove the `@blue-100` on hover. Keep this `@gray-700` as is, but the preferred hover state should just make the arrow color darker to `@gray-800` and just the white background remaining.

src/styles/vars.less
```diff
@@ -91,6 +91,57 @@
...
+
+    @calendar-day-color: @gray-800;
+    @calendar-day-hover-background-color: @blue-100;
+    @calendar-day-hover-color: @gray-900;
+    @calendar-day-out-month-background-color: @gray-100;
+    @calendar-day-selected-background-color: @blue-700;
+    @calendar-day-selected-color: @white;
+    @calendar-day-today-border-color: @gray-700;
+    @calendar-day-size: 30px;
+    @calendar-z-index: 100;
```
100stacks: @CITguy @ssalinas24, I recall we talked about reviewing `z-index` usage across `EncoureUI` at some point to possibly set reserved layers for different purposes.

Added a Tech Debt story (https://jira.rax.io/browse/FRMW-559) to capture the discussion.

src/molecules/rxDatePicker/rxDatePicker.less
```diff
@@ -0,0 +1,193 @@
...
+        .flexCentered();
+        box-sizing: border-box;
+
+        &.arrow {
+          color: @calendar-nav-arrow-color;
+          cursor: pointer;
+          flex: 0 0 @calendar-nav-arrow-width;
+
+          &:hover {
+            background-color: @blue-100;
```
ssalinas24: @CITguy Remove the `@blue-100` background color from hover and just keep the white background as is. Arrow will just change to `@gray-800` on hover with white background.

100stacks:
@CITguy excellent work on the first new `Molecule`! :tada: 

LGTM

CITguy:
@megalord, unfortunatly `$setViewValue` only [accepts a String as its argument](https://code.angularjs.org/1.2.29/docs/api/ng/type/ngModel.NgModelController#$setViewValue), so we can't save the moment object as desired. However, using the formatters and parsers may clean up the code a bit. I'll do some digging.

Droogans:
I didn't find any information about the correct way to close the datepicker. Currently it only supports when clicking the calendar icon, I think it may feel more familiar if it closed when clicking away from the calendar element as well.

@glynnis @ssalinas24 what are your thoughts?

megalord:
That is strange that the docs would say that, because I see no reason in the [source](https://github.com/angular/angular.js/blob/v1.2.29/src/ng/directive/input.js#L1204) why it would be limited to a string.  Ultimately, I just want the ability to run validations by using a directive that hooks into `ngModelCtrl` rather than adding a watcher to the `ngModel` expression.

src/styles/vars.less
ssalinas24: @CITguy  I think the var name would be more descriptive if we call it `@calendar-nav-arrow-icon-color`. Glynnis and I have been trying to keep "text", "icon", and "background" words in the var names to be more clear.

src/styles/vars.less
ssalinas24: Awesome. Thanks @100stacks. I remember that conversation as well.

CITguy:
@megalord, after looking into the formatters/parsers route, it may lead down a rabbit hole that we don't have time to explore at present. I've created [FRMW-560](https://jira.rax.io/browse/FRMW-560) to capture revisiting the idea of implementing full formatters/parsers. I understand your concern, but for this PR, we simply do not have the time to fully implement formatters/parsers as desired.

src/styles/vars.less
```diff
@@ -91,6 +91,57 @@
...
+
+    @control-width: 150px;
+    @control-background: @white;
+    @control-background-disabled: @gray-300; // @disabled-background-color
```
ssalinas24: @CITguy  I think we would flip the wording here, right? It should be `@control-disabled-background-color`.

src/styles/vars.less
CITguy: will do, adding `@calendar-nav-arrow-color-hover: @gray-800;`

src/molecules/rxDatePicker/rxDatePicker.less
CITguy: will do

src/styles/vars.less
```diff
@@ -91,6 +91,57 @@
...
+    @calendar-header-color: @gray-800;
+    @calendar-header-font-size: 10px;
+    @calendar-header-height: 20px;
+
+    @calendar-day-color: @gray-800;
+    @calendar-day-hover-background-color: @blue-100;
+    @calendar-day-hover-color: @gray-900;
+    @calendar-day-out-month-background-color: @gray-100;
+    @calendar-day-selected-background-color: @blue-700;
+    @calendar-day-selected-color: @white;
```
ssalinas24: @CITguy Should probably be `@calendar-day-selected-text-color`

src/styles/vars.less
CITguy: I try to keep the name of the css property in the variable name. For the `@control-background*` vars, it corresponds to the `background` CSS property. However, I should be using the `background-color` CSS property, so I'll update these to be `@control-background-color` and `@control-background-color-disabled`.

src/styles/vars.less
CITguy: will do

src/styles/vars.less
glynnis: :clap:

glynnis:
@Droogans Agree, I think usually with date pickers and similar UI elements, the user expects it to close or "finish" with the selection process (if they've selected something) when they click away.

CITguy:
@glynnis, can the "click away to close" functionality wait for a future PR, or is it a requirement for this molecule?

glynnis:
@CITguy I'm fine with an additional PR/iteration, preferably in the next sprint where folks are around to work on it :)

ssalinas24:
@CITguy Just update the screenshot or animation, and comment out the old stuff. Then I should be able to LGTM. Everything else looks good I think.

CITguy:
@ssalinas24, gif updated and I added a static screenshot of the popup.

ssalinas24:
LGTM

Droogans:
This has test work to do but that's coming likely next week (to be merged and released January 2016).

src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
```diff
@@ -0,0 +1,345 @@
...
+        module('encore.ui.molecules');
+        module('templates/rxDatePicker.html');
+
+        inject(function ($rootScope, $compile) {
+            scope = $rootScope.$new();
+            compile = $compile;
+        });
+
+        // "today" is December 25, 2015
+        clock = sinon.useFakeTimers(new Date(2015, 11, 25).getTime());
```
parlarjb: The `Date` is being passed November, but the comment says December

parlarjb: Actually, is `clock` used anywhere?

src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
```diff
@@ -0,0 +1,345 @@
...
+                    actual = _.some(isoScope.calendarDays, function (day) {
+                        return day.format(isoFormat) === '2016-01-02';
+                    });
+                    expect(actual).to.eq(true);
+                });
+
+                it('should not include Jan 3, 2016', function () {
+                    actual = _.some(isoScope.calendarDays, function (day) {
+                        return day.format(isoFormat) === '2016-01-03';
+                    });
```
parlarjb: How about turning this `_.some` pattern into a helper function in this file, rather than all the copy/paste

parlarjb:
This stuff looks pretty awesome

parlarjb:
Were you planning on including an exercise.js file?

CITguy:
@parlarjb `*.exercise.js` will be added at some point, yes

src/molecules/rxDatePicker/scripts/rxDatePicker.spec.js
CITguy: Unfortunately, that's not a typo. Vanilla `Date()` takes a 0-based month.

Droogans:
@CITguy stop force pushing against open PRs.

parlarjb:
LGTM

PR #1468, opened 2015-12-22
docs(layout): Nav points to flexbox grid instead of Pure

glynnis:
https://jira.rax.io/browse/FRMW-534

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

# Before
screen shot 2015-12-22 at 2 28 41 pm
screen shot 2015-12-22 at 2 28 49 pm

# After
screen shot 2015-12-22 at 2 22 33 pm
screen shot 2015-12-22 at 2 23 36 pm


demo/examples/layout.grid.html
```diff
@@ -0,0 +1,506 @@
+<div class="layout-demo">
```
glynnis: I copied the flexbox demo from http://rackerlabs.github.io/encore-ui/#/components/layout to make an rx-example for use on the grids page. However, this brings into question what should happen to the component page -- do we need to keep both?

demo/templates/layout/grid.html
```diff
@@ -0,0 +1,28 @@
...
+    The layout system is based upon element attributes rather than CSS classes. Attributes provide an easy way to
+    set a value (eg <code>layout="row"</code>) and help separate concerns:
+    attributes define layout, and classes define styling.
+  </p>
+
+  <p>
+    We have
+    <a href="#/layout/grid-1-0">a reference to the old Pure CSS grid</a>
+    for your convenience, but note that the Pure CSS grid is slated for
+    deprecation and shouldn't be used. Any pages using it should be updated.
```
glynnis: This method of referring to deprecated documentation pages via text link has precedent on the color palette page: http://rackerlabs.github.io/encore-ui/#/styles/color

demo/examples/layout.grid.html
ssalinas24: @glynnis So what is the final outcome in your branch so far? There's a Layout Page, Flex Grid Page, and Pure Grid Page? or Just Flex Grid Page and Pure Grid page?

glynnis: There's:
1) Flexbox Grid page, accessible in the nav under "Layout" section
2) Pure Grids page, accessible only via link on the Flexbox Grid page
3) A flexbox grid component page, under components, called "layout" (as has always been the case)

ssalinas24:
LGTM
@glynnis Visually I think everything is fine. I think the Flex grid page for sure has to go under the Layout tab/section. I really don't think anything should exist under the components section for it. However, I will let the devs chime in on it.

demo/examples/layout.grid.html
CITguy: Please correct all indentation to use 2 spaces per indentation.

glynnis: Fixed :+1:

CITguy: @glynnis: Considering that the layout "component" is purely CSS, I'd be inclined to move the source files to `src/styles` and keep the UI and examples in `demo/`.  However, that would definitely be out of scope of this PR because it may require some tweaks to the build tasks.

demo/examples/layout.grid.html
```diff
@@ -0,0 +1,505 @@
...
+      [flex]
+    </div>
+  </div>
+
+  <p>Add the <code>offset</code> attribute to a layout child to set its offset percentage within the layout.
+  Values must be multiples of <code>5</code>, or <code>33</code>, <code>34</code>, <code>66</code>,
+  <code>67</code>.</p>
+
+  <table>
+  <tr>
```
CITguy: `<tr>` tags should be indented one more level. Please correct.

demo/examples/layout.grid.html
glynnis: @CITguy Agreed that we should probably tackle that separately. Could you create the appropriate stories for the Framework backlog (I'm not sure I know enough re: build tasks to capture accurately). So we'll leave nav structure as is for this PR and refine later :+1:

demo/examples/layout.grid.html
glynnis: Der. Thank you. Corrected.

demo/examples/layout.grid.html
CITguy: [FRMW-563](https://jira.rax.io/browse/FRMW-563) created to capture discussions.

CITguy:
LGTM

parlarjb:
LGTM

PR #1469, opened 2015-12-22
style(layout1): Use flexbox grid for detail layout

glynnis:
https://jira.rax.io/browse/FRMW-533

- [ ] Dev LGTM
- [ ] Dev LGTM
- [ ] Design LGTM

# Before
screen shot 2015-12-22 at 3 08 26 pm

# After
screen shot 2015-12-22 at 3 08 14 pm

### DIFF
--- a/demo/templates/layout/page/detail-page.html
+++ b/demo/templates/layout/page/detail-page.html
@@ -13,7 +13,8 @@
     </ul>
   </div>
 
-  <rx-metadata class="clear">
+  <br>
+  <rx-metadata>
     <section>
       <rx-meta label="Name">Lorem ipsum dolor sit amet</rx-meta>
       <rx-meta label="Region">ORD</rx-meta>
@@ -32,8 +33,8 @@
   </rx-metadata>
 
   <div class="data-section">
-    <h3 class="clear-left">Header</h3>
-    <h3>Sub-header</h3>
+    <h3>Header</h3>
+    <h4>Sub-header</h4>
 
     <table class="table-striped">
       <thead>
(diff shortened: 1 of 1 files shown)


100stacks:
@glynnis looks cleaner with the update.  

LGTM :+1:

ElementE:
Finally we have fixed this! LGTM

CITguy:
On my macbook screen, I see the following.

### Production
detail-layout-smaller-resolution-old

### Pull Request
detail-layout-smaller-resolution-new
* The page actions `float` property is interfering with the layout of the metadata.

Have we figured out a solution to the page actions yet?

CITguy:
I recall there might be a long-term UX solution/replacement for page actions, but I've not heard anything about that lately. Is this still in the works?

glynnis:
@CITguy I believe what you're seeing on your macbook is indeed correct -- the metadata is wrapping into two columns when it doesn't have enough space to be three. My "after" screen shot was taken on a cinema display, where there's enough space for the metadata to be in three columns.

Figuring out page actions is still in the works. There's a story for it on the UX board here: https://jira.rax.io/browse/UXENCORE-448. I know there's work that's been done that isn't yet represented on the story, and I'll be working to sync designers to get the UX board more up to date and better representing what progress has been made, who's working on what, etc.

CITguy:
@glynnis, so what I'm seeing on my screen is acceptable?

glynnis:
@CITguy Yep! Everything's working as expected. I took a closer look in the inspector to figure out exactly when the wrap his happening for metadata:

screen shot 2016-01-06 at 11 25 57 am
screen shot 2016-01-06 at 11 27 04 am

It looks like at ~1450px the metadata doesn't have as much width as it needs for three columns + page actions, so it wraps.

PR #1475, opened 2015-12-28
refactor(commacleanup): FRMW-553 remove unnecessary commas

jgonz0010:
Got rid of the extra commas associated with arrays and the commas at the end of the arrays. Check out FRMW-553 for more info. Link Below.

https://jira.rax.io/browse/FRMW-553

- [x] Dev LGTM
- [x] Dev LGTM

### DIFF
--- a/src/components/rxApp/scripts/rxApp.spec.js
+++ b/src/components/rxApp/scripts/rxApp.spec.js
@@ -270,7 +270,7 @@ describe('encore.ui.rxApp', function () {
                 children: [
                     {
                         'href': '/app',
-                        'linkText': 'App',
+                        'linkText': 'App'
                     }
                 ]
             }];
--- a/src/components/rxApp/scripts/rxAppNavItem.spec.js
+++ b/src/components/rxApp/scripts/rxAppNavItem.spec.js
@@ -80,7 +80,7 @@ describe('encore.ui.rxApp', function () {
                 access: {
                     token:
                         {
-                            id: 'someid',
+                            id: 'someid'
                         },
                         user: {
                             id: 'joe.customer',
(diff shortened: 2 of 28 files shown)


Droogans:
I'm surprised @parlarjb was OK with this being a story at all in the first place. He really pushes for extraneous trailing commas, and I agree with him (although I've yet to make a solid habit out of it yet).

jgonz0010:
To be fair he has probably not seen it yet. You are not the first to mention it though. With that being said, I don't think putting them in or keeping them out is an issue anymore. Given that even IE has made changes so that commas at the end of the arrays are ignored. :-)

100stacks:
I'm for removing the trailing comma `,` -- if we keep them as a matter of preference, we should annotate this in our [JavaScript Styleguide](http://rackerlabs.github.io/encore-ui/#/guides/js-styleguide).
 
LGTM :+1:

CITguy:
Though we don't support IE, I'm still for removing the trailing commas. :+1:

CITguy:
LGTM

@parlarjb, do you have anything you'd like to add?

parlarjb:
I'll let Raymond Hettinger (one of the smartest people I've ever seen speak) show why trailing commas are the preferred way in the Python community 

http://stackoverflow.com/a/11597911

(Plus it makes it easier for people to add to the list)

CITguy:
Ok, @parlarjb. You've convinced me. Given that trailing commas are not breaking ES5 spec and that we don't officially support IE, I'm going to close this PR.

CITguy:
Sorry @jgonz0010. :(

jgonz0010:
it's all good! I'll remove the trailing commas rule from my analysis tool.

PR #1497, opened 2016-01-15
fix(npm): revert bandaid for dependency regression

CITguy:
* undo pinning to `vinyl-fs 2.2.1` (project has released 2.3.1 to correct regression)


CITguy:
Dependency regression fixed! ^_^

PR #1503, opened 2016-01-19
fix(rxSelect): set up local stacking context

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-637

By setting a `z-index` of the wrapper, the `z-index` values of the children do not escape the parent context, so they don't interfere with other elements on the page.

### LGTMs

- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/src/components/rxSelect/rxSelect.less
+++ b/src/components/rxSelect/rxSelect.less
@@ -2,6 +2,7 @@
  * rxSelect
  */
 .rxSelect {
+  z-index: 0; // creates local z-index stacking context
   .box-sizing(border-box);
   display: block;
   position: relative;


100stacks:
@CITguy great discovery! :smile: 

LGTM

CITguy:
### BEFORE
screen shot 2016-01-19 at 11 21 12 am

### AFTER
screen shot 2016-01-19 at 11 20 53 am

bry4n:
:+1:

glynnis:
Design LGTM.

@CITguy Do we know if the same issue exists with other rxForm components? Wondering if you've already verified or if maybe we need a spike to double-check others for similar problems.

CITguy:
@glynnis, haven't seen the issue with other components, but it might not be a bad idea to do an audit of z-index values for custom elements.

PR #1505, opened 2016-01-19
refactor(rxDebug): add rxDebug for ui testing purposes

CITguy:
* Add `rxDebug` directive to hide page content unless the `?debug` URL query parameter is present.
* This does not affect distributed code.
* Useful to hide content when midway tests are _not_ running to reduce clutter in the documentation.

### Example

```html
<p>I'll always show up!</p>

<rx-debug>
  <p>I'll only show up if "?debug" is present in url!</p>
</rx-debug>
```

### DIFF
--- a/demo/scripts/rxDebug.js
+++ b/demo/scripts/rxDebug.js
@@ -1,0 +1,11 @@
+angular.module('demoApp')
+.directive('rxDebug', function ($routeParams) {
+    return {
+        restrict: 'E',
+        transclude: true,
+        template: '<div ng-if="isDebugging" ng-transclude></div>',
+        link: function (scope) {
+            scope.isDebugging = $routeParams.debug === 'true';
+        }
+    };
+});
--- a/demo/index.html
+++ b/demo/index.html
@@ -62,6 +62,7 @@
     <script src="scripts/showGuideController.js"></script>
     <!-- directives -->
     <script src="scripts/alwaysInvalid.js"></script>
+    <script src="scripts/rxDebug.js"></script>
     <script src="scripts/rxPrism.js"></script>
     <script src="scripts/rxStyleguide.js"></script>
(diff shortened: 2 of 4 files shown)


Droogans:
I'll see about adding this to an existing end to end test. I think it'll be a big success.

demo/scripts/rxDebug.js
```diff
@@ -0,0 +1,11 @@
+angular.module('demoApp')
+.directive('rxDebug', function ($routeParams) {
+    return {
+        restrict: 'E',
+        transclude: true,
+        template: '<div ng-if="isDebugging" ng-transclude></div>',
+        link: function (scope) {
+            scope.isDebugging = !_.isUndefined($routeParams.debug);
```
Droogans: I wonder if it's best to strictly check for `true`. This is very permissive as is.

```js
scope.isDebugging = $routeParams.debug === true;
```

(or maybe that should be `'true'`, but you get the idea).

CITguy: The value of the param shouldn't matter. The presence of the param is all that's needed.

Droogans:
@CITguy I'd really prefer it if you could at least help me get some prep work done (possibly with @anur2213 and @glynnis) for outlining all demo sections that could benefit from using `rx-debug`.

As it stands, all midways should continue passing normally, no matter where you put them.

demo/scripts/rxDebug.js
Droogans: I just saw this. I already pushed up an explicit "must equal 'true'" commit. Sorry!

I can pull it, but if you don't care either way, I'd rather not undo that work and force push this up again.

100stacks:
LGTM

CITguy:
LGTM

PR #1514, opened 2016-01-26
style(vars): FRMW-594 Update Link Global Vars

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-594

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -60,8 +60,8 @@
 @app-disabled-background-color: @gray-300;
 @app-disabled-text-color: @gray-600;
 @app-disabled-icon-color: @gray-600;
-@link-color: @blue-700;
-@link-hover-color: saturate(@link-color, 15%);
+@app-link-text-color: @blue-700;
+@app-link-hover-text-color: saturate(@app-link-text-color, 15%);
 
 // # App Global Form Vars
 @app-label-font-size:12px;
@@ -429,6 +429,8 @@
 // deprecated
 @disabled-background-color: @app-disabled-background-color;
 @disabled-text-color: @app-disabled-text-color;
+@link-color: @app-link-text-color;
+@link-hover-color: saturate(@link-color, 15%);
 @linkColor: @link-color;
 @linkColorHover: @link-hover-color;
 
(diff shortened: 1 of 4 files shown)


src/styles/vars.less
```diff
@@ -427,6 +427,8 @@
 
 // # Text Colors
 // deprecated
+@link-color: @app-link-text-color;
+@link-hover-color: saturate(@app-link-hover-text-color, 15%);
```
glynnis: @ssalinas24 Your definitions here use the newer vars, `@app-link-hover-text-color` and `@app-link-text-color`. Instead, please preserve the original calculation/var and comment to indicate what the new var is called, as we've done elsewhere (see https://github.com/rackerlabs/encore-ui/blob/master/src/styles/vars.less#L472).

glynnis: Ah, nevermind -- I looked at the JIRA story and see what you did here :) Deleting this and my prior comment to give an LGTM

glynnis:
Design LGTM

src/styles/vars.less
ssalinas24: @glynnis actually I think your instinct was correct. I actually thought i did what you were trying to suggest, but i didn't. I should probably be doing `@link-hover-color: saturate(@link-color, 15%);`

glynnis: Hah, ok. I saw in the story on JIRA that you'd specified how to deprecate so I thought there was a good reason for that (it being a more recent variable that you were deprecating). So confusing.

src/styles/vars.less
```diff
@@ -427,6 +427,8 @@
 
 // # Text Colors
 // deprecated
+@link-color: @app-link-text-color;
+@link-hover-color: saturate(@link-color, 15%);
```
ssalinas24: @glynnis here is the corrected line.

CITguy:
LGTM

100stacks:
LGTM

CITguy:
@ssalinas24, please rebase

ssalinas24:
@CITguy alright I rebased and pushed back up again.

PR #1516, opened 2016-01-26
style(vars): FRMW-606 Correct rxButton Component Vars

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-606

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/src/components/rxButton/rxButton.less
+++ b/src/components/rxButton/rxButton.less
@@ -65,14 +65,15 @@
   }
 
   &.finish {
-    background: #fff;
-    color: @rxButton-default-background-color;
-    .box-shadow(inset 0 0 0 2px @rxButton-default-background-color);
+    background: @rxButton-finish-background-color;
+    color: @rxButton-finish-text-color;
+    // Inset line is created with .box-shadow instead of border property, because it avoids changing button height.
+    .box-shadow(inset 0 0 0 2px @rxButton-finish-outline-color);
 
     &:hover,
     &:focus {
-      color: @rxButton-default-hover-background-color;
-      .box-shadow(inset 0 0 0 2px @rxButton-default-hover-background-color);
+      color: @rxButton-finish-hover-text-color;
+      .box-shadow(inset 0 0 0 2px @rxButton-finish-hover-outline-color);
     }
   }
 
--- a/src/styles/vars.less
+++ b/src/styles/vars.less
@@ -204,6 +204,11 @@
 @rxButton-disabled-text-color: @gray-600;
 @rxButton-disabled-background-color: @gray-300;
 @rxButton-spinner-background-color: @white;
+@rxButton-finish-text-color: @blue-700;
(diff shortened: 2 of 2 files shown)


src/styles/vars.less
```diff
@@ -204,6 +204,11 @@
 @rxButton-disabled-text-color: @gray-600;
 @rxButton-disabled-background-color: @gray-300;
 @rxButton-spinner-background-color: @white;
+@rxButton-finish-text-color: @blue-700;
+@rxButton-finish-hover-text-color: lighten(@rxButton-finish-text-color, 10%);
+@rxButton-finish-background-color: @white;
+@rxButton-finish-outline-color: @blue-700;
+@rxButton-finish-hover-outline-color: lighten(@rxButton-finish-outline-color, 10%);
```
ssalinas24: @CITguy and @glynnis the rxButton had variables for default, positive, negative, and cancel. However, it never got specific vars for the finish state in the multi-modal stuff. I have created the new vars above and implemented them into the rxButton.less. Let me know if you agree with these names, or if you can think of anything more appropriate. FYI, drop shadow is used for the outline, but i decided to call it outline because its not really a visual shadow or border property. What do you think?

CITguy:
LGTM

src/styles/vars.less
glynnis: These var names look good to me :+1:

Interesting point about naming for the outline color. It seems like the CSS could be refactored to just utilize a regular border? I'm wondering why it's a box shadow in the first place. Did you play around with that at all @ssalinas24 ?

ssalinas24: @glynnis I can take a look at what the border property would do versus the current shadow property.

src/components/rxButton/rxButton.less
```diff
@@ -65,14 +65,15 @@
   }
 
   &.finish {
-    background: #fff;
-    color: @rxButton-default-background-color;
-    .box-shadow(inset 0 0 0 2px @rxButton-default-background-color);
+    background: @rxButton-finish-background-color;
+    color: @rxButton-finish-text-color;
+    // Inset line is created with .box-shadow instead of border property, because it avoids changing button height.
```
ssalinas24: @glynnis I included a comment as you suggested, but only one instance.

src/styles/vars.less
```diff
@@ -204,6 +204,11 @@
 @rxButton-disabled-text-color: @gray-600;
 @rxButton-disabled-background-color: @gray-300;
 @rxButton-spinner-background-color: @white;
+@rxButton-finish-text-color: @blue-700;
+@rxButton-finish-hover-text-color: lighten(@rxButton-finish-text-color, 10%);
+@rxButton-finish-background-color: @white;
+@rxButton-finish-outline-color: @blue-700; // Color of Inset Line For Button
```
ssalinas24: @glynnis I included a comment here in vars file as well.

glynnis:
Design LGTM

100stacks:
LGTM

PR #1517, opened 2016-01-28
fix(grunt): fix deploy tasks

CITguy:
rxPageObjects deploy tasks were deploying packed versions of the deployed files, resulting in a tarball (*.tgz) file within the distributed directory. This process also packed previous tarballs already in the directory, so each subsequent tarball contained all of the previous tarballs, causing exponential file size increases to the point of failure to publish because the payload was too large.

* no tasks made use of these tarballs (grunt tasks removed)
* in my effort to troubleshoot, I accidentally removed the `package.json` (readded)

@Droogans, I'd like your seal of approval before I merge this PR.

### LGTMs

- [x] Droogans LGTM
- [x] parlarjb LGTM

### DIFF
--- a/grunt-tasks/options/shell.js
+++ b/grunt-tasks/options/shell.js
@@ -2,16 +2,6 @@
 
 module.exports = function (grunt) {
     return {
-        rxPageObjects: {
-            command: 'npm pack',
-            options: {
-                stdout: true,
-                execOptions: {
-                    cwd: 'utils/rx-page-objects'
-                }
-            }
-        },
-
         npmPublish: {
             command: 'npm publish ./rx-page-objects',
             options: {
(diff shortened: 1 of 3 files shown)


grunt-tasks/options/shell.js
```diff
@@ -2,16 +2,6 @@
...
-        rxPageObjects: {
-            command: 'npm pack',
-            options: {
-                stdout: true,
-                execOptions: {
-                    cwd: 'utils/rx-page-objects'
-                }
-            }
-        },
-
```
CITguy: This task would pack the `utils/rx-page-objects` contents into a tarball *within* the `utils/rx-page-objects` directory.
Every additional release would include the tarball from the previous release, so the directory sent to npm for publishing got exponentially larger. No grunt tasks make use of these tarballs, so they were removed and this task removed from the build process.

CITguy:
@parlarjb, I'd like your input, as well, please.

parlarjb:
LGTM

grunt-tasks/options/shell.js
CITguy: The latest successful release resulted in a node module that was somewhere in the ballpark of 60MB in size.

Droogans:
Sorry, everyone!

PR #1520, opened 2016-01-29
style(vars): FRMW-603 Update Preprod Warning Var Values

ssalinas24:
JIRA: https://jira.rax.io/browse/FRMW-603

- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

### DIFF
--- a/src/components/rxApp/rxApp.less
+++ b/src/components/rxApp/rxApp.less
@@ -156,8 +156,8 @@
         &:hover,
         &:focus {
           background: @preprod-focus-background-color;
-          border-top: 1px solid @preprod-border-bottom-color;
-          border-bottom: 1px solid @preprod-border-bottom-color;
+          border-top: 1px solid @preprod-focus-border-top-color;
+          border-bottom: 1px solid @preprod-focus-border-bottom-color;
         }
       }
 
--- a/src/components/rxApp/rxApp.less
+++ b/src/components/rxApp/rxApp.less
@@ -143,30 +143,30 @@
 
   &.warning-bar.preprod {
     .rx-app-menu {
-      background: @preprodBackground;
+      background: @preprod-background-color;
     }
     .rx-app-nav-item {
       &.active:not(.has-children) > .item-link {
-        border-top-color: @preprodBorderTop;
+        border-top-color: @preprod-border-top-color;
       }
       .item-link {
-        background: @preprodBackground;
-        border-top: 1px solid @preprodBorderTop;
-        border-bottom: 1px solid @preprodBorderBottom;
+        background: @preprod-background-color;
+        border-top: 1px solid @preprod-border-top-color;
+        border-bottom: 1px solid @preprod-border-bottom-color;
         &:hover,
         &:focus {
-          background: @preprodBackgroundFocus;
-          border-top: 1px solid @preprodBorderBottom;
-          border-bottom: 1px solid @preprodBorderBottom;
+          background: @preprod-focus-background-color;
+          border-top: 1px solid @preprod-border-bottom-color;
+          border-bottom: 1px solid @preprod-border-bottom-color;
         }
       }
 
       // active nav, w/children (double arrow up, gray bg)
       &.active.has-children > .item-link {
-        border-top-color: @preprodBorderTop;
+        border-top-color: @preprod-border-top-color;
         // used to show BG of arrow (because :after is rotated)
         &:before {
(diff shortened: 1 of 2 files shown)


ssalinas24:
@glynnis and @ElementE I have updated the PreProd menu to have 2.0 colors. However, the hover state has issues with color contrast. Do we need to come up with another red, or will this not be an issue?

01_using-version-2-colors

One thing i tried was saturating the `@red-700` for the hover color. Would this be a possibility?
02_saturating-red700-on-hover

src/components/rxApp/rxApp.less
```diff
@@ -143,30 +143,30 @@
...
+        background: @preprod-background-color;
+        border-top: 1px solid @preprod-border-top-color;
+        border-bottom: 1px solid @preprod-border-bottom-color;
         &:hover,
         &:focus {
-          background: @preprodBackgroundFocus;
-          border-top: 1px solid @preprodBorderBottom;
-          border-bottom: 1px solid @preprodBorderBottom;
+          background: @preprod-focus-background-color;
+          border-top: 1px solid @preprod-border-bottom-color;
```
ssalinas24: @glynnis this border bottom var was repeated in this property because it was the same color. However, applying the new var name bottom on the actual border-top property breaks our rule. If i create a new var would you agree with `@preprod-hover-border-top-color`?

ssalinas24:
@ElementE  Here are updated screenshots from recent commit.

## Before
![00_preprod_before]
(https://cloud.githubusercontent.com/assets/14296817/12731819/4ab99234-c8f8-11e5-9a0d-2e910ae8b9fc.png)

## Hover State
I couldn't get saturate or desaturate to work like previously discussed. It was too similar to the active state. So i have tried this desaturate for hover. Do you think this works, or do you think its hard to tell what's selected?

03_hover

## Active Drop Down
04_active

CITguy:
@ssalinas24, which link is being hovered? I cannot tell which link is normal and which is hovered.

ssalinas24:
@CITguy Well that confirms my concern.. LOL. "All Modules" was the item hovered upon. I will have to figure out a way to make it more obvious.

ssalinas24:
@CITguy and @ElementE it seems like this is the best solution. Vars have been updated, new vars have been added to follow naming convention, and 2.0 color vars have been used. No hexadecimals were used, and I was able to use LESS saturate and darken properties. Contrast is also still good on all states.

These are screenshots that I took on my local. I replicated the environment by adding the warning preprod classes in the browser inspector. _**Ryan please let me know if I should this on my local in some other way if preferred.**_

## PreProd Before
00_preprod_before

## Local/After: Active Item (Orange/Yellow)
01_active-item

## Local/After: Hovered Item ("All Modules" In Hover State)
02_hover-item

## Local/After: Active Menu
03_active-menu

## Local/After: Active Menu Item
04_active-menu-item

src/components/rxApp/rxApp.less
```diff
@@ -484,7 +492,7 @@
     text-shadow: 1px 1px 0px rgba(26, 26, 26, 0.2);
 
     .rx-feedback .modal-link {
-      background: url(images/icon-feedback.png) no-repeat 20px calc(~"50% + 1px");
+      background: url("images/icon-feedback.png") no-repeat 20px calc(~"50% + 1px");
```
ssalinas24: @CITguy We had some missing quotations and I noticed that my code hinting was mentioning that everything after this point was messed up and possibly not working properly.

src/components/rxApp/rxApp.less
```diff
@@ -143,30 +143,38 @@
...
-          background: @preprodBackgroundFocus;
-          border-top: 1px solid @preprodBorderBottom;
-          border-bottom: 1px solid @preprodBorderBottom;
+          background: @preprod-focus-background-color;
+          border-top: 1px solid @preprod-focus-border-top-color;
+          border-bottom: 1px solid @preprod-focus-border-bottom-color;
+        }
+      }
+
+      // active nav, no children (orange/yellow bg, arrow right)
```
ssalinas24: This section was previously around line 149, didn't seem to be applying, and also seemed to be in incorrect order. So I moved to be placed in this new position of the LESS file.

glynnis:
LGTM

CITguy:
LGTM

parlarjb:
LGTM

PR #1525, opened 2016-02-04
WIP: feat(version): Include EncoreUI version in Feedback

CITguy:
Adding a `ENCORE_UI_VERSION` constant for use with rxFeedback

### LGTMS
- [ ] Dev LGTM
- [ ] Dev LGTM

### DIFF
--- a/src/quarks/rxFeedbackSvc/scripts/rxFeedbackSvc.js
+++ b/src/quarks/rxFeedbackSvc/scripts/rxFeedbackSvc.js
@@ -6,7 +6,7 @@ angular.module('encore.ui.quarks')
  * `rxFeedbackSvc` service supports `rxFeedback` directive functionality.  A `custom endpoint` may be set to override
  * the `default` endpoint.
  */
-.factory('rxFeedbackSvc', function ($resource, feedbackApi, $location, $window) {
+.factory('rxFeedbackSvc', function ($resource, feedbackApi, $location, $window, ENCORE_UI_VERSION) {
     var container = {
         api: undefined,
         email: 'encoreui@lists.rackspace.com'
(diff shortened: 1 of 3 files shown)


parlarjb:
Nice idea, LGTM

src/quarks/rxFeedbackSvc/scripts/rxFeedbackSvc.js
```diff
@@ -23,6 +23,7 @@ angular.module('encore.ui.quarks')
         var subject = 'Encore Feedback: ' + feedback.type.label;
         var body = [
             'Current Page: ' + $location.absUrl(),
+            'EncoreUI Version: ' + ENCORE_UI_VERSION,
```
julianbonilla: The EncoreUI version will only be reported if the Feedback Service fails.  We should also include it here: https://github.com/rackerlabs/encore-ui/blob/master/src/components/rxFeedback/scripts/rxFeedback.js#L65-L76.

Will require an update to the API side also.

parlarjb:
Bah, can't believe I didn't remember that this would need an API update. Good catch @julianbonilla 

@CITguy I'm removing the LGTM for now, because use of the feedback service is how (theoretically) all feedback notifications now take place.

CITguy:
Yeah, there's some other changes that need to take place as well w.r.t. grunt tasks because current implementation doesn't work for karma/unit tests.

src/quarks/rxFeedbackSvc/scripts/rxFeedbackSvc.js
CITguy: Thanks. I'll take a look.

CITguy:
I'm going to close this PR for now. It seems like there's more work than I had initially thought. I've created the following JIRA story to capture the work. https://jira.rax.io/browse/FRMW-682

PR #1541, opened 2016-02-11
docs(rxAge): Update return type

Droogans:
### LGTMs
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/quarks/rxAge/scripts/rxAge.page.js
+++ b/src/quarks/rxAge/scripts/rxAge.page.js
@@ -1,15 +1,17 @@
 /**
- * @exports encore.rxAge
+ * @namespace
  */
 exports.rxAge = {
 
     /**
      * @function
+     * @description A moment date object representing the point in time the `rxAgeString` refers to. This value does
+     * not come back as a promise, but is instead a direct value.
      * @param {String} rxAgeString - An age string, representing the time that has passed since a certain datetime.
-     * @returns {Date} A moment date object representing the point in time the `rxAgeString` refers to.
+     * @returns {moment}
      * @example
      * // given the current datetime is January 1st, 1970 at noon UTC.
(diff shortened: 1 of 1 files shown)


CITguy:
LGTM

src/quarks/rxAge/scripts/rxAge.page.js
```diff
@@ -36,7 +38,7 @@ exports.rxAge = {
         });
 
         var elapsed = _.zipObject(ageParts);
-        return moment().utc().subtract(moment.duration(elapsed));
+        return moment().utc().subtract(elapsed);
```
jj4th: The moment.js documentation does not support this usage as described.  You cannot just pass a string to subtract. http://momentjs.com/docs/#/manipulating/subtract/

jj4th: @droogans ^^

Droogans: I did it this way because @CITguy mentioned in another PR that using the `duration` object was redundant. I honestly don't mind either way, I'm not scared of deleting a little verbosity, especially if it doesn't affect the end result.

Also, `elapsed` is an object, which is the [third method signature](http://momentjs.com/docs/#/manipulating/subtract/) in the docs (yay JS libraries accepting 2+ method argument arrangements). I should probably document inline better what's going on.

```js
// ['d', 'h'], ['3', '10'] -> { 'd': '3', 'h': '10' }
var elapsed = _.zipObject(ageParts);
// subtract({ days: '3', hours: '10' });
return moment().utc().subtract(elapsed);
```

It uses named parameters that are pulled from the string via regex.

jj4th: No, that's my bad for not seeing the `_.zipObject` part.

jj4th:
QE-LGTM

PR #1548, opened 2016-02-12
docs(rxLocalStorage): Page objects to jsdoc3

Droogans:
### LGTMs
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/quarks/rxLocalStorage/scripts/rxLocalStorage.page.js
+++ b/src/quarks/rxLocalStorage/scripts/rxLocalStorage.page.js
@@ -1,7 +1,22 @@
 var Page = require('astrolabe').Page;
 
-exports.rxLocalStorage = Page.create({
+/**
+ * @namespace
+ */
+var rxLocalStorage = {
 
+    /**
+     * @function
+     * @description Add a new item to local storage. All parameters are
+     * passed through `JSON.stringify()` before being set.
+     * @param {String} key - The key to store the item under inside of local storage.
+     * @param {String} value - The value to associate with the given `key` under local storage.
+     * @example
+     * it('should set a key in local storage', function () {
+     *   encore.rxLocalStorage.setItem('key', 'value');
+     *   expect(encore.rxLocalStorage.getItem('key')).to.eventually.equal('value');
+     * });
+     */
     setItem: {
         value: function (key, value) {
             var command = function (key, value) {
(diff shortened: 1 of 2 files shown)


src/quarks/rxLocalStorage/scripts/rxLocalStorage.page.js
```diff
@@ -1,7 +1,20 @@
...
+var rxLocalStorage = {
 
+    /**
+     * @function
+     * @description Add a new item to local storage. All parameters are
+     * passed through `JSON.stringify()` before being set.
+     * @param {String} key - The key to store the item under inside of local storage.
+     * @param {String} value - The value to associate with the given `key` under local storage.
+     * @example
+     * rxLocalStorage.setItem('key', 'value');
```
CITguy: Should these be `encore.rxLocalStorage.*` instead of `rxLocalStorage.*`?

Droogans: `encore`.

I'll fix it.

src/quarks/rxLocalStorage/scripts/rxLocalStorage.page.js
```diff
@@ -38,4 +68,6 @@ exports.rxLocalStorage = Page.create({
         }
     }
 
-});
+};
+
+exports.rxLocalStorage = Page.create(rxLocalStorage);
```
jj4th: @Droogans elsewhere you've just set this object directly, why the deviating pattern here?  Not that I mind, I actually like this better.

jj4th:
QE-LGTM pending the addressing of Ryan's comments.

src/quarks/rxLocalStorage/scripts/rxLocalStorage.page.js
Droogans: This is more a question for @tomreece, who authored this.

It may work best to change the first `var rxLocalStorage = {` to `exports.rxLocalStorage = {`. I'm pretty sure this is a collection of static functions that don't need to be wrapped in `Page.create`.

Droogans:
@CITguy I added additional documentation and tests when I went back and reviewed this module.

I wasn't sure what would happen if you tried to get an item from local storage that did not exist, for documentation purposes. So I added a test for it and documented it.

CITguy:
LGTM

PR #1570, opened 2016-02-16
fix(rxLogout): Align with NAM integration

Droogans:
- [ ] Dev LGTM @CITguy 
- [ ] Dev LGTM @parlarjb 
- [ ] QE LGTM @jj4th 

I'd appreciate it if some folks more intimate with the dev code base could double check this and make sure I'm not accidentally deprecating an important feature by removing quark functionality.


src/quarks/UnauthorizedInterceptor/scripts/UnauthorizedInterceptor.js
```diff
@@ -34,7 +34,6 @@ angular.module('encore.ui.quarks')
         },
         responseError: function (response) {
             if (response.status === 401) {
-                Session.logout(); // Logs out user by removing token
```
parlarjb: I think this should be replaced with... something. We should still remove the user's token here. 

You _could_ do `rxLocalStorage.removeItem('encoreSessionToken');`, but I'm not thrilled about that. 

Maybe don't remove the `Session.logout` method from `Session.js`, or possibly rename it to `Session.clearToken()`

100stacks:
JIRA Stories:
https://jira.rax.io/browse/FRMW-679
https://jira.rax.io/browse/FRMW-701

src/quarks/UnauthorizedInterceptor/scripts/UnauthorizedInterceptor.js
halkon: More specifically, it is good to get a "fresh" token if we got a `401` from an API response.  Otherwise, the redirect would go:

* Login
* Seeming Valid Token
* Application
* 401
* Login....

Droogans: I already did that for a demo component focusing on this interceptor.

https://github.com/rackerlabs/encore-ui/pull/1570/files#diff-d24f1b818cc87b2739bfa6a3fc6ba36eL13

I only did that because it controlled the show/hide feature of the notification. Here, though, this isn't doing anything (removing the token is non-functional for logging out of NAM apps, as I've learned).

If you have your gut instinct telling you that something here is amiss, I will add it back gladly. Just let me know how you want it added back (or feel free to push over this PR with your own commits, just make sure it ends up in my fork).

src/quarks/Session/scripts/Session.js
```diff
@@ -61,7 +60,7 @@ angular.module('encore.ui.quarks')
         rxLocalStorage.setObject(TOKEN_ID, token);
     };
 
-    session.logout = function () {
+    session.clearToken = function () {
```
CITguy: To ensure we don't break the 1.x API contract, please alias `clearToken()` as `logout()` and provide a deprecation warning.

```javascript
/**
 * @deprecated
 * JSDOC stuffs here
 */
session.logout = function () {
    console.warn('some warning message here...');
    session.clearToken();
};
```

src/components/rxLogout/scripts/rxLogout.js
```diff
@@ -1,41 +0,0 @@
...
-                // @see http://stackoverflow.com/a/23624785
-                if (!$location.$$html5) {
-                    scope.logoutUrl = '#' + scope.logoutUrl;
-                }
-
-                $window.location = scope.logoutUrl;
-            });
-        }
-    };
-});
```
CITguy: Can we make `rxLogout` deprecated rather than removing it? Removing a stable directive constitutes an API change, and would require a major version bump.

Droogans: Good point. I will put some of this back.

src/components/rxLogout/scripts/rxLogout.js
```diff
@@ -26,7 +30,7 @@ angular.module('encore.ui.rxLogout')
             scope.logoutUrl = (_.isString(scope.rxLogout) && scope.rxLogout.length > 0) ? scope.rxLogout : '/login';
 
             element.on('click', function () {
-                Auth.logout();
+                Auth.clearToken();
```
Droogans: @CITguy @halkon @parlarjb I tried to find any usage of `logout` inside of this now-deprecated API to replace it with something less-deprecated. Let me know if you think I missed any.

src/components/rxLogout/scripts/rxLogout.page.js
```diff
@@ -1,25 +0,0 @@
...
-};
-
-exports.rxLogout = {
-    initialize: function (rxLogoutElement) {
-        rxLogout.rootElement = {
-            get: function () { return rxLogoutElement; }
-        };
-        return Page.create(rxLogout);
-    }
-};
```
Droogans: This page object can go away, if anyone breaks their end to end tests I'll be surprised, seeing how there is no logging out now that NAM is implemented. If there are tests that "logout", they can be deleted as this action is no longer supported.

src/quarks/UnauthorizedInterceptor/scripts/UnauthorizedInterceptor.js
```diff
@@ -34,7 +34,7 @@ angular.module('encore.ui.quarks')
...
+                Session.clearToken();
```
Droogans: @parlarjb I replaced this with `clearToken`, but it's not showing up here as an outdated diff because the old comment is above this one.

src/components/rxLogout/scripts/rxLogout.page.js
CITguy: I'm okay with this. The page object functionality is simple enough to implement, if needed (but it shouldn't be needed). :+1:

src/components/rxLogout/scripts/rxLogout.js
CITguy: Shouldn't this be using `Session.clearToken()` as stated in your deprecation message?

Droogans: They're equivalent. `Auth` was already provided, so it saved me a few keystrokes.

src/components/rxLogout/scripts/rxLogout.page.js
parlarjb: I'm with @Droogans on this one. We're generally more ok with breaking page object APIs, and this is actually a really good one to break. Namely, if someone still has `logout` in their tests, they should remove it right away, and this breaking will be a good signal to let them know.

Droogans:
Closed in favor of https://github.com/rackerlabs/encore-ui/pull/1588 (for now).

PR #1576, opened 2016-02-17
fix(rxAccountInfoBanner): apply null data pattern

CITguy:
Applying null data pattern to rxAccountInfoBanner output.

### LGTMs
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxAccountInfo/templates/rxAccountInfoBanner.html
+++ b/src/components/rxAccountInfo/templates/rxAccountInfoBanner.html
@@ -4,7 +4,7 @@
             <div class="label">Account Name:</div>
             <div class="account-data">
                 <a href="{{ accountPageUrl }}" target="_blank">
-                    {{ accountName }}
+                    {{ accountName || 'N/A' }}
                 </a>
             </div>
         </li>
@@ -18,11 +18,13 @@
         </li>
         <li>
             <div class="label">Account Status:</div>
-            <div class="account-data {{ statusClass }} account-status">{{ accountStatus }}</div>
+            <div class="account-data {{ statusClass }} account-status">
+              {{ accountStatus || 'N/A' }}
+            </div>
         </li>
         <li>
             <div class="label">Access Policy:</div>
-            <div class="account-data">{{ accountAccessPolicy }}</div>
+            <div class="account-data">{{ accountAccessPolicy || 'N/A' }}</div>
         </li>
         <li ng-if="showCurrentUser">
             <div class="label">Current User:</div>


CITguy:
## Preventing
screen shot 2016-02-17 at 4 06 47 pm

## rxAccountInfo Demo

### Before
screen shot 2016-02-17 at 4 34 30 pm

### After
screen shot 2016-02-17 at 4 34 43 pm

100stacks:
LGTM

Droogans:
Don't forget to make sure the page objects return `null` instead of "N/A" for these missing values.

CITguy:
@Droogans, page object properties updated to return `null` instead of `'N/A'`.

src/components/rxAccountInfo/scripts/rxAccountInfo.page.js
```diff
@@ -83,7 +83,12 @@ var rxAccountInfo = {
      */
     name: {
         get: function () {
-            return this.rootElement.element(by.binding('accountName')).getText();
+            return this.rootElement
+                .element(by.binding('accountName'))
+                .getText()
+                .then(function (text) {
+                    return (text === 'N/A' ? null : text);
```
Droogans: should this be spaced out a little more?

```js
.then(function (text) {
    if (text === 'N/A') {
        return null;
    } else {
        return text;
    }
});
```

Droogans:
Is there a way we could get some end to end tests that make sure this works? Also, the docs should mention that the `@type` is `{String|null}`, with updates to the description tag as well.

CITguy:
@Droogans, I'll take another look later this week.

glynnis:
Design LGTM

CITguy:
@Droogans: should the page object really be returning null instead of "N/A"? As a QE dev that's new to rx-page-objects, if I use the page object to determine the value seen on the page, and I get null (not on the page) instead of "N/A" (on the page), my first thought would be that there's a bug in the page object. Is this how it should be?

Droogans:
Hmmm...I think we may have already talked about this in person, and we never got back to writing any of it down.

I think that, as a tester in a project, I would want my end to end tests to recognize "N/A", the string, as a literal value for `null` expressed in a UX-friendly way. However, that would be _my_ application's decision to convert those values to `null`, not rx-page-objects.

If you think about it, if rx-page-objects returns `"N/A"` for this value to an application that is set up to translate that string to `null`, then it will all work seamlessly. If you haven't already set that up, then you're going to get what you're expecting -- "N/A".

Yeah, leave it as "N/A". Good catch. There's no need to force users of the framework to consumer opinionated value-conversion work like this unless they opt-into it themselves.

src/components/rxAccountInfo/scripts/rxAccountInfo.page.js
```diff
@@ -83,7 +83,12 @@ var rxAccountInfo = {
...
+                });
```
Droogans: @CITguy I was wrong about suggesting this `null` transformation. Leave that to the consumer of rx-page-objects to accomplish, not rx-page-objects itself.

CITguy: No worries, I'll revert the changes to the page objects.

CITguy:
@Droogans: I've reverted the page object to reflect what's visible on the page.

PR #1580, opened 2016-02-19
fix(less): apply lesshinting

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-248

Updated to latest lesshint `v1.3.1` to check `src/` and `demo/less/`.

```sh
# install globally
npm install -g lesshint@1.3.1

# check everything in src/
lesshint --config=.lesshint src

# check everything in demo/less
lesshint --config=.lesshint demo/less
```

* adds `.lesshint` configuration for @glynnis, @ssalinas24, or others to utilize
* corrects errors reported by latest version

### LGTMS
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM @Droogans

### DIFF
--- a/.lesshint
+++ b/.lesshint
@@ -1,0 +1,138 @@
+// OPTIONS: https://github.com/lesshint/lesshint/blob/master/lib/linters/README.md
+{
+    "fileExtensions": [".less", ".css"],
+    "excludedFiles": [
+        /*
+         * Ignoring these three until upstream dependency fixes support for
+         * guarded mixins.
+         *
+         * https://github.com/tonyganch/gonzales-pe/issues/121
+         */
+        "flexbox.less",
+        "mixins.less",
+        "layout.less",
+        // upstream dep choking on IE filter
+        "progressbar.less"
+    ],
+    "attributeQuotes": {
+        "enabled": true,
+        "style": "double"
+    },
+    "borderZero": {
+        "enabled": true,
+        "style": "zero"
+    },
+    "comment": {
+        "enabled": false
+    },
+    "decimalZero": {
+        "enabled": true,
+        "style": "leading"
+    },
+    "duplicateProperty": {
+        "enabled": true
+    },
+    "emptyRule": {
+        "enabled": true
+    },
+    "finalNewline": {
+        "enabled": true
+    },
+    "hexLength": {
+        "enabled": true,
+        "style": "long"
+    },
+    "hexNotation": {
+        "enabled": true,
+        "style": "lowercase"
+    },
+    "hexValidation": {
+        "enabled": true
+    },
+    "idSelector": {
+        "enabled": false
+    },
+    "importantRule": {
+        "enabled": false // TODO: enable on audit of !important
+    },
+    "importPath": {
+        "enabled": false
+    },
+    "propertyOrdering": {
+        "enabled": false
+    },
+    "propertyUnits": {
+        "enabled": false,
+        "global": ["px", "%", "em"]
+    },
+    "qualifyingElement": {
+        "allowWithAttribute": true,
+        "allowWithClass": true,
+        "allowWithId": false
+    },
+    "selectorNaming": {
+        "disallowUppercase": true,
+        "disallowUnderscore": true,
+        "disallowDash": false,
+        "exclude": []
+    },
+    "singleLinePerProperty": {
+        "enabled": true
+    },
+    "singleLinePerSelector": {
+        "enabled": true
+    },
+    "spaceAfterPropertyColon": {
+        "enabled": true,
+        "style": "one_space"
+    },
+    "spaceAfterPropertyName": {
+        "enabled": true,
+        "style": "no_space"
+    },
+    "spaceAfterPropertyValue": {
+        "enabled": true,
+        "style": "no_space"
+    },
+    "spaceAroundComma": {
+        "enabled": true,
+        "style": "after"
+    },
+    /* As of 1.3.1 this is returning false negatives when it comes across
+     * negative numbers in calculations or shorthand.
+     *
+     * https://github.com/lesshint/lesshint/issues/129
+     */
+    "spaceAroundOperator": {
+        "enabled": false, // disabled until lesshint#129 fixed
+        "style": "both"
+    },
+    "spaceBeforeBrace": {
+        "enabled": true,
+        "style": "one_space"
+    },
+    "spaceBetweenParens": {
+        "enabled": true,
+        "style": "no_space"
+    },
+    "stringQuotes": {
+        "enabled": true,
+        "style": "double"
+    },
+    "trailingSemicolon": {
+        "enabled": true
+    },
+    "trailingWhitespace": {
+        "enabled": true
+    },
+    "urlFormat": { // we want both
+        "enabled": false
+    },
+    "urlQuotes": {
+        "enabled": true
+    },
+    "zeroUnit": {
+        "enabled": true,
+        "style": "no_unit"
+    }
+}
--- a/demo/less/demoApp.less
+++ b/demo/less/demoApp.less
@@ -5,8 +5,8 @@
 
 /* TODO: (deprecated) remove with last of components */
 .module-description {
-  border: 1px solid #aaa;
-  background: #fff;
+  border: 1px solid #aaaaaa;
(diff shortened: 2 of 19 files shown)


.lesshint
```diff
@@ -0,0 +1,138 @@
...
+         * Ignoring these three until upstream dependency fixes support for
+         * guarded mixins.
+         *
+         * https://github.com/tonyganch/gonzales-pe/issues/121
+         */
+        "flexbox.less",
+        "mixins.less",
+        "layout.less",
+        // upstream dep choking on IE filter
+        "progressbar.less"
```
CITguy: Rather than try to correct upstream dep, it might be worth looking into removing unused styles instead.

Droogans:
This should be installed locally and run at the TravisCI level. I think something like `npm run lesshint` would be more than plenty to make sure no regressions pop up as less files are modified over time.

.lesshint
```diff
@@ -0,0 +1,138 @@
+// OPTIONS: https://github.com/lesshint/lesshint/blob/master/lib/linters/README.md
+{
+    "fileExtensions": [".less", ".css"],
+    "excludedFiles": [
+        /*
+         * Ignoring these three until upstream dependency fixes support for
+         * guarded mixins.
+         *
+         * https://github.com/tonyganch/gonzales-pe/issues/121
```
CITguy: First issue causing havoc in results. Causes parsing to fail because upstream dependency doesn't recognize guarded Less mixins as valid Less.

.lesshint
```diff
@@ -0,0 +1,138 @@
...
+        "style": "no_space"
+    },
+    "spaceAroundComma": {
+        "enabled": true,
+        "style": "after"
+    },
+    /* As of 1.3.1 this is returning false negatives when it comes across
+     * negative numbers in calculations or shorthand.
+     *
+     * https://github.com/lesshint/lesshint/issues/129
```
CITguy: Second issue causing havoc. Greedy regex causes false negative with negative numbers.

CITguy:
@Droogans, I would install it locally, except there are bugs preventing us from performing a full check. This PR is to get the configuration in place for other devs to play around with, and to correct what was reported.

The two bugs blocking us from a  local install are as follows.

* lesshint/lesshint#129
* tonyganch/gonzales-pe#121

Droogans:
So what I'm getting is that the intention is to have designers and devs run it locally every so often and submit a PR to scrub less files of warts?

CITguy:
@Droogans: currently, yes. Once the upstream bugs are taken care of, we can look into integrating it with TravisCI/Grunt.

parlarjb:
Do you want to add usage instructions to our docs somewhere?

LGTM

CITguy:
@parlarjb, hows this... https://github.com/rackerlabs/encore-ui/wiki/Linting-and-Hinting

parlarjb:
Works for me

100stacks:
@CITguy, should we add a caveat to the documentation indicating this version of `lesshint` is Rackspace specific?  

LGTM

demo/less/demoApp.less
```diff
@@ -39,7 +39,7 @@
   h4 {
     font-size: 1.1em;
     font-weight: bold;
-    color: #777;
+    color: #777777;
```
glynnis: I think the `#aaaaaa`, `#777777` and `#cccccc` in this file should be converted to use vars from the palette. @CITguy is that something you'd like to take care of as part of this PR, or would you prefer that I make a story to tackle that separately?

glynnis: Or maybe it doesn't matter since this is slated for deprecation?

src/components/rxApp/rxApp.less
```diff
@@ -504,7 +504,7 @@
   // TODO: Update to 2.0 Color Variables
   .page-header {
     background: linear-gradient(to bottom, #e7e7e7 0%, #d6d6d6 100%);
-    border-bottom: 1px solid #bbb;
+    border-bottom: 1px solid #bbbbbb;
```
glynnis: We should convert `#bbbbbb` to use vars from the palette as above

src/components/rxBreadcrumbs/rxBreadcrumbs.less
```diff
@@ -16,7 +16,7 @@
       color: @gray-900;
       position: relative;
       text-decoration: none;
-      text-shadow: 0 1px 0 rgba(255,255,255,.5);
+      text-shadow: 0 1px 0 rgba(255, 255, 255, 0.5);
```
glynnis: Convert to vars

src/components/rxBulkSelect/docs/rxBulkSelect.demo.less
```diff
@@ -1,5 +1,11 @@
...
-  .success-status { color: @green; }
-  .failure-status { color: @red; }
+  .pending-status {
+    color: @orange;
+  }
+  .success-status {
+    color: @green;
+  }
+  .failure-status {
+    color: @red;
```
glynnis: All of these are old vars which will be deprecated. We should convert these to the new palette.

src/components/rxForm/rxForm.less
```diff
@@ -188,10 +188,10 @@
 
   &:after {
     // down caret from font awesome
-    content: '\f0d7';
+    content: "\f0d7";
     font-family: FontAwesome;
     font-size: 15px;
-    color: #aaa;
+    color: #aaaaaa;
```
glynnis: Convert to use new palette vars

glynnis:
:clap: :clap: :clap: :clap: :clap: :clap: :balloon: :cake: :tada: :sparkles: :sparkling_heart: 

Let me know if the remaining hex codes I commented on here should be split out into a separate PR -- I'm happy to create JIRA stories if that's the preferred route (instead of updating this work).

demo/less/demoApp.less
CITguy: I'm comfortable with leaving it as is for now. It will be going away with final Atoms/Molecules migration.

src/components/rxApp/rxApp.less
CITguy: `@gray-500`?

src/components/rxBulkSelect/docs/rxBulkSelect.demo.less
CITguy: Ah, true. Which vars should be used instead? Something like the following?

```
@orange --> @orange-700
@green --> @green-500
@red --> @red-500
```

src/components/rxForm/rxForm.less
CITguy: Would this be closer to `@gray-500` or `@gray-600`?

CITguy: This is actually for a deprecated style for old rxForm stuff (`.field-select`). I mentioned it in [FRMW-713](https://jira.rax.io/browse/FRMW-713), but I won't lose any sleep if this weren't converted.

CITguy:
@glynnis, I've created FRMW-713 to handle the conversion to variables. I'll get this PR rebased so it is mergeable.

CITguy:
I'm going to try closing and reopening the PR to see if that has any effect on the merge issue.

100stacks:
@CITguy have you tried making a change to a file (e.g., add a comment) to trigger a new build hash?

CITguy:
@100stacks: yeah... it might help to push to `upstream` instead of `origin` d'oh

100stacks:
screen shot 2016-02-29 at 2 31 52 pm

.lesshint
```diff
@@ -0,0 +1,138 @@
...
+        "style": "lowercase"
+    },
+    "hexValidation": {
+        "enabled": true
+    },
+    "idSelector": {
+        "enabled": false
+    },
+    "importantRule": {
+        "enabled": false // TODO: enable on audit of !important
```
CITguy: As of `lesshint 1.4.0` you can [disable rules inline](https://github.com/lesshint/lesshint#inline-configuration) similar to other hinters.

CITguy:
@glynnis, updates have been made... or stories have been made to capture updates.

ssalinas24:
LGTM
As long as updating colors will be addressed in a future story or stories.

PR #1581, opened 2016-02-20
refactor(Buttons): rearrange Buttons/rxButton docs

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-271

### LGTMS
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM


CITguy:
Actually... I know exactly why it's failing... Time Zones (it's currently after 6:00pm CST, which makes it tomorrow UTC).

100stacks:
@CITguy, nice work!

Only suggestion would be to include screenshots of the new `Buttons` and `rxButton` pages.  LGTM :+1:

CITguy:
@100stacks, yeah. I took them, but forgot to add them, because I thought the tests would fail. I'll add them.

CITguy:
### rxButton ngdocs
rxbutton-ngdocs-after

### rxButton Atom demo
rxbutton-after

### Button Styleguide Docs
buttons-after

parlarjb:
LGTM

glynnis:
@CITguy A lot of the styling/visual changes from the mock-up/JIRA story aren't represented in the work here. Is that being tackled as part of a separate effort?

CITguy:
There are two things that cannot be completed at present.

#### Example Styling (vertical and side-by-side)
I recall having a discussion including @lfielder and a couple other designers around questions regarding the examples in the mockup. There were questions around whether to automatically collapse the source tabs, and how we should handle long source code in side-by-side examples. I thought that it was being discussed on the UX board, but I'm not seeing any stories around this topic.

#### Page layout
The only update that can currently be made with the page headers is the addition of keywords (last updated cannot be reliably determined with current methods). However, this can be updated at a later time, because I've not seen a full list of keywords for each Atom/Molecule/Quark.

#### Summary
With the above points in mind, it makes more sense to get the information in the right place, first. The updated page headers and examples can be completed at a later time.

CITguy:
@glynnis: I've made FRMW-711 to capture updating the page layout and FRMW-712 for examples.

glynnis:
:+1: Thanks for the clarification @CITguy.

Design LGTM

Droogans:
The only thing that doesn't look good to me is that there's still page objects for rxButton :-1: :dancer:

PR #1582, opened 2016-02-23
refactor(tables): Put Tables into Atoms

parlarjb:
https://jira.rax.io/browse/FRMW-273

 - [x] Dev LGTM
 - [x] Dev LGTM
 - [x] Design LGTM

screen shot 2016-02-23 at 10 28 31 am

### DIFF
--- a/demo/templates/atoms/tables.html
+++ b/demo/templates/atoms/tables.html
@@ -1,0 +1,28 @@
+
+<rx-page title="'Tables'">
+  <ul class="list">
+    <li>
+      Tables have default styles: regular rows have white backgrounds, while
+      headers use white text with a gray background. An additional
+      <code>.table-striped</code> class is available for alternate row striping.
+    </li>
+    <li>
+      <span class="msg-info">ENCORE-SPECIFIC:</span> You can use tables to show
+      data-table related details, such as listing of objects.
+    </li>
+    <li>
+      If you need to show summary-style important pieces of metadata that aren't
+      necessarily related to each other, consider using an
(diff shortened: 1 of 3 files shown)


100stacks:
LGTM

demo/templates/atoms/tables.html
```diff
@@ -0,0 +1,28 @@
+
+<rx-page title="'Tables'">
```
ssalinas24: Insert `<hr/>` tag after line number 2 to get the horizontal rule line below title.

demo/templates/atoms/tables.html
```diff
@@ -0,0 +1,28 @@
...
+  <ul class="list">
```
ssalinas24: @glynnis  and @CITguy The mockups don't show content formatted with lists, but previously we have had items in lists. Are we going to keep the content formatted in lists, and ignore mockups?

demo/templates/atoms/tables.html
CITguy: This would be part of FRMW-711

demo/templates/atoms/tables.html
CITguy: The mockup seems to have removed the list that currently exists in documentation. I'd rather keep the list.

glynnis: @ssalinas24 I think given the rest of the cosmetic/visual changes that will need to be made down the line, it's okay to leave this in a list for now.

glynnis:
Design LGTM

CITguy:
LGTM

PR #1583, opened 2016-02-24
fix(rxMultiSelect): Fix overlapping rxMultiSelect issue

parlarjb:
https://jira.rax.io/browse/FRMW-641

This is one possible solution, James and Steven will be doing
PRs with other solutions

### DIFF
--- a/src/components/rxMultiSelect/scripts/rxMultiSelect.js
+++ b/src/components/rxMultiSelect/scripts/rxMultiSelect.js
@@ -93,10 +93,13 @@ angular.module('encore.ui.rxMultiSelect')
             rxSelectDirective[0].link.apply(this, arguments);
 
             var previewElement = rxDOMHelper.find(element, '.preview')[0];
+            var parentElement = element.parent(); // Should be a <div class="rxSelect">
+            var originalZIndex = parentElement.css('z-index'); // Capture z-index, if any
 
             var documentClickHandler = function (event) {
                 if (event.target !== previewElement) {
                     scope.listDisplayed = false;
+                    parentElement.css('z-index', originalZIndex);
                     scope.$apply();
                 }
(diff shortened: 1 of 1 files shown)


src/components/rxMultiSelect/scripts/rxMultiSelect.js
```diff
@@ -111,6 +114,12 @@ angular.module('encore.ui.rxMultiSelect')
             scope.toggleDisplay = function (event) {
                 if (event.target === previewElement) {
                     scope.listDisplayed = !scope.listDisplayed;
+                    if (scope.listDisplayed) {
+                        // Ensure the dropdown from this rxMultiSelect doesn't
+                        // have weird issues with another rxMultiSelect that lives
+                        // right under it
+                        parentElement.css('z-index', 1000);
```
Droogans: What do you think about using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER instead?

You could subtract a few from it if you wanted to be extra safe for users who need to force something over this directive.

CITguy:
Does it make sense to fix a CSS issue with javascript?

parlarjb:
@CITguy That was a big part of the conversation that I had with @ssalinas24 and @100stacks.

If we can do it pure CSS, without major refactoring, then let's do that. Otherwise we have this as a solution.

100stacks:
@parlarjb, just pushed an alternate solution with PR https://github.com/rackerlabs/encore-ui/pull/1586.  Paired up with @CITguy on this one where we removed `rxSelect` dependencies to make `rxMultiSelect` a stand-alone module.

CITguy:
Closing. Handled in #1586

PR #1586, opened 2016-03-02
refactor(rxMultiSelect): FRMW-641 z-index bug + convert to Molecule

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-641

### LGTMs
- [ ] Dev LGTM
- [ ] Dev LGTM
- [ ] Design LGTM
- [ ] QE LGTM

---
- Refactored `rxMultiSelect` to remove `rxSelect` dependencies
- Resolves `z-index` content stacking issues


100stacks:
### `rxMultiSelect` Demo Page

screen shot 2016-03-02 at 1 27 50 pm
rxmultiselect-refactored


### `rxMultiSelect` API Docs
screen shot 2016-03-02 at 1 28 28 pm

100stacks:
@CITguy, thanks for pairing up with me on this one! :100: :tada:

CITguy:
screen shot 2016-03-02 at 4 24 46 pm

@100stacks, I'd like to see a State table similar to the above to verify the style in different states. This should be similar to what's at the bottom of the [rxSelect demo](http://rackerlabs.github.io/encore-ui/#/components/rxSelect) and shouldn't need the `Disabled (disabled)` column.

src/components/rxMultiSelect/templates/rxMultiSelect.html
```diff
@@ -1,10 +1,21 @@
...
+  </div>
+
+  <div class="control" ng-click="toggleMenu()">
+    <div class="preview">{{ preview }}</div>
+
+    <div class="select-trigger">
+      <i class="fa fa-fw fa-caret-down"></i>
+    </div>
+
+    <ul ng-show="listDisplayed">
```
ssalinas24: To me this feels wrong to have a `<ul>` without any list tags in it. Can we switch this to something else, like another HTML tag or a custom element?

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -1,34 +1,83 @@
...
+
+      // This allows pointer events to pass THROUGH to the
+      // native <select> element
+      //pointer-events: none !important;
+    }//.select-trigger
+
+    ul {
+      z-index: 100;
+      position: absolute;
+      top: -1px;
```
ssalinas24: Unless trying to hide something, the negative values are usually a bad idea. This usually indicates that other pixel measurements, padding, or margins aren't totaling up well. We can look into this if needed and see if we improve on anything.

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
 @rxSelect-trigger-color-disabled: @rxSelect-border-color-disabled;
 
 /*
+ * rxMultiSelect
+ */
+@rxMultiSelect-min-width: @rxSelect-min-width;
```
ssalinas24: Is this for the overall list, or for the options. Perhaps the naming can be more specific.

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-height: @rxSelect-height;
```
ssalinas24: At first glance, i wonder if this means the whole list, or the list options/items. Maybe naming can be more specific to the  "list" or option item.

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-color: @rxSelect-color;
+@rxMultiSelect-background: @rxSelect-background;
```
ssalinas24: if this is for background color then it may benefit to be "@rxMultiSelect-list-background-color" or "@rxMultiSelect-list-ption-background-color".

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-trigger-background: @rxSelect-trigger-background;
```
ssalinas24: Should be "@rxMultiSelect-trigger-background-color".

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-color-disabled: @rxSelect-color-disabled;
```
ssalinas24: If this is text color then it should be "@rxMultiSelect-disabled-text-color" to fit with our current naming pattern.

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-background-disabled: @rxSelect-background-disabled;
```
ssalinas24: Should be "@rxMultiSelect-disabled-background-color", and may also need to include verbiage to describe which object or element this is targeting. Like "List" or "option".

src/styles/vars.less
```diff
@@ -291,6 +291,26 @@
...
+@rxMultiSelect-border-color: @rxSelect-border-color;
```
ssalinas24: Meet with me for the rest of the vars below. I think we can straighten out the naming.

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -1,34 +1,83 @@
...
+      display: flex;
+      align-items: center;
+      justify-content: center;
+
+      // This allows pointer events to pass THROUGH to the
+      // native <select> element
+      //pointer-events: none !important;
+    }//.select-trigger
+
+    ul {
```
ssalinas24: make a class and target the class instead of the UL.

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -1,34 +1,83 @@
 /*
  * rxMultiSelect
  */
-rx-multi-select {
+.rx-multi-select {
   z-index: auto !important;
```
ssalinas24: Will it break if we get rid of the "!important" portion of this property? If not, then lets try to get rid of it. If removing, then inspect to see if any z-index is being applied from something else.

src/components/rxMultiSelect/rxMultiSelect.less
CITguy: This is to ensure that the popup menu covers the border of its parent.

src/components/rxMultiSelect/rxMultiSelect.less
CITguy: We can probably remove the `!important`, but the property must stay `auto` to ensure that there is no stacking context.

src/components/rxMultiSelect/templates/rxMultiSelect.html
100stacks: @ssalinas24, great point.  We changed this wrapper a `<div>` tag.

100stacks:
@ssalinas24, feedback implemented. :+1:

100stacks:
Current updates fix the `z-index` bug.  After further discussions with @CITguy, we decided to refactor this to a `molecule`.

100stacks:
Refactored `rxMultiSelect` into a `Molecule`.  Screenshots and gif updated in https://github.com/rackerlabs/encore-ui/pull/1586#issuecomment-191397368.

src/components/rxMultiSelect/rxMultiSelect.module.js
```diff
@@ -10,12 +10,12 @@
  * * Links to service APIs provided by rxMultiSelect component.
  *
  * ## Directives
- * * {@link rxMultiSelect.directive:rxMultiSelect rxMultiSelect}
+ * * {@link molecules.directive:rxMultiSelect rxMultiSelect}
  * * {@link rxMultiSelect.directive:rxSelectOption rxSelectOption}
  *
  * ## Related Directives (if applicable)
  * * Links to directive APIs provided by other components.
```
CITguy: screen shot 2016-03-07 at 1 21 26 pm

* remove "Services" section
* remove "Related Directives" section
* migrate `rxSelectOption` to its own Molecule
* eliminate `encore.ui.rxMultiSelect` module

src/molecules/rxMultiSelect/rxMultiSelect.less
```diff
@@ -0,0 +1,118 @@
...
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: transparent;
+  }
+}//rx-multi-select
+
+.rx-select-option {
```
CITguy: This should be split out when you migrate `rxSelectOption` to a Molecule.

src/molecules/rxMultiSelect/scripts/rxMultiSelect.exercise.js
```diff
@@ -82,7 +82,7 @@ exports.rxMultiSelect = function (options) {
         });
 
         it('hides the menu when another element is clicked', function () {
```
CITguy: Should probably rename the test to reflect the action. I'd say... `should hide the menu when backdrop is clicked`.

src/molecules/rxMultiSelect/scripts/rxMultiSelect.js
```diff
@@ -1,11 +1,12 @@
...
 /**
  * @ngdoc directive
- * @name rxMultiSelect.directive:rxMultiSelect
+ * @name molecules.directive:rxMultiSelect
  * @restrict E
  * @scope
+ * @requires rxMultiSelect.directive:rxSelectOption
  * @description
- * This component is a multi-select dropdown with checkboxes for each option.
+ * This molecules is a multi-select dropdown with checkboxes for each option.
```
CITguy: `molecules` -> `molecule`

CITguy:
If we're going to convert to a Molecule, might as well convert `rxSelectOption` as well, so we can get rid of the `encore.ui.rxMultiSelect` module.

src/components/rxMultiSelect/rxMultiSelect.module.js
100stacks: @CITguy, from your https://github.com/rackerlabs/encore-ui/pull/1586#issuecomment-193414575 I am eliminating `encore.ui.rxMultiSelect` component with this PR.

CITguy: yes, `rxMultiSelect` and `rxSelectOption` should be migrated to Molecules (leaving nothing left in `encore.ui.rxMultiSelect`).

100stacks:
## `rxSelectOption` Molecule

### Demo Page
screen shot 2016-03-07 at 3 55 10 pm

### API Page
screen shot 2016-03-07 at 3 55 56 pm

100stacks:
@CITguy, `rxSelectOption` molecule added.  :white_check_mark:

src/molecules/rxMultiSelect/scripts/rxMultiSelect.spec.js
```diff
@@ -61,10 +61,19 @@ describe('rxMultiSelect', function () {
             });
 
             it('does not toggle the visibility of the menu when a child element is clicked', function () {
-                angular.element(el[0].querySelector('rx-select-option')).click();
+                angular.element(el[0].querySelector('.menu')).click();
                 expect(isolateScope.listDisplayed).to.be.false;
             });
 
+            it('should close an open menu', function () {
```
100stacks: @CITguy, added this for code coverage of `closeMenu()` here: https://github.com/rackerlabs/encore-ui/pull/1586/files#diff-4aa8891d7fce42bfb412345afc7b4efdR107.

src/molecules/rxMultiSelect/docs/rxMultiSelect.docs.html
```diff
@@ -0,0 +1,78 @@
+<div ng-controller="rxMultiSelectSimpleCtrl">
+  <p>
+    This component is a multi-select dropdown with checkboxes for each option.
+  </p>
+
+  <rx-example name="rxMultiSelect.simple"></rx-example>
+
+  <h2>Molecule States</h2>
```
100stacks: @CITguy, please let me know if you want me to update the exercises to accommodate the `rxMultiSelect` State Table added by https://github.com/rackerlabs/encore-ui/pull/1586#issuecomment-191463862.

src/molecules/rxMultiSelect/docs/rxMultiSelect.docs.html
```diff
@@ -0,0 +1,78 @@
...
+            id="msInvalidDisabled"
+            ng-model="data.invalidDisabled"
+            always-invalid
+            ng-disabled="true">
+
+            <rx-select-option value="Not allowed"></rx-select-option>
+          </rx-multi-select>
+        </td>
+      </tr>
+   </tbody>
```
CITguy: indentation off by one space

src/molecules/rxMultiSelect/docs/rxMultiSelect.docs.html
```diff
@@ -0,0 +1,78 @@
+<div ng-controller="rxMultiSelectSimpleCtrl">
```
CITguy: Use a different controller for the demo itself. The `rxMultiSelectSimpleCtrl` should be used in the example only.

I'd go with `rxMultiSelectDemoCtrl`.

100stacks:
@CITguy, I had a `rxSelectOption` categorization question after discussing with @glynnis.  Can `rxSelectOption` be a visual element used independently of other `molecules`?  If not, would we make it an `atom`?

Glynnis raised concerns if would it make sense as a `molecule` from a design perspective.

CITguy:
Because `rxSelectOption` makes use of `rxCheckbox`, I'd categorize it as a Molecule.

glynnis:
Thanks for chiming in @100stacks, you beat me to it! I've been struggling with how to articulate why I think rxSelectOption should be an Atom (words are hard). @CITguy I struggled to word this to @100stacks and I'm sure I'll do it here, too. I am not braining or Englishing very well today, but let me give this a shot:

I get the `<rx-select-option>` is to `<rx-multi-select>` as `<option>` is to `<select>` analogy, and I think that analogy does a good job of helping explain why rxSelectOption, in my mind, doesn't work well as a molecule. When I look at the list of other molecules, all of them are things that I think of as "complete" -- they can stand alone and still function. But if I'm understanding rxSelectOption correctly, it can't really be used outside of rxMultiSelect as a stand-alone UI element (just as `<option>` can't be used outside of `<select>` in a meaningful way). 

Even though rxSelectOption uses other atoms like rxCheckbox in its implementation, I think rxSelectOption is more like an atom than a molecule, since there's not a very good way to show a visual demo or how it can be utilized by itself. I think if a component has a dependency on something else to be visible or feature-complete for the user, it probably isn't a good candidate for a molecule.

Hopefully that's a coherent explanation. I realize it's leaning toward a more user/design-centric use of the taxonomy, so let me know what you think. Primarily I just want to round out my understanding (and work toward a shared understanding that we can easily explain to other people) of atoms vs. molecules, whichever way we end up going :+1:

100stacks:
@Droogans, I added `QE LGTM` back for your review, since we revised `rxMultiSelect` demo page to include a State Table with updated tests.

Thanks, all! :+1:

src/molecules/rxSelectOption/scripts/rxSelectOption.js
```diff
@@ -1,10 +1,10 @@
-angular.module('encore.ui.rxMultiSelect')
+angular.module('encore.ui.molecules')
 /**
  * @ngdoc directive
- * @name rxMultiSelect.directive:rxSelectOption
+ * @name molecules.directive:rxSelectOption
  * @restrict E
  * @description
```
CITguy: Please add a `@requires atoms.directive:rxCheckbox`

parlarjb:
@CITguy I'm inclined to agree with @glynnis on her perception for why this should be an atom. The ability to use make use of something/visually demonstrate its functionality is a good decider for something being a molecule.

It's of course confused by the fact that "atom" implies "atomic" which implies "stands on its own", but I'd be happy with "may only be defined as a molecule if it can be displayed on its own"

Note that this is "if" not "if and only if", i.e. some atoms can be displayed on their own as well.

CITguy:
Given the Atomic design definition, Molecules are made of Atoms and other Molecules. If I have a directive that depends on another directive to generate its HTML, I consider that a Molecule. It cannot function independently of its requirement(s).

rxSelectOption requires rxCheckbox to build its HTML, so it cannot exist without rxCheckbox. For this reason, I would consider rxSelectOption a Molecule.

100stacks:
We might be missing a level of categorization in our current Quarks, Atoms, and Molecules nomenclature.

I understand where @glynnis and @parlarjb are coming from in that `rxSelectOption` would not be used as a stand-alone interactive element.  If we go with the next level in Atomic Design, reusable, stand-alone interactive elements combine Molecules into Organisms.  So this would make `rxMultiSelect` an Organism, `rxSelectOption` a Molecule, `rxCheckBox` and text fields Atoms.

I don't know if this would resolve the different nomenclature perspectives, or we could possibly have different types (stand-alone/reusable or dependent) of Molecules?

src/molecules/rxSelectOption/scripts/rxSelectOption.js
100stacks: @CITguy updated this along with the screenshot in https://github.com/rackerlabs/encore-ui/pull/1586#issuecomment-193833833.

100stacks:
@glynnis, I added `DO NOT MERGE` and `needs discussion` labels based on your scheduled follow-up meeting.

CITguy:
@100stacks, considering that there's some disagreement on whether `rxSelectOption` should be an Atom or a Molecule, lets leave it where it is so we can move forward with this PR.

src/components/rxMultiSelect/rxMultiSelect.module.js
```diff
@@ -1,21 +0,0 @@
...
- * ## Directives
- * * {@link rxMultiSelect.directive:rxMultiSelect rxMultiSelect}
- * * {@link rxMultiSelect.directive:rxSelectOption rxSelectOption}
- *
- * ## Related Directives (if applicable)
- * * Links to directive APIs provided by other components.
- */
-angular.module('encore.ui.rxMultiSelect', [
-    'encore.ui.rxSelectFilter'
-]);
```
CITguy: * Put this module back to support `rxSelectOption` as a component.
* Remove/Relocate `rxMultiSelect` documentation.

100stacks:
@glynnis, after discussing with @CITguy, we're going to revert `rxMultiSelect` and `rxSelectOption` molecules back to `rxMultiSelect` component.  This will allow us to push the `z-index` issue fix for our users. :busts_in_silhouette: :+1:

100stacks:
Hey all, I'm going to close this PR and open other with the `z-index` fix to https://jira.rax.io/browse/FRMW-641 pre-molecule shuffle and to save everyone's sanity. :smiley:

PR #1588, opened 2016-03-03
fix(rxApp): Drop all user id display and logout display

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-679

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM

This is the least amount of work possible to keep everything working as is
while still blocking users from attempting to logout in the most obvious place
within encore apps, the left hand nav.

Anyone who is still supporting "logout" functionality will still be able to
"logout" users, even though it's not supported.

BREAKING CHANGE: Anyone who is checking the rxApp page objects for the userId
property will have their tests fail. This isn't a huge deal because part of the
update to the new version of the framework will make those tests irrelevant. Just
delete all tests exercising logout functionality and checking the currently logged
in user from your end to end test suite to successfully migrate to the newest
version of rx-page-objects.

### DIFF
--- a/src/components/rxApp/docs/rxApp.js
+++ b/src/components/rxApp/docs/rxApp.js
@@ -1,6 +1,10 @@
 /*jshint unused:false*/
 angular.module('demoApp')
-.controller('rxAppCtrl', function ($scope, $location, $rootScope, $window, encoreRoutes, rxVisibility) {
+.controller('rxAppCtrl', function ($scope, $location, $rootScope, $window, encoreRoutes, rxVisibility, Session) {
+    Session.getUserId = function () {
+        return 'bert3000';
+    };
+
     $scope.subtitle = 'With a subtitle';
 
     $scope.changeSubtitle = function () {
--- a/src/components/rxApp/docs/rxApp.midway.js
+++ b/src/components/rxApp/docs/rxApp.midway.js
@@ -19,10 +19,8 @@ describe('rxApp', function () {
         expect(rxAppCustom.sectionTitle).to.eventually.equal('Example Menu');
     });
 
-    it('should logout', function () {
-        rxAppCustom.logout();
-        expect(demoPage.currentUrl).to.eventually.contain('login');
-        demoPage.go('#/components/rxApp');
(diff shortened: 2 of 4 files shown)


src/components/rxApp/docs/rxApp.js
```diff
@@ -1,6 +1,10 @@
 /*jshint unused:false*/
 angular.module('demoApp')
-.controller('rxAppCtrl', function ($scope, $location, $rootScope, $window, encoreRoutes, rxVisibility) {
+.controller('rxAppCtrl', function ($scope, $location, $rootScope, $window, encoreRoutes, rxVisibility, Session) {
+    Session.getUserId = function () {
```
CITguy: What is using this?

Droogans: This is just for the demo, making sure that you don't see the username anywhere in the left hand nav anymore. Is it confusing?

I follow up with an assertion in the end to end tests. It may not be appropriate to place this code in the demo.

CITguy: I'm confused to how overriding the function prevents the value from being displayed in the end-to-end test. Is there something I'm missing?

Droogans: If you put the html back like it was before, this `Session.getUserId` over ride will place `'Logout (bert3000)'` in the rxApp demo. I wanted to make sure that no user id would be displayed at all for an end to end test.

CITguy: Ok, so you purposefully set a username to verify its removal from the page. gotcha

CITguy: Can we place a comment to mark for removal in a future release? This seems to make sense only for this PR. Once it's merged, the test and override code becomes unnecessary.

CITguy:
LGTM

100stacks:
LGTM

glynnis:
Design LGTM

jj4th:
This feels so dirty, I love everything about it.  QE-LGTM

PR #1589, opened 2016-03-03
E2e no dates just moment

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-715

### LGTMs
- [ ] Dev LGTM
- [ ] Dev LGTM
- [ ] QE LGTM

I have altered the output of all page objects to be `moment` objects.

This was started a few weeks ago, before @jj4th and @CITguy decided to switch over to strings for input. Does this also mean that there will be strings for output too?

This PR may no longer be relevant. However, there are pieces of it that are nice to have, namely rxMisc.compareMoments (https://github.com/Droogans/encore-ui/commit/e6301a405fe17332d0baadaa6e6bedd12a54c251), which allows me to test moment promises in one line instead of two.


jj4th:
My motivation was that you aren't, as a user, dealing with moment objects, you're dealing with strings.  So the user on the page is going to see `[Feb 03, 2016]` or `[01:50PM]`.  In practicality, when writing tests it simplifies to the difference in:

 ```javascript
// this
expect(blah).to.eventually.eql(moment('Feb 03, 2016 (+0600)'));
// or this
expect(blah).to.eventually.eql('Feb 03, 2016 (+0600)');
```

Which doesn't make that big a difference to me.  Since then I've thought of pros and cons to both approaches... but since one format is never more than a single, simple function call away from the other (moment to string conversion and vice versa is trivial), the only strong case I could make is potentially that using moment makes doing tests involving `today` easier.

File: src/components/rxMisc/scripts/rxMisc.page.js [hunk 1]
```diff
@@ -93,6 +94,64 @@ exports.rxMisc = {
...
+     * attempt to create a moment object by passing in a date string, instead of using Javascript's `new Date`.
+     * @see https://github.com/moment/moment/issues/1407
+     * @param {String} dateText - A string that represents a date that `new Date` can parse into a date object.
+     * @returns {moment}
+     * @example
+     * var myPage = {
+     *     get startDate() { return element(by.binding('startDate')).getText().then(encore.rxMisc.moment); }
+     * }
+     */
+    moment: function (dateText) {
```
CITguy: I'd name this something like `asMoment` to denote that you're creating a moment object from a date string.

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
```diff
@@ -70,13 +70,13 @@ var rxDatePicker = {
      * @function
      * @instance
      * @description Whether or not the date passed in matches the "today" date in the calendar.
-     * @param {Date} [date=new Date] - The date to check against the calendar to confirm that it is today's date.
+     * @param {moment|Date} [m=moment()] - The date to check against the calendar to confirm that it is today's date.
```
CITguy: Where in the logic is `m` defaulted to `moment()`?

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
```diff
@@ -86,39 +86,42 @@ var rxDatePicker = {
...
                 ])) {
                     return;
                 }
 
-                if (date > currentDate) {
+                if (m.isAfter(currentMoment)) {
                     self.nextMonth();
                 } else {
                     self.previousMonth();
                 }
```
CITguy: How do you handle a moment corresponding to a date that is more than 1 month away from the current month? Am I reading that this logic will only navigate to an adjacent month?

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
```diff
@@ -320,29 +323,29 @@ var rxDatePicker = {
...
+            return this.rootElement.$('.displayValue').getText().then(exports.rxMisc.moment);
         },
 
-        set: function (date) {
+        set: function (m) {
             this.open();
-            this.monthAndYear = date;
-            var targetDateString = date.toISOString().split('T')[0];
+            this.monthAndYear = moment(m);
+            var targetDateString = m.format().split('T')[0];
```
CITguy: Can't you just use `var targetDateString = m.format('YYYY-MM-DD');`?

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
```diff
@@ -70,13 +70,13 @@ var rxDatePicker = {
...
-     * @param {Date} [date=new Date] - The date to check against the calendar to confirm that it is today's date.
+     * @param {moment|Date} [m=moment()] - The date to check against the calendar to confirm that it is today's date.
      * @returns {Boolean}
      */
     isDateToday: {
-        value: function (date) {
+        value: function (m) {
             this.open();
-            var dateString = moment(new Date(date)).format().split('T')[0];
+            var dateString = moment(m).format().split('T')[0];
```
CITguy: Why not `var dateString = moment(m).format('YYYY-MM-DD');`?

CITguy:
Rather than `moment(m).format().split('T')[0]` you should be able to use `moment(m).format('YYYY-MM-DD')`.

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
Droogans: It's using recursion. It will keep evaluating the current month and year (and clicking `nextMonth`/`previousMonth`) over and over until it matches the desired target year and month.

src/molecules/rxDatePicker/scripts/rxDatePicker.page.js
Droogans: @CITguy 

> Where in the logic is m defaulted to moment()?

`moment(m)` will default to `moment()` if `m === undefined`.

Droogans:
I honestly feel like the effort going into this PR would be better spent switching the datepicker to using strings for input/output data.

Why should we move from Date -> moment, if we're just going to do moment -> string later? I think this PR should be closed and reworked as `e2e-no-dates-just-moment`.

Droogans:
Reimplementing as string input -> string output, which is the ultimate goal of how date and time pickers will work soon.

PR #1590, opened 2016-03-07
fix(rxSearchbox): FRMW-714 support undefined and null value

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-714

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxSearchBox/scripts/rxSearchBox.js
+++ b/src/components/rxSearchBox/scripts/rxSearchBox.js
@@ -51,7 +51,7 @@ angular.module('encore.ui.rxSearchBox')
             $scope.rxPlaceholder = $scope.rxPlaceholder || 'Search...';
 
             $scope.$watch('searchVal', function (newVal) {
-                if ($scope.isDisabled) {
+                if (!newVal || $scope.isDisabled) {
                     $scope.isClearable = false;
                 } else {
                     $scope.isClearable = newVal.toString() !== '';
--- a/src/components/rxSearchBox/scripts/rxSearchBox.spec.js
+++ b/src/components/rxSearchBox/scripts/rxSearchBox.spec.js
@@ -1,12 +1,12 @@
 describe('rxSearchBox', function () {
-    var scope, compile, rootScope, el, template;
+    var scope, isoScope, compile, el, template;
 
     beforeEach(function () {
         module('encore.ui.rxSearchBox');
         module('templates/rxSearchBox.html');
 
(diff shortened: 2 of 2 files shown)


src/components/rxSearchBox/scripts/rxSearchBox.js
```diff
@@ -51,7 +51,7 @@ angular.module('encore.ui.rxSearchBox')
             $scope.rxPlaceholder = $scope.rxPlaceholder || 'Search...';
 
             $scope.$watch('searchVal', function (newVal) {
-                if ($scope.isDisabled) {
+                if (!newVal || $scope.isDisabled) {
                     $scope.isClearable = false;
                 } else {
                     $scope.isClearable = newVal.toString() !== '';
```
Droogans: This ought to be a function.

`</pedantry>`

CITguy: Which part? left or right side of assignment?

Droogans:
```js
$scope.isClearable = function () {
    return newVal !== undefined && newVal.toString() !== '';
}
```

I'm shooting off  the hip here -- that may be the wrong inner definition for `$scope.isClearable()` the function.

(this has nothing to do with your PR, I just noticed it.)

parlarjb:
LGTM

src/components/rxSearchBox/scripts/rxSearchBox.spec.js
```diff
@@ -14,73 +14,131 @@ describe('rxSearchBox', function () {
...
                 beforeEach(function () {
-                    scope.clearSearch();
+                    scope.testModel = 'hiyoo';
+                    scope.$apply();
                 });
 
-                it('should set searchVal to empty string', function () {
-                    expect(scope.searchVal).to.be.empty;
+                it('should not be clearable', function () {
+                    expect(isoScope.isClearable).to.be.true;
```
100stacks: Based on your `it` clause, should the `expect` clause be `false`?

100stacks:
Other than the one question...LGTM

src/components/rxSearchBox/scripts/rxSearchBox.spec.js
CITguy: good catch, description is incorrect. will update

PR #1592, opened 2016-03-09
fix(mentionBot): Only mention rackerlabs contributors

Droogans:
### LGTMs
- [ ] Dev LGTM

This will prevent @rnreekez, @klamping and @chriscantu from trolling us on PRs that touch core parts of our codebase.


chriscantu:
I wouldn't do that! 

devilish grin

PR #1594, opened 2016-03-11
fix(rxMultiSelect): FRMW-641 remove rxSelect dependency

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-641

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM
- [x] QE LGTM

---
- Refactor `rxMultiSelect` component to remove `rxSelect` dependency.
- Fix `z-index` context stacking bug.

### DIFF
--- a/src/components/rxMultiSelect/docs/rxMultiSelect.html
+++ b/src/components/rxMultiSelect/docs/rxMultiSelect.html
@@ -1,21 +1,93 @@
-<!-- Sample HTML goes here as a live example of how the component can be used -->
 <div ng-controller="rxMultiSelectCtrl">
-    <h3>rxMultiSelect Input</h3>
-    <p class="multi-select-container">
-        <span class="field-label">Classification:</span>
-        <rx-multi-select ng-model="data.classification" id="classification">
+  <h3>rxMultiSelect Input</h3>
+  <p class="multi-select-container">
+    <span class="field-label">Classification:</span>
+    <rx-multi-select ng-model="data.classification" id="classification">
+      <rx-select-option value="A">Type A</rx-select-option>
+      <rx-select-option value="B">Type B</rx-select-option>
+      <rx-select-option value="C">Type C</rx-select-option>
+      <rx-select-option value="D">Type D</rx-select-option>
+    </rx-multi-select>
+  </p>
+  <span>Model value: [{{ data.classification.join(', ') }}]</span>
+
+  <p class="multi-select-container">
+    <span class="field-label">Disabled rxMultiSelect:</span>
+    <rx-multi-select ng-model="data.other" ng-disabled="true">
+      <rx-select-option value="Not allowed"></rx-select-option>
+    </rx-multi-select>
+  </p>
+  <br />
+
+  <h3>rxMultiSelect States</h3>
+  <p>
+    Below you'll find examples of how <code>rxMultiSelect</code> will appear in different states.
+  </p>
+
+  <table>
+    <thead>
+      <tr>
+        <th></th>
+        <th>Enabled</th>
+        <th>Disabled</th>
+      </tr>
+    </thead>
+    <tbody>
+      <!-- Valid States -->
+      <tr>
+        <th>Valid</th>
+        <!-- Valid Enabled -->
+        <td>
+          <rx-multi-select
+            id="msValidEnabled"
+            ng-model="validEnabled">
+
             <rx-select-option value="A">Type A</rx-select-option>
             <rx-select-option value="B">Type B</rx-select-option>
             <rx-select-option value="C">Type C</rx-select-option>
             <rx-select-option value="D">Type D</rx-select-option>
-        </rx-multi-select>
-    </p>
-    <span>Model value: [{{ data.classification.join(', ') }}]</span>
+          </rx-multi-select>
+        </td>
+        <!-- Valid Disabled -->
+        <td>
+          <rx-multi-select
+            id="msValidDisabled"
+            ng-model="validDisabled"
+            ng-disabled="true">
 
-    <p class="multi-select-container">
-        <span class="field-label">Disabled rxMultiSelect:</span>
-        <rx-multi-select ng-model="data.other" disabled>
-            <rx-select-option value="Not allowed"></rx-select-option>
-        </rx-multi-select>
-    </p>
+            <rx-select-option value="Not allowed">Not Allowed</rx-select-option>
+          </rx-multi-select>
+        </td>
+      </tr>
+      <!-- Invalid States -->
+      <tr>
+        <th>Invalid</th>
+        <!-- Invalid Enabled -->
+        <td>
+          <rx-multi-select
+            id="msInvalidEnabled"
+            ng-model="invalidEnabled"
+            always-invalid>
+
+            <rx-select-option value="A">Type A</rx-select-option>
+            <rx-select-option value="B">Type B</rx-select-option>
+            <rx-select-option value="C">Type C</rx-select-option>
+            <rx-select-option value="D">Type D</rx-select-option>
+          </rx-multi-select>
+        </td>
+
+        <!-- Invalid Disabled -->
+        <td>
+          <rx-multi-select
+            id="msInvalidDisabled"
+            ng-model="invalidDisabled"
+            always-invalid
+            ng-disabled="true">
+
+            <rx-select-option value="Not allowed">Not Allowed</rx-select-option>
+          </rx-multi-select>
+        </td>
+      </tr>
+    </tbody>
+  </table>
 </div>
--- a/src/components/rxMultiSelect/rxMultiSelect.less
+++ b/src/components/rxMultiSelect/rxMultiSelect.less
@@ -2,33 +2,104 @@
  * rxMultiSelect
  */
 rx-multi-select {
+  #vars.rxMultiSelect();
   z-index: auto;
+  box-sizing: border-box;
+  display: block;
+  color: @container-list-text-color;
+  background: @container-background-color;
+  position: relative;
 
-  &[disabled] .preview {
-    pointer-events: none;
+  // Element should be resizable via wrapper alone
+  min-width: @container-min-width;
+
+  border: 1px solid @container-border-color;
+
+  .control {
+    box-sizing: border-box;
+    display: flex;
+    flex-flow: row no-wrap;
+    justify-content: space-around;
+    align-items: stretch;
+    height: @container-height;
+    cursor: pointer;
+
+    .preview {
+      flex: 1 1;
+      display: flex;
+      align-items: center;
+      justify-content: flex-start;
+      margin: 0 5px;
+      overflow: hidden;
+      white-space: nowrap;
+    }
+
+    .select-trigger {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      flex: 0 0 @trigger-width;
+      color: @trigger-color;
+    }
   }
 
-  .preview {
-    padding: 5px;
-    overflow: hidden;
-    white-space: nowrap;
-  }
-
-  ul {
+  .menu {
     z-index: 100;
     position: absolute;
+    cursor: pointer;
     top: -1px;
     left: -1px;
     width: 100%;
-    border: 1px solid @rxForm-input-border-color;
-    background: @rxForm-input-background-color;
+    border: 1px solid @container-border-color;
+    background: @container-background-color;
     box-shadow: 0 3px 15px 0 rgba(0, 0, 0, 0.35);
   }
 
-  rx-select-option[value="all"] label {
-    border-bottom: 1px solid @rxForm-input-border-color;
+  .disabled() {
+    color: @container-list-disabled-text-color;
+    cursor: not-allowed ;
+    border-color: @container-disabled-border-color;
+    border-width: @container-border-width;
+    background: @container-disabled-background-color;
+
+    .select-trigger {
+      background: @trigger-disabled-background-color;
+      border-left-color: @trigger-disabled-color;
+      color: @trigger-disabled-color;
+    }
+  }//.disabled()
+
+  &.ng-invalid.ng-dirty {
+    border-color: @container-invalid-border-color;
+    border-width: 2px;
+
+    .select-trigger {
+      color: @trigger-invalid-color;
+    }
+
+    &[disabled] {
+      .disabled();
+    }
+  }//.ng-invalid.ng-dirty
+
+  &[disabled] {
+    .disabled();
   }
-}
+
+  rx-select-option[value="all"] label {
+    border-bottom: 1px solid @container-border-color;
+  }
+
+  .backdrop {
+    z-index: 90;
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: transparent;
+  }
+}//rx-multi-select
 
 .rx-select-option {
   label {
@@ -46,4 +117,4 @@ rx-multi-select {
     height: @app-font-size;
     vertical-align: middle;
   }
-}
+}//.rx-select-option
--- a/src/components/rxMultiSelect/rxMultiSelect.module.js
+++ b/src/components/rxMultiSelect/rxMultiSelect.module.js
@@ -6,16 +6,12 @@
  * This component is a multi-select dropdown with checkboxes for each option.
  * It is a replacement for `<select multiple>` when space is an issue, such as
  * in the header of a table.
- * ## Services
- * * Links to service APIs provided by rxMultiSelect component.
  *
  * ## Directives
  * * {@link rxMultiSelect.directive:rxMultiSelect rxMultiSelect}
  * * {@link rxMultiSelect.directive:rxSelectOption rxSelectOption}
  *
- * ## Related Directives (if applicable)
- * * Links to directive APIs provided by other components.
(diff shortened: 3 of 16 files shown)


100stacks:
## `rxMultiSelect` Component 

### Demo Page
rxmultiselect - encoreui 2016-03-11 14-45-38

### API Docs
screen shot 2016-03-11 at 2 46 00 pm

encore ui developer documentation rxmultiselect 2016-03-11 14-47-27

encore ui developer documentation rxselectoption 2016-03-11 14-49-46

src/components/rxMultiSelect/docs/rxMultiSelect.html
```diff
@@ -1,21 +1,95 @@
-<!-- Sample HTML goes here as a live example of how the component can be used -->
 <div ng-controller="rxMultiSelectCtrl">
+  <div class="demoContainer">
```
CITguy: What's the purpose of this div?

src/components/rxMultiSelect/rxMultiSelect.module.js
```diff
@@ -6,16 +6,12 @@
...
  * ## Directives
  * * {@link rxMultiSelect.directive:rxMultiSelect rxMultiSelect}
  * * {@link rxMultiSelect.directive:rxSelectOption rxSelectOption}
  *
- * ## Related Directives (if applicable)
- * * Links to directive APIs provided by other components.
  */
 angular.module('encore.ui.rxMultiSelect', [
-    'encore.ui.rxSelectFilter'
+    'encore.ui.quarks'
```
CITguy: I'm almost certain we can do away with this dependency.

src/components/rxMultiSelect/scripts/rxSelectOption.js
```diff
@@ -3,6 +3,7 @@ angular.module('encore.ui.rxMultiSelect')
  * @ngdoc directive
  * @name rxMultiSelect.directive:rxSelectOption
  * @restrict E
+ * @requires rxMultiSelect.directive:rxMultiSelect
```
CITguy: Please revert this change. The `rxSelectOption` does not use `rxMultiSelect` in its template.

src/molecules/molecules.module.js
```diff
@@ -9,6 +9,7 @@
  *
  * ## Directives
  * * {@link molecules.directive:rxDatePicker rxDatePicker}
+ * * {@link molecules.directive:rxTimePicker rxTimePicker}
```
CITguy: Thanks! Forgot about that. :)

src/components/rxMultiSelect/scripts/rxMultiSelect.exercise.js
```diff
@@ -38,52 +42,70 @@ exports.rxMultiSelect = function (options) {
...
+                component.deselect(['Select All']);
+                expect(component.selectedOptions).to.eventually.be.empty;
+                expect(component.preview).to.eventually.equal('None');
+            });
+
+            it('should hide the menu when backdrop is clicked', function () {
+                component.rootElement.$('.backdrop').click();
+                expect(component.isOpen()).to.eventually.be.false;
+            });
+        }//if options.disabled
```
Droogans: @100stacks Thanks for switching all of these to `it('should ...')` style test names. It's our preferred standard (it helps our static test analyzer return better results for our coverage reports).

src/components/rxMultiSelect/docs/rxMultiSelect.html
100stacks: Ok, will remove.  I was following this pattern https://github.com/rackerlabs/encore-ui/blob/master/src/components/rxSelect/docs/rxSelect.html#L2

src/components/rxMultiSelect/rxMultiSelect.module.js
100stacks: @CITguy, we need this to pull in the `rxDOMHelper` quark: https://github.com/100stacks/encore-ui/blob/FRMW-641-z-index-fix-rxMultiSelect/src/components/rxMultiSelect/scripts/rxMultiSelect.js#L44.

src/components/rxMultiSelect/scripts/rxMultiSelect.exercise.js
100stacks: @Droogans, refactored based on @CITguy's recommendations.  Thanks for the explanation on purpose...I'll make sure to look out for this in future refactoring.  :wink:

src/components/rxMultiSelect/scripts/rxSelectOption.js
100stacks: This should be `@requires atoms.directive:rxCheckbox`.  Thanks for the catch. :wink:

src/components/rxMultiSelect/rxMultiSelect.module.js
CITguy: Ok, carry on. :+1:

100stacks:
Updated `rxSelectOption` API Docs in https://github.com/rackerlabs/encore-ui/pull/1594#issuecomment-195562324 to reflect `rxCheckbox` dependency.

glynnis:
Design LGTM

src/styles/vars.less
```diff
@@ -223,6 +223,28 @@
...
+    @container-height: 26px;
+    @container-background-color: @white;
+    @container-border-color: @gray-500;
+    @container-disabled-background-color: @app-disabled-background-color;
+    @container-invalid-border-color: @red-500;
+    @container-disabled-border-color: @container-border-color;
+    @container-border-width: 1px;
+    // List Items
+    @container-list-text-color: @app-body-text-color;
+    @container-list-disabled-text-color: @app-disabled-text-color;
```
glynnis: I don't think you need `@container-list-disabled-text-color` or `@container-list-text-color` -- I think you can just use `@app-body-text-color` and `@app-disabled-text-color` directly in the component CSS without the additional level of abstraction. @ssalinas24 would you agree?

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
+    background: @container-background-color;
     box-shadow: 0 3px 15px 0 rgba(0, 0, 0, 0.35);
   }
 
+  .disabled() {
+    color: @container-list-disabled-text-color;
+    cursor: not-allowed ;
+    border-color: @container-disabled-border-color;
+    border-width: @container-border-width;
+    background: @container-disabled-background-color;
```
glynnis: I would just use `@app-disabled-background-color` directly here.

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
     width: 100%;
-    border: 1px solid @rxForm-input-border-color;
-    background: @rxForm-input-background-color;
+    border: 1px solid @container-border-color;
+    background: @container-background-color;
     box-shadow: 0 3px 15px 0 rgba(0, 0, 0, 0.35);
   }
 
+  .disabled() {
+    color: @container-list-disabled-text-color;
```
glynnis: I would just use `@app-disabled-text-color` directly here.

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
+
+    .select-trigger {
+      background: @trigger-disabled-background-color;
+      border-left-color: @trigger-disabled-color;
+      color: @trigger-disabled-color;
+    }
+  }//.disabled()
+
+  &.ng-invalid.ng-dirty {
+    border-color: @container-invalid-border-color;
```
glynnis: I think you can just use `@rxForm-input-border-color-invalid` here, since this is a form element.

src/styles/vars.less
```diff
@@ -223,6 +223,28 @@
     @invalid-control-border-color: @red-500;
     @invalid-control-icon-color: @red-500;
   }//.rxTimePicker
+
+  .rxMultiSelect {
+    // Container
+    @container-min-width: 60px;
+    @container-height: 26px;
+    @container-background-color: @white;
```
glynnis: I would just use the rxForm vars that already exist instead of creating additional vars here. For example, `@rxForm-input-background-color`. See others here: https://github.com/rackerlabs/encore-ui/blob/befe3007826acb4aa3fb6b8ac650cda68ec3613f/src/styles/vars.less#L256

src/styles/vars.less
```diff
@@ -223,6 +223,28 @@
...
+    @container-border-color: @gray-500;
+    @container-disabled-background-color: @app-disabled-background-color;
+    @container-invalid-border-color: @red-500;
```
glynnis: I would use `@rxForm-error-accent-color` instead, rather than define a new var here.

src/styles/vars.less
```diff
@@ -223,6 +223,28 @@
...
+    @container-disabled-border-color: @container-border-color;
+    @container-border-width: 1px;
+    // List Items
+    @container-list-text-color: @app-body-text-color;
+    @container-list-disabled-text-color: @app-disabled-text-color;
+    // Trigger
```
glynnis: There are likely other existing variables for rxForm that would be useful here. See https://github.com/rackerlabs/encore-ui/blob/befe3007826acb4aa3fb6b8ac650cda68ec3613f/src/styles/vars.less#L256

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
+    border-width: 2px;
+
+    .select-trigger {
+      color: @trigger-invalid-color;
```
glynnis: See rxForm vars here: https://github.com/rackerlabs/encore-ui/blob/befe3007826acb4aa3fb6b8ac650cda68ec3613f/src/styles/vars.less#L256

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
+    color: @container-list-disabled-text-color;
+    cursor: not-allowed ;
+    border-color: @container-disabled-border-color;
+    border-width: @container-border-width;
+    background: @container-disabled-background-color;
+
+    .select-trigger {
+      background: @trigger-disabled-background-color;
+      border-left-color: @trigger-disabled-color;
+      color: @trigger-disabled-color;
```
glynnis: See rxForm vars here: https://github.com/rackerlabs/encore-ui/blob/befe3007826acb4aa3fb6b8ac650cda68ec3613f/src/styles/vars.less#L256

src/styles/vars.less
ssalinas24: It could go either way actually. Our guidelines we talked about allow those `@app` to be included directly in each component .less file like Glynnis is mentioning. One good thing about line 238, and line 239 is the flexibility that it offers in the long run. It allows the color to be changed right in the vars.less file. However, maybe we should only leave line 238 as is. @glynnis am i making sense?

src/components/rxMultiSelect/rxMultiSelect.less
CITguy: Two things:

1. `@rxForm-input-border-color-invalid` (because of the prefix) is namespaced for the `rxForm` styles, and I would like to avoid/limit cross-component variable usage.  If you'd like to use a global variable for control styling, I'd suggest `@app-input-border-color-invalid` (out of scope for this PR).
2. Because of the Less vars namespacing technique ([`#vars.rxDatePicker`](https://github.com/rackerlabs/encore-ui/blob/master/src/styles/vars.less#L149), [`#vars.rxTimePicker`](https://github.com/rackerlabs/encore-ui/blob/master/src/styles/vars.less#L199), etc.), it's ~~challenging~~ bloody difficult to include another component's variables while not clobbering existing variables.  If you're sharing variables/values, that's a good sign that they should probably be global.

We should sync up to discuss strategies around the use of global vars.

100stacks: Ok @glynnis @ssalinas24 @CITguy, we'll sync up to discuss preferred `Global Vars` naming convention(s).  :+1:

src/styles/vars.less
ssalinas24: The following story also exists to clean up rxForm vars. https://jira.rax.io/browse/FRMW-608

src/components/rxMultiSelect/rxMultiSelect.less
ssalinas24: I agree with @CITguy. When i was last working on global vars, I believe we had discussed that the Inputs were high potential for being global app vars.

100stacks:
Removed the `on hold` and `needs discussion` labels based on our LESS namespaces discussion on Friday.

CITguy:
screen shot 2016-03-22 at 3 21 48 pm

Enabled dropdown icon should match border color.

src/components/rxMultiSelect/rxMultiSelect.less
```diff
@@ -2,33 +2,102 @@
...
  */
 rx-multi-select {
+  #vars.rxMultiSelect();
   z-index: auto;
+  box-sizing: border-box;
+  display: block;
+  color: @container-list-text-color;
+  background: @container-background-color;
+  position: relative;
+  cursor: pointer;
```
CITguy: This should be removed from the `rx-multi-select` block and placed within both the `.control` and `.menu` blocks.

CITguy: This is showing a "pointer" when you hover over the transparent backdrop.

100stacks:
@CITguy updated the trigger color...great :eyes:

screen shot 2016-03-22 at 4 16 28 pm

100stacks:
@CITguy, updated the cursor to change to a `pointer`when outside the selection box:

rxmultiselect-cursor-demo

glynnis:
Per discussions we had as a group and JIRA stories created to capture additional refactor/vars work that needs to be done (https://jira.rax.io/browse/FRMW-592 so far, but there is more to be captured), this work LGTM! We will address other issues from comments separately.

@CITguy can you be sure that the work with namespacing gets captured in the JIRA epic linked here ^? A quick search through the backlog came up dry, I think because it's reflected in the 2.0 etherpad but not JIRA yet, though it's possible I just didn't search with the right terms.

CITguy:
@glynnis: https://jira.rax.io/browse/FRMW-753 created to capture documenting variable usage.

src/components/rxMultiSelect/scripts/rxMultiSelect.spec.js
```diff
@@ -61,10 +59,19 @@ describe('rxMultiSelect', function () {
...
             it('does not toggle the visibility of the menu when a child element is clicked', function () {
-                angular.element(el[0].querySelector('rx-select-option')).click();
+                angular.element(el[0].querySelector('.menu')).click();
                 expect(isolateScope.listDisplayed).to.be.false;
             });
 
+            it('should close an open menu', function () {
+                var previewElement = angular.element(el[0].querySelector('.preview'));
+                previewElement.click();
+                expect(isolateScope.listDisplayed).to.be.true;
```
CITguy: We can simplify the setup as follows:

```javascript
isolateScope.listDisplayed = true;
```

Considering that we're testing the `closeMenu()` functionality, we only need to test if `isolateScope.listDisplayed` changes. Clicking the preview to make `listDisplayed` equal true is inefficient, just set it directly.

100stacks:
@CITguy, appreciate the incorporated feedback. :+1:

CITguy:
LGTM

src/components/rxMultiSelect/scripts/rxMultiSelect.js
```diff
@@ -39,15 +41,19 @@ angular.module('encore.ui.rxMultiSelect')
...
         transclude: true,
-        require: ['rxMultiSelect', 'ngModel'],
+        require: [
+            'rxMultiSelect',
+            'ngModel'
+        ],
         scope: {
             selected: '=ngModel',
             options: '=?',
+            isDisabled: '=ngDisabled',
```
halkon: So I'm not opposed to using `isDisabled` locally, and it does read better, I just have a fear that it may cause some confusion later on, was the reason just readability?

src/components/rxMultiSelect/scripts/rxMultiSelect.js
```diff
@@ -90,32 +96,18 @@ angular.module('encore.ui.rxMultiSelect')
...
-            scope.$on('$destroy', function () {
-                $document.off('click', documentClickHandler);
-            });
-
             scope.listDisplayed = false;
 
-            scope.toggleDisplay = function (event) {
-                if (event.target === previewElement) {
+            scope.toggleMenu = function () {
+                if (!scope.isDisabled) {
```
halkon: :+1:

src/components/rxMultiSelect/scripts/rxMultiSelect.js
CITguy: The reasoning was readability, yes. If we find a better way of setting a local boolean in the future, then we won't have to rename the variable.

halkon:
Dev LGTM, but, there are many other comments and some rebasing :wink: that need some attention

100stacks:
@halkon, thanks for the feedback! :+1:

PR #1598, opened 2016-03-18
test(rxToggleSwitch): Realign page object API to reduce confusion

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-706

### LGTMs
- [x] Dev LGTM
- [x] Dev/QE LGTM

This takes the confusing `isEnabled`/`isDisabled` functions and renames them to the more
reasonable `isOn` and `isOff`, respectively. This replaces `isEnabled`/`isDisabled` with
checks on the toggle switch button to ensure that it is not currently "greyed out" (disabled).

FRMW-706 #in-review

### DIFF
--- a/src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
+++ b/src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
@@ -14,7 +14,7 @@ var rxToggleSwitch = {
     /**
      * @function
      * @instance
-     * @description Whether the root element is currently displayed.
+     * @description Whether the toggle switch is currently displayed.
      * @returns {Boolean}
      */
     isDisplayed: {
@@ -26,14 +26,12 @@ var rxToggleSwitch = {
     /**
      * @function
      * @instance
-     * @description Whether or not the switch component is currently set to the "on" position.
+     * @description Whether the toggle switch has interaction enabled.
      * @returns {Boolean}
      */
     isEnabled: {
         value: function () {
-            return this.btnToggleSwitch.getAttribute('class').then(function (classes) {
-                return classes.split(' ').indexOf('on') > -1;
-            });
+            return this.btnToggleSwitch.getAttribute('disabled').then(_.isNull);
         }
     },
 
@@ -52,28 +50,13 @@ var rxToggleSwitch = {
      *     expect(switch.isEnabled()).to.eventually.be.true;
      * });
      */
-    enable: {
+    toggleOn: {
         value: function () {
             var page = this;
-            return this.isDisabled().then(function (disabled) {
-                if (disabled) {
+            return this.isToggled().then(function (toggled) {
+                if (!toggled) {
                     page.btnToggleSwitch.click();
                 }
-            });
-        }
-    },
-
-    /**
-     * @todo Rename this function. This sounds like it checks for `ng-disable` on the directive.
-     * @function
-     * @instance
-     * @description Whether or not the switch component is currently set to the "off" position.
-     * @returns {Boolean}
-     */
-    isDisabled: {
-        value: function () {
-            return this.isEnabled().then(function (enabled) {
-                return !enabled;
             });
         }
     },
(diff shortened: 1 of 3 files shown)


Droogans:
`grunt karma:minified` is still failing. I'll try and get that addressed here and now in this PR.

ErnieAtLYD:
Big fan of this one! :D

100stacks:
Will these changes accommodate failed asynchronous operations, where we represent an `ON` => `OFF` state || `OFF` => `ON` state with the toggle switch temporary disabled?

Probably not an issue...
https://github.com/rackerlabs/encore-ui/blob/12a9710025d994814e07bc0c22d4eed9059b2987/src/components/rxToggleSwitch/docs/rxToggleSwitch.midway.js#L18-L21

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
```diff
@@ -26,14 +26,12 @@ var rxToggleSwitch = {
     /**
      * @function
      * @instance
-     * @description Whether or not the switch component is currently set to the "on" position.
+     * @description Whether the toggle switch is currently enabled.
```
CITguy: maybe `Whether the toggle switch has interaction enabled.`

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
```diff
@@ -64,17 +62,14 @@ var rxToggleSwitch = {
     },
 
     /**
-     * @todo Rename this function. This sounds like it checks for `ng-disable` on the directive.
      * @function
      * @instance
-     * @description Whether or not the switch component is currently set to the "off" position.
+     * @description Whether the toggle switch is currently disabled.
```
CITguy: maybe `Whether the toggle switch has interaction disabled.`

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
```diff
@@ -105,6 +100,34 @@ var rxToggleSwitch = {
...
+     */
+    isOff: {
+        value: function () {
+            return this.text.then(function (text) {
+                return text === 'OFF';
+            });
+        }
+    },
+
+    /**
```
CITguy: I feel like these two predicate functions could be combined into `isToggled()` (True for ON, False for OFF)

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
```diff
@@ -96,7 +91,7 @@ var rxToggleSwitch = {
     disable: {
```
CITguy: This action no longer makes sense with the other changes made in this PR. I'd suggest changing to `toggleOff()`.

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
```diff
@@ -55,7 +53,7 @@ var rxToggleSwitch = {
     enable: {
```
CITguy: This action no longer makes sense with the other changes made in this PR. I'd suggest changing to `toggleOn()`.

CITguy:
Changes will need made to the [exercise](https://github.com/rackerlabs/encore-ui/blob/master/src/components/rxToggleSwitch/scripts/rxToggleSwitch.exercise.js#L28-L29) to reflect updates to the page object verbiage.

I'd suggest updating the following exercise options.
* `enabledAtStart` -> `toggledAtStart`
* `enabledAtEnd` -> `toggledAtEnd`

src/components/rxToggleSwitch/scripts/rxToggleSwitch.page.js
Droogans: Diff isn't outdated UI, but I did do this.

Droogans:
@CITguy Shouldn't I delete `isEnabled()`? `isDisabled()` will just return `false` if it's enabled.

Droogans:
Updated Wiki: https://github.com/rackerlabs/encore-ui/wiki/2.0-Migration-Guide#updated-encorerxtoggleswitch

CITguy:
@Droogans: yeah, go ahead and nix `isEnabled()`

Droogans:
I decided to keep `isEnabled()`, since thinking that something should not be in the negative state is confusing.

I will be removing all "negative" properties in 2.x, preferring instead to use a positive-only API to reduce the amount of testing that needs to occur in our end to end test suite.

```js
expect(component.isDisabled()).to.eventually.be.false;
```

is better expressed in the positive case,

```js
expect(component.isEnabled()).to.eventually.be.true;
```

100stacks:
@Droogans, what do you think about:

```js
expect(component.isToggleEnabled()).to.eventually.be.true;
```

Just a :thought_balloon:  ....the overall logic is cleaner. LGTM :+1:

CITguy:
LGTM

CITguy:
@Droogans, please rebase.

PR #1632, opened 2016-04-22
fix(demo): FRMW-812 rxDatePicker use local time zone

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-812

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

---
Updated the demo and midway tests to use local time zone versus the current default UTC.

### DIFF
--- a/src/molecules/rxDatePicker/docs/examples/rxDatePicker.simple.js
+++ b/src/molecules/rxDatePicker/docs/examples/rxDatePicker.simple.js
@@ -1,4 +1,4 @@
 angular.module('demoApp')
 .controller('rxDatePickerSimpleCtrl', function ($scope) {
-    $scope.dateModel = new Date().toISOString().split('T')[0];
+    $scope.dateModel = moment(new Date()).format('YYYY-MM-DD');
 });
--- a/src/molecules/rxDatePicker/docs/rxDatePicker.midway.js
+++ b/src/molecules/rxDatePicker/docs/rxDatePicker.midway.js
@@ -11,7 +11,7 @@ describe('rxDatePicker', function () {
         var isoFormat = 'YYYY-MM-DD';
         var formatMonth = 'MMM';
         var formatYear = 'YYYY';
-        var today = moment(new Date());
+        var today = moment(new Date()).format(isoFormat);
         var month = moment(today).clone().format(formatMonth);
         var year = moment(today).clone().format(formatYear);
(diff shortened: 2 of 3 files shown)


src/molecules/rxDatePicker/docs/examples/rxDatePicker.simple.js
```diff
@@ -1,4 +1,4 @@
 angular.module('demoApp')
 .controller('rxDatePickerSimpleCtrl', function ($scope) {
-    $scope.dateModel = new Date().toISOString().split('T')[0];
+    $scope.dateModel = moment(new Date()).format('YYYY-MM-DD', 'LL');//use local time zone
```
CITguy: What's the `LL` parameter for?

100stacks: It's used for setting the `moment` object to the localized date.

CITguy: The `LL` format would generate `April 22 2016`. The model value should be in the format of `YYYY-MM-DD`. The directive will care of converting the model value to the appropriate display format (`MMMM DD, YYYY`).  

Please remove the `LL` argument. I do not believe it is necessary.

src/molecules/rxDatePicker/docs/rxDatePicker.midway.js
```diff
@@ -11,7 +11,7 @@ describe('rxDatePicker', function () {
         var isoFormat = 'YYYY-MM-DD';
         var formatMonth = 'MMM';
         var formatYear = 'YYYY';
-        var today = moment(new Date());
+        var today = moment(new Date()).format(isoFormat, 'LL');//use local time zone
```
CITguy: Please remove the `LL` argument.

100stacks:
@CITguy, ok I removed `LL` argument usage from the update.

CITguy:
LGTM

CITguy:
FYI: I'm going to trigger Travis after 6PM to verify.

100stacks:
@CITguy, when I tested the patch locally, I switched between different times of day and time zones to try to replicate the 6PM to Midnight UTC differential.  I think that's why I originally used the `LL` argument to explicitly use local time over UTC. ` ¯\_(ツ)_/¯ `

CITguy:
Looks like Travis agrees. 👍

parlarjb:
LGTM

PR #1640, opened 2016-05-02
feat(formatting): Add image examples

glynnis:
JIRA: https://jira.rax.io/browse/FRMW-567

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] Design LGTM

## Before
screen shot 2016-05-02 at 4 54 24 pm

## After
[Outdated screenshot](https://cloud.githubusercontent.com/assets/1529366/14969180/aeb2b4e8-1086-11e6-904a-b351da5a7b46.png)

screen shot 2016-05-05 at 4 01 29 pm


glynnis:
@lfielder @ssalinas24 @ElementE Can you let me know what you think of this? Right now the borders are defined in CSS as `@green-500`, `@orange-500`, and `@red-500`. Is that right? Or would we use the `@orange-700` and `@red-700` values instead?

I also followed the text styles that were in the mock-up on the JIRA story, but I'd like your feedback about the "Avoid" and "Exception" titles. Are they in the correct style, or should we make them header elements? Is that the language we want to use? Should there be more copy? Compare with some other similar examples in stories like https://jira.rax.io/browse/FRMW-794, which have the copy "DO" and "DON'T" to the left of the examples.

Please weigh in and let me know if you think I should make updates on the design side.

ElementE:
@glynnis 
I think it works, I like the brighter values for these - they help call out/highlight to the user that these aren't just part of a design. 

For the titles, I think either verbiage works. I do think the titles should be more integrated into the demos. If not along the left (like in the JIRA story linked) then maybe in the upper left corner, though not sure then if we need to color the text or make it distinguishable somehow to reinforce its not part of the demo.

halkon:
LGTM

grunt-tasks/options/copy.js
```diff
@@ -103,6 +103,7 @@ module.exports = {
             src: [
                 'css/*.css',
                 'images/**/*',
+                'templates/**/*.png',
```
parlarjb: @CITguy We were a bit back and forth on this one. We do have a top-level `images/` directory available for the demo, so we don't strictly need this. I didn't have a strong feeling on the "right way"

100stacks:
@glynnis, we are currently generating ISO 8601 Date strings - `YYYY-MM-DD` - where we are representing the numerical day and month with a leading zero `0` in the output.  I defer to @CITguy @parlarjb on stripping the leading zero out when displaying the dates.

grunt-tasks/options/copy.js
CITguy: This glob is a bit too specific with regards to file type.  Additionally, this promotes disorganization in the file structure with multiple locations for images.  I'd place all documentation images in `demo/images` to simplify file globbing and provide a single location for images.

glynnis: Backed this change out and moved images to the `demo/images` directory.

CITguy:
I just noticed that there's a spelling error in the exception image (orange). `instrucitons` should be `instructions`

demo/less/demoApp.less
```diff
@@ -142,3 +142,28 @@ p {
 .align-right {
   text-align: right;
 }
+
+.example-do,
+.example-exception,
+.example-avoid {
```
CITguy: I'd simplify to using `.example` instead of these three selectors. This will allow you to add a `.do`, `.avoid`, and `.exception` class in addition...

```less
.example {
  /* common example properties */

  &.do { /* ... */ }
  &.avoid { /* ... */ }
  &.exception { /* ... */ }
}
```

```html
<div class="example do"> ... </div>
<div class="example avoid"> ... </div>
<div class="example exception"> ... </div>
```

CITguy: This is actually how material design does its class names :)

CITguy:
Should the explanation text for the Avoid and Exception be part of the image? It would seem to make more sense to have that text directly in the markup.

ssalinas24:
@glynnis From what I see, this before and after looks good in relation to the JIRA story. Just let me know when you are done updating this PR and I will check back on it for an Approval.

demo/less/demoApp.less
glynnis: Done!

glynnis:
Moved titles inside the demo areas and updated the screenshots in the description of this issue.

glynnis:
@CITguy @parlarjb any response to @100stacks comment above re: leading zeroes?

ssalinas24:
Design LGTM

parlarjb:
@glynnis I like the consistency of keeping leading zeros present, but I guess it's a personal thing? I don't know if there's an "industry standard" for that.

CITguy:
The topic of leading zeros seems to head into Design territory, but personally I'd go with leading zeros. Additionally, we already have a story to discuss/define various date and time formats ([FRMW-817](https://jira.rax.io/browse/FRMW-817)), so further changes can be made in another PR. This PR is good to go.

LGTM

CITguy:
Merging. Failing tests are unrelated to changes in this PR.

PR #1641, opened 2016-05-03
feat(rxBytesConvert): convert from bytes

kevincareer427:
JIRA: https://jira.rax.io/browse/FRMW-432

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/utilities/rxBytesConvert/docs/examples/rxBytesConvert.demo.js
+++ b/src/utilities/rxBytesConvert/docs/examples/rxBytesConvert.demo.js
@@ -1,0 +1,6 @@
+angular.module('demoApp')
+.controller('rxBytesConvertCtrl', function ($scope) {
+    $scope.sizeGB = 42e10; // 420 GB
+    $scope.sizeTB = 125e12; // 125 TB
+    $scope.sizePB = 17134e13; // 171.34 PB
+});
--- a/src/utilities/rxBytesConvert/docs/rxBytesConvert.docs.html
+++ b/src/utilities/rxBytesConvert/docs/rxBytesConvert.docs.html
@@ -1,0 +1,5 @@
+<p>
+  Converts byte disk size into a more readable format (e.g. GBs, TBs, PBs)
+</p>
+
+<rx-example name="rxBytesConvert.demo"></rx-example>
--- a/src/utilities/rxBytesConvert/rxBytesConvert.meta.json
+++ b/src/utilities/rxBytesConvert/rxBytesConvert.meta.json
@@ -1,0 +1,7 @@
+{
+    "displayName": "rxBytesConvert",
+    "stability": "unstable",
+    "description": "Converts raw byte disk size into a more readable format",
+    "api": "filter:rxBytesConvert",
+    "keywords": []
+}
--- a/src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
+++ b/src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
@@ -1,0 +1,47 @@
+angular.module('encore.ui.utilities')
+
+/**
+ * @ngdoc filter
+ * @name utilities.filter:rxByteSize
+ * @description
+ *
+ * Converts Byte disk size into a more readable format (e.g. MBs, GBs, TBs, PBs)
+ *
+ *
+ * <pre>
+ * 1000 → 1 KB
+ * 12000000 → 12 MB
+ * </pre>
+ **/
+.filter('rxBytesConvert', function () {
+    return function (bytes, unit) {
+        var units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
+        var index;
+        
+        // check if unit is in the list of units
+        if (_.isUndefined(unit) || _.indexOf(units, unit.toUpperCase()) === -1) {
+            // determine closest unit
+            if (bytes > 0) {
+                index = Math.floor(Math.log(bytes) / Math.log(1000));
+                if (index > 5) {
+                    index = 5; // if the data is too large, default PB
+                }
+            } else {
+                index = 0;
+                bytes = 0;
+            }
+        } else {
+            index = _.indexOf(units, unit.toUpperCase());
+        }
+        
+        // calculate result in exected unit
+        var result = bytes / Math.pow(1000, index);
+        // check if result is integer (karma doesn't know Number.isInteger())
+        // https://github.com/ariya/phantomjs/issues/14014
+        if (result % 1 === 0) {
+            return result + ' ' + units[index];
+        } else {
+            return result.toFixed(2) + ' ' + units[index];
+        }
+    };
+});
--- a/src/utilities/rxBytesConvert/scripts/rxBytesConvert.spec.js
+++ b/src/utilities/rxBytesConvert/scripts/rxBytesConvert.spec.js
@@ -1,0 +1,99 @@
+describe('rxBytesConvert', function () {
+    var bytesize;
+    var bytes;
+    var unit;
+    beforeEach(function () {
+        module('encore.ui.utilities');
+
+        inject(function ($filter) {
+            bytesize = $filter('rxBytesConvert');
+        });
+    });
+    
+    it('should exist', function () {
+        expect(bytesize).to.exist;
+    });
+    
+    it('should convert 1000 to 1 KB', function () {
+        bytes = 1000;
+        expect(bytesize(bytes)).to.equal('1 KB');
+    });
+
+    it('should convert 1000000 to 1 MB', function () {
+        bytes = 1000000;
+        expect(bytesize(bytes)).to.equal('1 MB');
+    });
+    
+    it('should convert 1250000 to 1.25 MB', function () {
+        bytes = 1250000;
+        expect(bytesize(bytes)).to.equal('1.25 MB');
+    });
+    
+    it('should convert 1250000 to 1250 KB', function () {
+        bytes = 1250000;
+        unit = 'Kb';
+        expect(bytesize(bytes, unit)).to.equal('1250 KB');
+    });
+    
+    it('should convert 1250000 to 0.00 GB', function () {
+        bytes = 1250000;
+        unit = 'gB';
+        expect(bytesize(bytes, unit)).to.equal('0.00 GB');
+    });
+    
+    it('should convert 0 to 0 B', function () {
+        bytes = 0;
+        expect(bytesize(bytes)).to.equal('0 B');
+    });
+
+    it('should convert 0 with a MB unit to 0 MB', function () {
+        bytes = 0;
+        unit = 'MB';
+        expect(bytesize(bytes, unit)).to.equal('0 MB');
+    });
+
+    it('should convert 0 with a TB unit to 0 TB', function () {
+        bytes = 0;
+        unit = 'TB';
+        expect(bytesize(bytes, unit)).to.equal('0 TB');
+    });
+
+    it('should convert 0 with a PB unit to 0 PB', function () {
+        bytes = 0;
+        unit = 'PB';
+        expect(bytesize(bytes, unit)).to.equal('0 PB');
+    });
+
+    it('should convert undefined to 0 B', function () {
+        expect(bytesize()).to.equal('0 B');
+    });
+
+    it('should convert null to 0 B', function () {
+        bytes = null;
+        expect(bytesize(bytes)).to.equal('0 B');
+    });
+    
+    it('should convert null with a MB unit to 0 MB', function () {
+        bytes = null;
+        unit = 'MB';
+        expect(bytesize(bytes, unit)).to.equal('0 MB');
+    });
+    
+    it('should ignore the unit if the unit is not valid', function () {
+        bytes = 150000;
+        unit = 'XB';
+        expect(bytesize(bytes, unit)).to.equal('150 KB');
+    });
+    
+    it('should convert data to PB if data is too large when unit is not given', function () {
+        bytes = 12500000000000000000;
+        expect(bytesize(bytes)).to.equal('12500 PB');
+    });
+    
+    it('should convert data to PB if data is too large when unit is not valid', function () {
+        bytes = 256000000000000000000000;
+        unit = 'sB';
+        expect(bytesize(bytes, unit)).to.equal('256000000 PB');
+    });
+});
+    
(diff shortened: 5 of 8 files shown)


src/utilities/rxDiskSize/scripts/rxDiskSize.page.js
```diff
@@ -22,18 +22,6 @@ exports.rxDiskSize = {
...
-    /**
-     * @function
-     * @description A shorthand way of converting a bytes string to gigabytes.
-     * @param {String} rxDiskSizeString - The string to convert to gigabytes
-     * @returns {Number} The number of gigabytes represented by the human readable byte size string
-     * @example
-     * expect(encore.rxDiskSize.toGigabytes('1000 MB')).to.equal(1);
-     */
-    toGigabytes: function (rxDiskSizeString) {
-        return this.toBytes(rxDiskSizeString) / Math.pow(10, 9);
```
Droogans: As much I like that you deleted that code, it probably should go in the 2.x branch only.

I'm pretty certain this is another case of YANGI in the page objects.

src/utilities/rxDiskSize/scripts/rxDiskSize.js
```diff
@@ -29,4 +29,42 @@ angular.module('encore.ui.utilities')
...
+                index = Math.floor(Math.log(size) / Math.log(1000));
+            } else {
+                index = 0;
+                size = 0;
+            }
+        } else {
+            index = units.indexOf(unit.toUpperCase());
+        }
+        
+        var res = size / Math.pow(1000, index);
```
Droogans:
```js
index = units.indexOf(unit.toUpperCase());
```

This line, when coupled with

```js
var res = size / Math.pow(1000, index);
```

will be extremely wrong should a dev pass in a unit (by mistake) that doesn't exist.

```js
it('should ignore the unit if the unit is not valid', function () {
    var size = 150000;
    var unit = 'XB';
    expect(bytesize(size, unit)).to.equal('150 KB');
});
```

Right now, it returns `'150000000 undefined'`.

Droogans:
@kevincareer427 When you push up a fix for the issue I highlighted, please rebase this PR as well so that the datepicker fix gets in there and gets us that green checkmark from Travis.

src/utilities/rxDiskSize/scripts/rxDiskSize.js
CITguy: @Droogans: good point. That was my oversight when I suggested the change to the condition. I'll work with @kevincareer427 to get this updated.

100stacks:
Looks like you may need to revert the changes @Droogans mentioned in https://github.com/rackerlabs/encore-ui/pull/1641#discussion-diff-61956103.

src/utilities/rxBytesConvert/docs/rxBytesConvert.docs.html
```diff
@@ -0,0 +1,5 @@
+<p>
+  Converts Byte disk size into a more readable format (e.g. GBs, TBs, PBs)
```
Droogans: Byte is capitalized, it shouldn't be.

This could also be worded a little better, I think.

src/utilities/rxBytesConvert/rxBytesConvert.meta.json
```diff
@@ -0,0 +1,7 @@
+{
+    "displayName": "rxBytesConvert",
+    "stability": "stable",
```
Droogans: Should we be releasing stable components on the first commit?

I'd like to see it eventually reach "stable" after a few releases in the wild.

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Converts Byte disk size into a more readable format (e.g. MBs, GBs, TBs, PBs)
+ *
+ *
+ * <pre>
+ * 1000 → 1 KB
+ * 12000000 → 12 MB
+ * </pre>
+ **/
+.filter('rxBytesConvert', function () {
```
Droogans: This is the second definition of the same filter object.

You'll need to delete one of them. I suggest keeping the one that is in it's own component directory, away from `rxDiskSize`.

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
```diff
@@ -0,0 +1,39 @@
...
+    return function (size, unit) {
+        var units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
+        var index;
+        
+        if (_.isUndefined(unit)) {
+            if (size > 0) {
+                index = Math.floor(Math.log(size) / Math.log(1000));
```
Droogans: I had to research this because I liked it so much.

Turns out this works too.

```js
(Math.floor(Math.log10(size)) / 3) - 1
```

Great work. May want to include this link as a comment for folks like me who want to know how this works: https://en.wikipedia.org/wiki/Order_of_magnitude#Uses is what I used.

kevincareer427:
@Droogans thanks for the review advice! I'm working on it.

src/utilities/rxBytesConvert/rxBytesConvert.meta.json
```diff
@@ -0,0 +1,7 @@
+{
+    "displayName": "rxBytesConvert",
+    "stability": "unstable",
```
CITguy: 👍

src/utilities/rxDiskSize/scripts/rxDiskSize.page.js
```diff
@@ -22,7 +22,7 @@ exports.rxDiskSize = {
             'P': Math.pow(10, 15)
         }[parts[1].toUpperCase()[0]];
         return size * magnitude;
-    },
+        },
```
CITguy: @kevincareer427, please remove indent.

src/utilities/rxBytesConvert/docs/examples/rxBytesConvert.demo.js
```diff
@@ -0,0 +1,6 @@
+angular.module('demoApp')
+.controller('rxBytesConvertCtrl', function ($scope) {
+    $scope.sizeGB = 420 * Math.pow(10, 9);
+    $scope.sizeTB = 125000 * Math.pow(10, 9);
+    $scope.sizePB = 171340000 * Math.pow(10, 9);
```
CITguy: So, we can actually simplify these values...

```javascript
$scope.sizeGB = 42e10;
$scope.sizeTB = 125e12;
$scope.sizePB = 17134e13;
```

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
```diff
@@ -0,0 +1,42 @@
...
+            } else {
+                index = 0;
+                size = 0;
+            }
+        } else {
+            index = _.indexOf(units, unit.toUpperCase());
+        }
+        
+        var res = size / Math.pow(1000, index);
+        if (res % 1 === 0) {
```
Droogans: Everything is divisible by one. Was this changed from something else, and left behind by mistake?

The only edge case here is if `res` is less than zero, but there's better ways of checking if a number is less than zero.

src/utilities/rxBytesConvert/docs/examples/rxBytesConvert.demo.js
100stacks: Is it easier to to read 
> $scope.sizeGB = 420 * Math.pow(10, 9);

or 

> $scope.sizeGB = 42e10;

If we go with the simplified version, a comment might help.  💭

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
kevincareer427: @Droogans actually that's for checking if it's an integer, I thought it might be not good to leave and integer to two decimal

Droogans: Ok. Use `_.isNumber` instead.

kevincareer427: @Droogans but _.isNumber is not for checking it's an integer or float. Number.isInteger will work

Droogans: > Number.isInteger will work

I guess we meant, "*should work*": https://github.com/ariya/phantomjs/issues/14014 

😢

Droogans:
@kevincareer427 are you able to have a vidyo code walkthrough with @100stacks or @parlarjb today?

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
```diff
@@ -0,0 +1,47 @@
...
+ * 12000000 → 12 MB
+ * </pre>
+ **/
+.filter('rxBytesConvert', function () {
+    return function (bytes, unit) {
+        var units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
+        var index;
+        
+        // check if unit is in the list of units
+        if (_.isUndefined(unit) || _.indexOf(units, unit.toUpperCase()) === -1) {
```
halkon: So I kind of feel iffy about checking for undefined, and then checking for the uppercase of unit, and then on the else, doing an indexOf again without really knowing that `unit` is a string.

```javascript

unit = new String(unit);
index = _.indexOf(units, unit.toUpperCase());
if (index === -1) {
   // logic for figuring out the index ....
}
// the rest of the function ...
```

Thoughts?

halkon: Also to add to my proposition: 

```javascript
x = undefined
// undefined
y = new String(x)
// String {0: "u", 1: "n", 2: "d", 3: "e", 4: "f", 5: "i", 6: "n", 7: "e", 8: "d", length: 9, [[PrimitiveValue]]: "undefined"}
y.toUpperCase()
// "UNDEFINED"
```

Droogans: I don't mind how it's written. If `unit` is not a string, it'll come back as `-1` from the search of the list. The only small gripe is

```js
if (_.isUndefined(unit) || !_.contains(units, unit.toUpperCase()) {
```

but that's not big enough to matter to me, honestly.

halkon: It's not huge, the main reason for converting it to a string is so that `.toUpperCase` doesn't fail on undefined/null/boolean/etc.  I think that's the issue I have is, we're assuming that if it's not `Undefined` that it must be a string.

src/utilities/rxBytesConvert/docs/examples/rxBytesConvert.demo.js
```diff
@@ -0,0 +1,6 @@
+angular.module('demoApp')
+.controller('rxBytesConvertCtrl', function ($scope) {
+    $scope.sizeGB = 42e10;
+    $scope.sizeTB = 125e12;
+    $scope.sizePB = 17134e13;
```
100stacks: Since this is a demo, commenting  these numbers, e.g. `420 GB`, `125 TB`, and `171.44 PB`, would help users looking at this.

src/utilities/rxBytesConvert/rxBytesConvert.meta.json
100stacks: @CITguy @kevincareer427, should we mark this as `prototype` to align with the new stability badges?

src/utilities/rxBytesConvert/scripts/rxBytesConvert.spec.js
```diff
@@ -0,0 +1,101 @@
+/* jshint node: true */
```
CITguy: You can remove this jshint comment

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
Droogans: Hmm...well, getting that defensive in the programming of inputs is definitely a step in a new direction in the project.

In the past, I've tended to rely on documenting input/output types and expecting users to RTFM before filing bugs about using numbers where strings are required.

That being said, I did notice that there's very little documentation around this filter, particularly the input/output types. @kevincareer427 you've only specified a couple of examples, and none of them demonstrate passing in a string for custom `unit`s.

halkon: That's a very good point, and even having one less possible call to `_.indexOf` is not a huge performance gain, just seemed redundant and iffy about just `isUndefined` and then calling `toUpperCase`. 

But as you said that's a step in a new direction, and well, yeah.  Let's not do that

halkon: A question to you @Droogans should we error out early if the values are not what we expect?

kevincareer427:
I think Freddy's right, it does fail when users pass in null, boolean or number for unit.

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
Droogans: If we attempted to catch/cast everything we see (and then unit test it) in a dynamic language, our codebase would become much, much larger, without adding any real value to the codebase.

If you want to catch and reject based on type signatures, use a strongly typed language. I find that Javascript works best when you use [duck typing](https://en.wikipedia.org/wiki/Duck_test).

halkon: 👍  and yes, with javascript i totally agree with that. 

however it is not useful when the error is just
```
Uncaught TypeError: Cannot read property 'toUpperCase' of undefined(…)
```

Which is what my proposal was trying to handle (and remove logic).  I wouldn't try/catch either, all I'm saying is if we're going to do magic, let's make sure we're doing without causing more confusion to our users. 

Either way, moot point, and really it's something so micro that it doesn't matter.  I don't want @kevincareer427 to think he shouldn't put up PRs 😁

halkon:
LGTM

src/utilities/rxBytesConvert/scripts/rxBytesConvert.js
CITguy: sooo.... moving on. 😄

100stacks:
Nice work @kevincareer427, on your first PR!

LGTM 🎉

PR #1650, opened 2016-05-06
chore(template): Update to newer page object principles

Droogans:
I was getting started removing all references of jshint and saw this file...decided it was worth the extra ten minutes to update it to our newer style of page object generation.

JIRA: N/A

### LGTMs
- [x] Dev LGTM


CITguy:
likely going to pass, but I'll give Travis the benefit of the doubt

LGTM

PR #1655, opened 2016-05-12
Update deps moment

Droogans:
JIRA: N/A

This is prep for upgrading protractor to 3.0+, which means some of our out of date deps are going to get upgraded too.

### LGTMs
- [x] Dev LGTM


CITguy:
LGTM

PR #1656, opened 2016-05-12
fix(npm): Drop unused semver library

Droogans:
JIRA: N/A

I did not find a single use of this library in our project.

### LGTMs
- [x] Dev LGTM

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "mocha": "^2.1.0",
     "moment": "~2.9.0",
     "node-timing.js": "^1.1.0",
-    "protractor": "^2.0.0",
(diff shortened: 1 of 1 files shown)


parlarjb:
Could you double check with @CITguy that it's not used in the release process?

Droogans:
Yep, that's my big fear. He mentioned something about accounting for a missing peer dep in another node module, and it was just a hacky workaround that he hopes is no longer an issue.

CITguy:
LGTM

That dependency was added to correct an issue with a nested dependency a while back. Apparently, that issue has been resolved.

PR #1658, opened 2016-05-12
fix(eslint): Address errors that emerged from the void

Droogans:
JIRA: N/A

I think the commit message says everything that needs to be said here.

I am thinking that perhaps a version change is to blame? I honestly have no idea how this happened.

### LGTMs
- [x] Dev LGTM


parlarjb:
Damn you e-slint, damn you straight to hell!

LGTM

PR #1662, opened 2016-05-13
refactor(rxExample): New Layout

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-860

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM


CITguy:
Example below shows "Tabs" with updated source tab styling (for extra Inception points)

### BEFORE
rxexample-tabs-and-tabs-content-before


### AFTER
rxexample-tabs-and-tabs-content

demo/css/prism.css
```diff
@@ -1,20 +1,21 @@
+/* http://prismjs.com/download.html?themes=prism-coy&languages=markup+css+clike+javascript&plugins=line-numbers */
 /**
- * prism.js default theme for JavaScript, CSS and HTML
- * Based on dabblet (http://dabblet.com)
- * @author Lea Verou
+ * prism.js Coy theme for JavaScript, CoffeeScript, CSS and HTML
```
CITguy: Replaced default theme with "Coy" theme

demo/less/_code.less
```diff
@@ -1,14 +1,25 @@
...
+/* Prism style overrides */
 pre[class*="language-"] {
+  &:before,
+  &:after {
+    content: none;
+  }
+
+  // reduced by 1px because line-height math for zebra stripes has to result
+  // in whole pixel values to align correctly with content
+  font-size: 12px;
```
CITguy: Striping depends on some calculations to determine where to apply 50% color stops in a `linear-gradient()`. 

### Broken Font Size
The original font-size was 13px, so let's walk through the calculations using this value. 

Given a `font-size: 13px` (inherited via CSS in the framework) and a `line-height: 1.5` (defined in Coy theme), we can calculate that the px value for the line height is `19.5px` (`13px * 1.5`) -- a partial-pixel value. Since the background gradient size is 3em x 3em (define in the Coy theme), we can calculate the size to be `39px` (`13px * 3`) -- twice line height.  With 50% color stops (defined in Coy theme), each color in the stripe will calculate to half the height of the image (`19.5px`). However, because there's no physical representation for `0.5px`, the browser has to do some rounding to draw the gradient. This results in one of the stripes to be larger than the other, and we get uneven striping.

### Corrected Font Size
To correct this pixel rounding, we change the font-size to a value that results in whole pixels. 

Given a `font-size: 12px;` and a `line-height: 1.5`, we can calculate that the px value for the line height is `18px` (`12px * 1.5`) -- whole pixels ^_^. Since the background gradient is 3em x 3em, we can calculate the size to be `36px` (`12px * 3)` -- twice line height, still whole pixels. With 50% color stops, each color in the stripe will calculate to half the height of the image (`18px`). The calculated values in this scenario all result in whole pixel values, so the browser doesn't have to perform any pixel rounding and our stripes are even height and line up beautifully with the content.

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
-	}
-}
-
 /* Code blocks */
 pre[class*="language-"] {
-	padding: 1em;
+	position: relative;
 	margin: .5em 0;
+	-webkit-box-shadow: -1px 0px 0px 0px #358ccb, 0px 0px 0px 1px #dfdfdf;
+	-moz-box-shadow: -1px 0px 0px 0px #358ccb, 0px 0px 0px 1px #dfdfdf;
```
parlarjb: Do we still need the vendor prefixes for box-shadow?

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
+	box-shadow: -1px 0px 0px 0px #358ccb, 0px 0px 0px 1px #dfdfdf;
+	border-left: 10px solid #358ccb;
+	background-color: #fdfdfd;
+	background-image: -webkit-linear-gradient(transparent 50%, rgba(69, 142, 209, 0.04) 50%);
+	background-image: -moz-linear-gradient(transparent 50%, rgba(69, 142, 209, 0.04) 50%);
+	background-image: -ms-linear-gradient(transparent 50%, rgba(69, 142, 209, 0.04) 50%);
+	background-image: -o-linear-gradient(transparent 50%, rgba(69, 142, 209, 0.04) 50%);
```
parlarjb: And I think `linear-gradient` is supported well now

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
 /* Inline code */
 :not(pre) > code[class*="language-"] {
-	padding: .1em;
-	border-radius: .3em;
+	position: relative;
+	padding: .2em;
+	-webkit-border-radius: 0.3em;
+	-moz-border-radius: 0.3em;
+	-ms-border-radius: 0.3em;
+	-o-border-radius: 0.3em;
```
parlarjb: same

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
+	width: 40%;
+	height: 20%;
+	max-height: 13em;
+	-webkit-box-shadow: 0px 13px 8px #979797;
+	-moz-box-shadow: 0px 13px 8px #979797;
+	box-shadow: 0px 13px 8px #979797;
+	-webkit-transform: rotate(-2deg);
+	-moz-transform: rotate(-2deg);
+	-ms-transform: rotate(-2deg);
+	-o-transform: rotate(-2deg);
```
parlarjb: same

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
+	z-index: -2;
+	display: block;
+	position: absolute;
+	bottom: 0.75em;
+	left: 0.18em;
+	width: 40%;
+	height: 20%;
+	max-height: 13em;
+	-webkit-box-shadow: 0px 13px 8px #979797;
+	-moz-box-shadow: 0px 13px 8px #979797;
```
parlarjb: same

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
+}
+
+:not(pre) > code[class*="language-"]:after,
+pre[class*="language-"]:after {
+	right: 0.75em;
+	left: auto;
+	-webkit-transform: rotate(2deg);
+	-moz-transform: rotate(2deg);
+	-ms-transform: rotate(2deg);
+	-o-transform: rotate(2deg);
```
parlarjb: same

demo/css/prism.css
```diff
@@ -26,101 +27,251 @@ pre[class*="language-"] {
...
+.namespace {
+	opacity: .7;
+}
+
+@media screen and (max-width: 767px) {
+	pre[class*="language-"]:before,
+	pre[class*="language-"]:after {
+		bottom: 14px;
+		-webkit-box-shadow: none;
+		-moz-box-shadow: none;
```
parlarjb: same

demo/scripts/rxPrism.js
```diff
@@ -2,11 +2,10 @@ angular.module('demoApp')
 .directive('rxPrism', function ($timeout, $http) {
     return {
         restrict: 'E',
-        template: '<pre><code class="language-{{language}}" ng-transclude></code></pre>',
+        template: '<pre><code class="line-numbers language-{{language}}" ng-transclude></code></pre>',
         scope: {
             language: '@',
-            codeUrl: '@',
-
+            codeUrl: '@'
```
parlarjb: nooooooooooooo

demo/css/prism.css
CITguy: this is vendor css, I don't think we need to worry about modifying it

demo/css/prism.css
```diff
@@ -1,20 +1,21 @@
+/* http://prismjs.com/download.html?themes=prism-coy&languages=markup+css+clike+javascript&plugins=line-numbers */
```
CITguy: This file was downloaded directly from prismjs.com and should be considered a vendor file and left as-is.

demo/css/prism.css
CITguy: vendor file, no need to update

demo/css/prism.css
CITguy: vendor file, no need to update

demo/css/prism.css
CITguy: vendor file, no need to update

demo/css/prism.css
CITguy: vendor file, no need to update

demo/css/prism.css
CITguy: vendor file, no need to update

demo/css/prism.css
CITguy: vendor file, no need to update

demo/scripts/rxExample.js
```diff
@@ -1,5 +1,12 @@
 angular.module('demoApp')
 .directive('rxExample', function ($http) {
+    var MAX_PREVIEW_LINES = 8;
```
CITguy: change to `10`

demo/scripts/rxExample.js
```diff
@@ -8,21 +15,61 @@ angular.module('demoApp')
         },
         link: function (scope) {
             var baseName = 'examples/' + scope.name;
+            var _canExpand = { markup: false, js: false, less: false };
+            var _isExpanded = { markup: false, js: false, less: false };
+            // markup not included b/c it's required for minimum functionality
+            var _hasSource = { js: false, less: false };
 
+            // used by ng-include and code-url
             scope.markup_url = baseName + '.html';
```
CITguy: make lowerCamelCase

glynnis:
We should lose the left margin on the line numbers with an override so it matches mock-ups a little better. Here's the mock-up:
screen shot 2016-05-16 at 10 41 09 am

glynnis:
Increase spacing between demos themselves and the source code tabs.

demo/scripts/rxExample.js
halkon: Out of pure curiosity and not a suggestion but, would it be overkill to just stuff this into a constant somewhere?

halkon: or even here, but just a curiosity

demo/scripts/rxPrism.js
halkon: yessssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss

halkon:
I'll give the DEV LGTM 😀  another dev that was on vidyo can do the other

demo/scripts/rxExample.js
CITguy: I'm thinking it may be overkill to make an angular constant. This value only makes sense in the context of the `rxExample` directive, so it doesn't make much sense exposing it to other things.

demo/scripts/rxPrism.js
CITguy: maybe?!

CITguy:
Spoke with @glynnis, but the left "margin" on the line numbers was placed there to accommodate 3+ digit line numbers. Additionally, it's being set by `width: 3em;`, so rather than muck with vendor styles, it was decided that we would ignore trying to tweak this styling.

glynnis:
LGTM

parlarjb:
LGTM

PR #1666, opened 2016-05-16
refactor(less): FRMW-746 remove obsolete LESS mixins

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-746

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM

---
Removed obsolete LESS Mixins.


100stacks:
These LESS Mixin deprecations are documented here: https://github.com/rackerlabs/encore-ui/wiki/1.x-Deprecated-Items#less-mixins

src/components/rxToggleSwitch/rxToggleSwitch.less
```diff
@@ -17,9 +17,9 @@
   position: relative;
   text-align: right;
 
-  .box-shadow(inset 1px 2px @rxToggleSwitch-blur-dist @rxToggleSwitch-blur-spread rgba(0, 0, 0, 0.50));
+  box-shadow: inset 1px 2px @rxToggleSwitch-blur-dist @rxToggleSwitch-blur-spread rgba(0, 0, 0, 0.50);
   .user-select(none);
-  .box-sizing(border-box);
+  box-sizing: padding;
```
glynnis: Was the switch away from `border-box` here intentional?

100stacks: @glynnis, no...thanks for the catch! 👍

src/styles/mixins.less
```diff
@@ -78,9 +42,9 @@
 
 // Reset native styling of form input elements
 .ResetFormInput() {
-  .box-sizing(border-box);
-  .background-clip(padding);
```
halkon: so technically this never worked huh?

halkon:
LGTM

src/styles/mixins.less
```diff
@@ -78,9 +42,9 @@
...
-  .border-radius(0);
+  box-sizing: border-box;
+  background-clip: padding-box;
```
CITguy: Any reason for the switch from `padding` to `padding-box`?

halkon: Per https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip `padding-box` seems like the right value, hence my comment that this perhaps never worked before?

src/styles/mixins.less
CITguy: apparently not

src/styles/mixins.less
CITguy: Considering that the value previously passed to the mixin produced `background-clip: padding;`, the browser would've ignored this and used the default `border-box` (https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip). I'd change this to `border-box` to avoid any unexpected changes.

CITguy: actually, better yet. remove this property. it's clearly not doing anything useful.

glynnis:
LGTM

src/styles/mixins.less
100stacks: Removed the non-functioning `.background-clip(padding)` from `.ResetFormInput()`.

src/components/rxModalAction/rxModalAction.less
```diff
@@ -55,8 +55,8 @@
   transform: translate(-50%, 0);
   -webkit-transform: translate(-50%, 0);
   -moz-transform: translate(-50%, 0);
-  .box-shadow(0 3px 7px rgba(0, 0, 0, 0.3));
-  .background-clip();
+  box-shadow: 0 3px 7px rgba(0, 0, 0, 0.3);
+  background-clip: padding-box;
```
CITguy: 👍

CITguy:
LGTM

PR #1667, opened 2016-05-17
feat(rxAppNavItem): FRMW-804 Target & Href prefix for nav items

halkon:
JIRA: https://jira.rax.io/browse/FRMW-804

This PR allows for applications to define a set of injectables (constant, value, factory, service, provider) that will allow for further customization of the nav. 

* HREF prefixes for nav items
* target prefixes for nav items

In the future we could also add `target` as nav item definitions (not developed in this PR, just worth noting).

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxApp/scripts/rxAppNavItem.js
+++ b/src/components/rxApp/scripts/rxAppNavItem.js
@@ -57,7 +57,58 @@ angular.module('encore.ui.rxApp')
         scope: {
             item: '='
         },
-        controller: function ($scope, $location, rxVisibility, Permission) {
+        controller: function ($scope, $location, $injector, rxVisibility, Permission, urlUtils) {
+            /*
+             * @description Determines whether or not a nav item should have its href prefixed
+             * based on whether the `$injector` has a `NAV_ITEM_PREFIX` injectable
+             *
+             * _This is *NOT* meant for general consumption, this is strictly for the Origin Project_
+             * _This will eventually be deprecated and removed_
+             *
+             * @param {string} [url] - URL for the nav item's href
+             */
+            $scope.getUrl = function (url) {
+                // For URLs that have no URL definition, let's go ahead and return right away
+                // this avoids issues when we do have a prefix but really the nav item should not have
+                // any defined href, i.e. items that have subitems
+                if (_.isEmpty(url)) {
+                    return url;
+                }
+
+                // Check if we have a definition of NAV_ITEM_PREFIX, if so let's retrieve it and return the given URL
+                // appended to the prefix.  This allows applications like origin to prefix nav items, while not
+                // messing with nav items in the demo/documentation.
+                //
+                // _This is *NOT* meant for general consumption, this is strictly for the Origin Project_
+                // _This will eventually be deprecated and removed_
+                //
+
+                if ($injector.has('NAV_ITEM_PREFIX')) {
+                    var prefix = urlUtils.parseUrl($injector.get('NAV_ITEM_PREFIX'));
+                    return prefix.protocol.concat('//').concat(prefix.host).concat(url);
+                } else {
+                    // Return as normal if no prefix
+                    return url;
+                }
+
+            };
+            /*
+             * @description Determines whether or not the links need to point to a target, this allows
+             * for origin and applications that show the nav to implement a target in which to have the links
+             * open in.
+             *
+             * If ever there was a need to point links to a different target than an application specific
+             * target, we could implement logic here to inspect the item and determine the target.
+             * (i.e. opening an external application in a new window)
+             */
+            $scope.getTarget = function () {
+                // Check if we have a definition of NAV_ITEM_TARGET, if so let's retrieve it and enable the target attr
+                // on the nav item.  This allows applications like origin to give a target to it's nav items, while not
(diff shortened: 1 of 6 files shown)


src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -57,7 +57,52 @@ angular.module('encore.ui.rxApp')
         scope: {
             item: '='
         },
-        controller: function ($scope, $location, rxVisibility, Permission) {
+        controller: function ($scope, $location, rxVisibility, Permission, $injector) {
+            /*
+             * @description Determines whether or not a nav item should have it's url prefixed
```
Droogans: ...should have *its* url prefixed...

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -57,7 +57,52 @@ angular.module('encore.ui.rxApp')
...
+                // this avoids issues when we do have a prefix but really the nav item should not have
+                // any defined href, i.e. items that have subitems
+                if (_.isEmpty(url)) {
+                    return url;
+                }
+
+                // Check if we have a definition of NAV_ITEM_PREFIX, if so let's retrieve it and return the given URL
+                // appended to the prefix.  This allows applications like origin to prefix nav items, while not
+                // messing with nav items in the demo/documentation.
+                var prefix = $injector.has('NAV_ITEM_PREFIX') ? $injector.get('NAV_ITEM_PREFIX') : false;
```
Droogans: The docs don't say what it returns when there is no injector matching the name found. It just returns `* - the instance`.

Do you have to manually catch the case where there is no `'NAV_ITEM_PREFIX'` with a ternary expression?

src/components/rxApp/scripts/rxAppNavItem.spec.js
```diff
@@ -1,6 +1,6 @@
 describe('encore.ui.rxApp', function () {
     describe('rxAppNavItem', function () {
-        var scope, compile, el, someProp, rxvisibility;
+        var scope, compile, el, someProp, rxvisibility, rxenvironmenturlfilter, provide;
```
Droogans: Is there a reason these aren't camelCased?

halkon:
* Add Documentation around this not being something that people should use
* Convert unit tests to use `_injectable_`
* Use `document.createElement('a')` to do better URL construction
*

parlarjb:
LGTM

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -57,7 +57,58 @@ angular.module('encore.ui.rxApp')
...
+
+                // Check if we have a definition of NAV_ITEM_PREFIX, if so let's retrieve it and return the given URL
+                // appended to the prefix.  This allows applications like origin to prefix nav items, while not
+                // messing with nav items in the demo/documentation.
+                //
+                // _This is *NOT* meant for general consumption, this is strictly for the Origin Project_
+                // _This will eventually be deprecated and removed_
+                //
+
+                var prefix = $injector.has('NAV_ITEM_PREFIX') ? $injector.get('NAV_ITEM_PREFIX') : false;
```
CITguy: creating a variable regardless of if it being used seems smelly, see suggestion on line 93

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -57,7 +57,58 @@ angular.module('encore.ui.rxApp')
...
+                if (prefix) {
+                    prefix = urlUtils.parseUrl(prefix);
```
CITguy: overriding the type seems smelly, see suggestion on line 93

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -57,7 +57,58 @@ angular.module('encore.ui.rxApp')
...
+                    return prefix.protocol.concat('//').concat(prefix.host).concat(url);
+                }
+
+                // Return as normal if no prefix
+                return url;
```
CITguy: The logic of 86-93 seems smelly. I'd rather see something like...

```javascript
if ($injector.has('NAV_ITEM_PREFIX')) {
  var prefix = urlUtils.parseUrl($injector.get('NAV_ITEM_PREFIX'));
  return prefix.protocol.concat('//').concat(prefix.host).concat(url);
} else {
  return url;
}
```

halkon: Do you really want the else? seems redundant.  Otherwise, 👍

halkon: by that question I mean:
```javascript
if ($injector.has('NAV_ITEM_PREFIX')) {
    var prefix = urlUtils.parseUrl($injector.get('NAV_ITEM_PREFIX'));
    return prefix.protocol.concat('//').concat(prefix.host).concat(url);
}

// Return as normal if no prefix
return url;
```

CITguy:
LGTM

CITguy:
@halkon, please squash

halkon:
@CITguy squashed

PR #1673, opened 2016-05-21
Rx page objects in utils

Droogans:
JIRA:

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM


Droogans:
I'm going to try and get an identical pull request for this against 2.x, because so much of this has changed.

package.json
```diff
@@ -14,8 +14,8 @@
   "scripts": {
     "test": "grunt test",
     "postinstall": "grunt hooks",
-    "eslint-dev": "eslint -c .eslintrc.json --ignore-path .eslintignore src/** demo/**",
-    "eslint-test": "eslint -c .eslintrc.spec.json --ignore-path .eslintignore-spec src/** demo/**"
+    "eslint-dev": "eslint -c .eslintrc.json --ignore-path .eslintignore-dev {src,demo}",
+    "eslint-test": "eslint -c .eslintrc.test.json --ignore-path .eslintignore-test {src,demo} utils/rx-page-objects/{test,src}"
```
Droogans: Double check to make sure this is truly a recursive scan of the directory tree.

Droogans:
You forgot to delete more grunt tasks.

```
ack PageObject grunt-tasks/
```

Droogans:
You missed some documentation around "Odd patterns in the page object code". Delete it.

demo/templates/modules/showModule.html
```diff
@@ -42,9 +42,5 @@ <h2 class="readme-title">Readme</h2>
     <tab heading="Less" ng-if="module.docs.less">
       <rx-prism language="css">{{ module.docs.less }}</rx-prism>
     </tab>
-
-    <tab heading="Protractor" ng-if="module.docs.midway">
```
parlarjb: How come you're taking this out?

parlarjb:
Is there a jira or something that goes into the reasoning here?

utils/rx-page-objects/index.js
```diff
@@ -0,0 +1,29 @@
...
+var srcPath = require('path').join(__dirname, 'src');
+var pageObjects = {};
+var exercises = {};
+
+/**
+ * We're going to be iterating over every file in `src` for page object functionality.
+ * This will take each `require`'d file and iterate over all it's exported members, and
+ * add them to the `destination` (either the page objects, or the exercises).
+ */
+var addModule = (destination, moduleLocation) => {
```
parlarjb: oooooh, `=>`

CITguy: Yeah, that's the nice thing about the page objects. We can use ES6! :)

demo/templates/modules/showModule.html
Droogans: They were a bad idea from the outset (my fault, by the way).

So, we have a demo component. In that demo component, there's just *that component's* JS in a tab. *Just that component's* LESS in a tab...and then, the *entire* end to end test suite? For *multiple* components on the demo page? That really doesn't make any sense.

It doesn't serve as a nice introduction to end to end testing a particular instance of a component. Most times, it existed to confuse people (looking at these super meta tests that test edge cases around particular use cases of multiple components, exercises being invoked directly in the midway, etc.).

Putting the `// the component's exercise file, if it exists, is below` was the nail in the coffin for me. It's complete gibberish to anyone but us (and even then I wonder).

CITguy:
@Droogans, I'm still not convinced that we need to move the `*.midway.js` files. The existing midways are primarily to test component functionality, not page object functionality.

Droogans:
@parlarjb there isn't an explicit Jira story to track this, but we have a few that are a burden to get done without this reorg.

https://jira.rax.io/browse/FRMW-442 "Split rxFieldName page object into its own file."

That would require there to be two page objects in the `rxForm` directory, or worse, an entire component directory under `src/elements/rxFieldName`, just because the page objects would go there.

Essentially, the problem is that there isn't a 100% one to one correlation between "component in UI" -> "page object file". `rxMisc` https://jira.rax.io/browse/FRMW-517 is a huge example of this. Those functions are in desperate need of reorganization. It's a junk drawer of unrelated functionality (low cohesion) that could be better expressed as multiple separate files. One for interacting with text on the screen (currencyToPennies, unlessNull, date helpers), one for interacting with positioning elements in the screen/scrolling (what was rxFloatingHeader stuff).

Droogans:
> The existing midways are primarily to test component functionality, not page object functionality.

Is that the argument for keeping the midways away from the page objects? If so, I'm not sure how that's going to make anyone's life easier. It seems like an arbitrary decision that would isolate a key aspect of the page object library (its tests) away from its source.

CITguy:
Please disregard my previous comment. Based on our conversation, my concern shouldn't be an issue.

parlarjb:
LGTM

demo/templates/modules/showModule.html
100stacks: @Droogans, thanks for explaining this yesterday.  Definitely makes since to remove from the demos.

guides/testing.md
```diff
@@ -112,11 +112,11 @@ The filename for these page objects follows the `componentName.page.js` conventi
 
 On a build of EncoreUI, all page object files are concatenated and tarballed into the `dist` directory. They are then published either manually or via Travis. To use these page objects, developers should include the following dependency in their app's `package.json` file:
 
-    "rx-page-objects": "rx-page-objects-1.48.0.tgz"
+    "rx-page-objects": "^1.0.0"
 
 Alternatively, they can install the file using this command:
 
-    npm install --save-dev rx-page-objects@1.7.1
+    npm install --save-dev rx-page-objects@1
```
CITguy: Is the version here correct?

utils/rx-page-objects/.npmignore
```diff
@@ -1 +1,2 @@
-doc
\ No newline at end of file
+doc
+test
```
CITguy: Why are we ignoring `test`?

guides/testing.md
Droogans: It's the same as saying `@1.x`, just less typing.

PR #1675, opened 2016-05-22
Making sure travis fails on forks.

Droogans:
This should fail (pretty early, too).

I'll open the upstream version in a moment.


Droogans:
Yeah...this "failed"....right.

PR #1676, opened 2016-05-22
test(ci): Move all selenium based work to codeship

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-853

Half of this stuff is now located at https://codeship.com/projects/102410

This does not include getting rid of `grunt phantomjs-check:allModules`, which will happen after this PR lands.

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM

### DIFF
--- a/utils/quitIfFork.js
+++ b/utils/quitIfFork.js
@@ -1,0 +1,34 @@
+#!/usr/bin/env node
+
+'use strict';
+
+let errorMessage = `
+*************************************************************
+*************************************************************
+*******************  "YOU SHALL NOT PASS"  ******************
+*************************************************************
+*************************************************************
+
+Stop pushing from forks! This is to remind you that all builds should be targeting rackerlabs/encore-ui.
+For most people seeing this message, it means to stop running "git push origin branch-name", and start
+running "git push upstream branch-name" when creating pull requests. To fix this:
+
+If you are not a Rackspace employee:
+
+1. Thank you for your interest in making EncoreUI better.
+2. Make a comment on the pull request mentioning @rackerlabs/encore-ui-admin stating that you need your PR reviewed.
+
+We'll do our best to make sure that someone can clone your branch and re-run it for you to make sure that
+it passes all of our quality checks. In the meantime, thank you for your patience.
+
+Otherwise:
+
+1. Close the PR associated with this travis build.
+2. Push the exact same branch up to the upstream remote located at git@github.com:rackerlabs/encore-ui.git
+3. Create a new pull request.
+`;
+
+if (process.env.TRAVIS_SECURE_ENV_VARS === 'false') {
+    console.error(errorMessage);
+    process.exit(1);
+};
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,40 +2,21 @@ sudo: false
 language: node_js
 node_js:
   - "5.1.1" # same as .node-version
-addons:
-  firefox: "35.0"
-  apt:
-    packages:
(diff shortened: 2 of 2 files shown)


utils/quitIfFork.js
```diff
@@ -0,0 +1,52 @@
...
+*******************************************************************************************************
+***************************************  "YOU SHALL NOT PASS"  ****************************************
+*******************************************************************************************************
+*******************************************************************************************************
+*******************************************************************************************************
+*******************************************************************************************************
+*******************************************************************************************************
+*******************************************************************************************************
+
+Stop pushing from forks! This is to remind you that all builds should be targeting rackerlabs/encore-ui.
```
100stacks: @Droogans, not sure what 

> Stop pushing from forks!

means.  Does this mean we no longer fork `EncoreUI` and everyone should now push directly to `EncoreUI` `master`?  Also, for the visual regression tool you built...will we be able to push from our branches vs. pushing to `EncoreUI` `master`?

Droogans: I set up nearly all my projects in git this way:

1. Fork the project (in this case, it's called encore-ui)
2. Grab the SSH link in my fork's main page (`git@github.com:Droogans/encore-ui.git`)
3. Clone that. Since it's the first one used, git automatically calls it `origin`.
4. Add the original endpoint as `upstream`. I'd do this with `git remote add upstream git@github.com:rackerlabs/encore-ui.git`

I can now pull in new code to from `upstream`, push it to `origin`, and then do my work in `origin`.

This makes it so that I can keep my many, many branches (135 currently) out of the main repo (rackerlabs/encore-ui has 51). There's a few other benefits, but that's the big one.

So, when I submit a pull request from `Droogans:encore-ui`, TravisCI sees that and says "you're an outsider, I'm not going to expose any of these sensitive environment variables to you". These environment variables include our github service account's token for making API requests against the project. If random people could open pull requests that updated the protractor end to end tests to simply email them the sensitive environment variables, we'd be in for a bad time.

Codeship does something similar. If you push from `Droogans:encore-ui`, it straight up doesn't run anything. So, now that we're doing our end to end tests in codeship, we *have* to start using it. So, from now on, if anyone pushes up PRs from an endpoint other than `rackerlabs:encore-ui`, we're going to show them this message and ask that they close the PR, push to rackerlabs, then reopen it.

If you're not a Rackspace employee, I've stated that they should at-mention me in the PR asking for help. I'd review the PR to make sure it's not malicious, add their endpoint as a remote (`git add remote j-random-hacker git@github.com:jrandomhacker/encore-ui.git`), fetch the branch, and then push it up as a new branch to `rackerlabs`. This would allow us to run all our end to end tests (including visual regression) as we'd expect before merging it into master.

All of this would happen on branches. Nobody will start pushing to `master`, it's a protected branch, so they couldn't do it anyway.

parlarjb:
I gave the Dev LGTM, but let me know when the vidyo review happens too, I'd like to be there for that

utils/quitIfFork.js
```diff
@@ -0,0 +1,52 @@
...
+For most people seeing this message, it means to stop running "git push origin branch-name", and start
+running "git push upstream branch-name" when creating pull requests. To fix this:
+
+If you are not a Rackspace employee:
+
+1. Pat yourself on the back.
```
CITguy: This seems a wee bit condescending. I'd change to `Thank you for your interest in making EncoreUI better.`

CITguy:
LGTM

PR #1681, opened 2016-05-23
refactor(common): Remove filter-box class

glynnis:
JIRA: https://jira.rax.io/browse/FRMW-219

### LGTMs
- [x] Dev LGTM
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxFloatingHeader/scripts/rxFloatingHeader.spec.js
+++ b/src/components/rxFloatingHeader/scripts/rxFloatingHeader.spec.js
@@ -67,7 +67,7 @@ describe('rxFloatingHeader', function () {
         scope.$digest();
     });
 
-    it('should apply .filter-header and .filter-box to headers and inputs', function () {
+    it('should apply .filter-header to headers and inputs', function () {
         var ths = el.find('th');
         var inputs = el.find('input');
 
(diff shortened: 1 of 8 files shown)


src/components/rxFloatingHeader/scripts/rxFloatingHeader.spec.js
```diff
@@ -79,7 +79,6 @@ describe('rxFloatingHeader', function () {
         expect(ths.eq(2).hasClass('filter-header'), 'third th').to.be.false;
 
         expect(inputs, 'one input').to.have.length(1);
-        expect(inputs.eq(0).hasClass('filter-box')).to.be.true;
```
halkon: I know we talked about this and I said to go ahead and remove it, but if the `rxSearchBox` still applies here, maybe we ought to check for it?

I don't know though, still seems like an unnecessary flimsy test

halkon:
LGTM

src/components/rxFloatingHeader/scripts/rxFloatingHeader.spec.js
CITguy: I'm okay with this. The logic originally assumed that the first input field in a floating header would be a search box. However, with more advanced filtering patterns emerging, this assumption is becoming less reliable and can lead to unexpected styling problems. The `rxSearchBox` directive isn't added automatically by `rxFloatingHeader` so there's no need to check for it.

CITguy:
LGTM

PR #1702, opened 2016-06-06
refactor(tabs): remove tabs.tab#visit function

kevincareer427:
JIRA: https://jira.rax.io/browse/FRMW-909

### LGTMs
- [x] Dev LGTM
- [x] QE LGTM

### DIFF
--- a/utils/rx-page-objects/src/tabs.page.js
+++ b/utils/rx-page-objects/src/tabs.page.js
@@ -1,3 +1,4 @@
+var rxMisc = require('./rxMisc.page').rxMisc;
 var Page = require('astrolabe').Page;
 
 var tabFromElement = function (tabElement) {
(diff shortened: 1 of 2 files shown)


utils/rx-page-objects/src/tabs.page.js
```diff
@@ -131,19 +131,11 @@ var tabFromElement = function (tabElement) {
...
-         * @function
-         * @memberof tabs.tab
-         */
-        visit: {
-            value: function () {
-                browser.actions().mouseDown(tabElement).mouseUp().perform();
+                if (slowClick) {
+                    exports.rxMisc.slowClick(tabElement);
+                } else {
+                    browser.actions().mouseDown(tabElement).mouseUp().perform();
```
Droogans: These two operations are identical.

The second one should just be `tabElement.click()`.

CITguy: `tabElement.click()` didn't seem to be working, so we replaced it with the logic from `visit()`

CITguy: additionally, we didn't want to change the API of `#click()`

Droogans: Depending on the user's version of selenium and firefox, this may or may not change over time.

I've seen teams need to use `browser.actions()` to click something, then have it not work a few months later, only to have it required to function later on!

I find it best to allow for both, and to keep them up to date for the framework project's needs as they evolve. But those needs are not the same for every project that uses these page objects.

CITguy:
LGTM

PR #1714, opened 2016-06-10
Chrome e2e tests

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-910

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM

I've included a Design LGTM because I had to modify the action menu's less file to get chrome to stop complaining about an overlapping element. I already showed @ssalinas24 he said it looked ok.

### DIFF
--- a/src/elements/ActionMenu/styles/rxActionMenu.less
+++ b/src/elements/ActionMenu/styles/rxActionMenu.less
@@ -23,7 +23,7 @@
   z-index: 99;
   position: absolute;
   top: -10px;
-  right: 15px;
+  right: 25px;
   min-width: 20em;
   font-size: 0.9em;
   border: 4px solid @white;
--- a/utils/rx-page-objects/src/rxDatePicker.page.js
+++ b/utils/rx-page-objects/src/rxDatePicker.page.js
@@ -275,7 +275,8 @@ var rxDatePicker = {
 
         set: function (value) {
             this.open();
-            rxSelect.initialize(this.rootElement.element(by.model('currentMonth'))).select(value);
+            var slowClick = false;
+            rxSelect.initialize(this.rootElement.element(by.model('currentMonth'))).select(value, slowClick);
         }
     },
 
@@ -295,7 +296,8 @@ var rxDatePicker = {
 
         set: function (value) {
             this.open();
-            rxSelect.initialize(this.rootElement.element(by.model('currentYear'))).select(value);
+            var slowClick = false;
+            rxSelect.initialize(this.rootElement.element(by.model('currentYear'))).select(value, slowClick);
         }
     },
 
--- a/utils/rx-page-objects/src/rxNotify.page.js
+++ b/utils/rx-page-objects/src/rxNotify.page.js
@@ -88,6 +88,14 @@ var notification = function (rootElement) {
                 return this.isDismissable().then(function (dismissable) {
                     if (dismissable) {
                         page.btnDismiss.click();
+                        // hack for chrome support -- this is for any possible
+                        // alerts that are opened by closing this notification.
+                        // also, before you check, EC.isAlertPresent does not work here.
+                        return browser.getCapabilities().then(function (capabilities) {
+                            if (capabilities.get('browserName') === 'chrome') {
+                                browser.sleep(500);
+                            }
+                        });
                     }
                 });
             }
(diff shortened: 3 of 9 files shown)


utils/rx-page-objects/src/rxNotify.page.js
```diff
@@ -88,6 +88,10 @@ var notification = function (rootElement) {
                 return this.isDismissable().then(function (dismissable) {
                     if (dismissable) {
                         page.btnDismiss.click();
+                        // hack for chrome support -- this is for any possible
+                        // alerts that are opened by closing this notification.
+                        // also, before you check, EC.isAlertPresent does not work here.
+                        browser.sleep(500);
```
Droogans: Instead of waiting 500ms in FF, you could ask what the browser is before you do a sleep.

100stacks:
LGTM

glynnis:
Design LGTM

src/elements/ActionMenu/styles/rxActionMenu.less
```diff
@@ -23,7 +23,7 @@
   z-index: 99;
   position: absolute;
   top: -10px;
-  right: 15px;
+  right: 25px;
```
CITguy: FYI:

### BEFORE
* menu is partially covering the cog

screen shot 2016-06-13 at 4 20 01 pm


### AFTER

* menu isn't covering the cog

screen shot 2016-06-13 at 4 22 18 pm

utils/rx-page-objects/src/rxDatePicker.page.js
```diff
@@ -275,7 +275,8 @@ var rxDatePicker = {
 
         set: function (value) {
             this.open();
-            rxSelect.initialize(this.rootElement.element(by.model('currentMonth'))).select(value);
+            var slowClick = false;
+            rxSelect.initialize(this.rootElement.element(by.model('currentMonth'))).select(value, slowClick);
```
CITguy: I'm not sure what value you are gaining with `slowClick`. I'd be more inclined to use a var for the ElementFinder. 

```javascript
var elem = this.rootElement.element(by.model('currentMonth'));
rxSelect.initialize(elem).select(value, true);
```

utils/rx-page-objects/src/rxDatePicker.page.js
```diff
@@ -295,7 +296,8 @@ var rxDatePicker = {
 
         set: function (value) {
             this.open();
-            rxSelect.initialize(this.rootElement.element(by.model('currentYear'))).select(value);
+            var slowClick = false;
+            rxSelect.initialize(this.rootElement.element(by.model('currentYear'))).select(value, slowClick);
```
CITguy: same as above

utils/rx-page-objects/src/rxStatusColumn.page.js
```diff
@@ -190,7 +190,19 @@ var rxStatusColumn = {
...
+                                // In case the user wants to check another status column
+                                // directly above this one, we'll need to open the popover,
+                                // pull the text out, then hover somewhere else and return
+                                // the text. That way, the popup is gone and they can hover
+                                // over the cell above (which would have been obscured by
+                                // the tooltip, had we'd not hovered over another element).
+                                return tooltip.rootElement.getText().then(function (text) {
+                                    browser.actions().mouseMove($('body')).perform();
+                                    return text;
+                                });
```
CITguy: @Droogans: can you explain this to me IRL. I'm having a hard time following the logic.

Droogans: In an end to end test, I was getting the text of a popup. The next test, I was doing the same thing for the text of the popup of the row above.

screen shot 2016-06-13 at 16 37 48

As you can see here, the `DELETED` popup is covering the "Active" row above it. I was not able to hover over it because the pop up was obscuring it.

So, now I get the text, hover over some other place (the `$('body')`), and then return control back to the caller.

I'm all for suggestions on making this more clear in the comments in the source code. I certainly struggled to write it, and felt it wasn't as good as it could be.

CITguy: So, you're essentially:

0. grab text
1. stop hovering
2. repeat

Droogans: I'll summarize appropriately. I think less is more in this case.

CITguy:
LGTM

PR #1724, opened 2016-06-17
docs(rxToggleSwitch): rename 'var switch = ' in @example

ericcarraway:
This PR is a quick improvement to the example documentation for `rxToggleSwitch`.
http://rackerlabs.github.io/encore-ui/rx-page-objects/rxToggleSwitch.html

Because `switch` is a reserved word in JavaScript,
`var switch = ...`
would be better as
`var mySwitch = ...` 

This caused an issue for @arc6789 this morning.

screen shot 2016-06-17 at 11 03 38 am

screen shot 2016-06-17 at 11 08 22 am

### LGTMs
- [x] Dev LGTM


CITguy:
LGTM

PR #1730, opened 2016-06-27
Last minute stuff

Droogans:
Note: ~~All deleted code from the page objects were private-only. There is no deprecation warnings possible (that'd show up in the documentation, anyway).~~ 

✅ ~~I'll update the code that was publicly listed in the documentation as deprecated.~~

I also got the rx-page-objects project up to the latest lodash.

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM


.gitignore
```diff
@@ -37,6 +37,7 @@ results
 /tmp/
 /bower/
 /dist/
+/utils/rx-page-objects/node_modules/*
```
Droogans: This allows contributors to run `npm install` inside of the `utils/rx-page-objects` directory, and node will refer to those versions of installed dependencies when requiring lodash and moment in page object code.

This is nice since it allows the dev dependencies for just rx-page-objects to be run and tested with it's own version of lodash, outside of our build system/dev code/etc.

utils/rx-page-objects/src/rxDatePicker.page.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 var moment = require('moment');
+var _ = require('lodash');
```
CITguy: How was this file working without this require?!

utils/rx-page-objects/package.json
```diff
@@ -10,8 +10,7 @@
   },
   "dependencies": {
     "moment": "2.13.0",
-    "lodash": "2.4.1",
-    "node-timing.js": "1.1.0"
```
Droogans: `node-timings.js` was used to report performance metrics from selenium to the user. This isn't as important since use of newrelic is so widespread inside of our teams. It's better to have folks use the superior version of this feature via newrelic.

utils/rx-page-objects/src/rxMisc.page.js
```diff
@@ -198,31 +198,6 @@ exports.rxMisc = {
...
-     * Selenium test. Can return just one item of interest, or can accept a list of items to return
-     * if you need many different timing metrics. If no parameter is used, all timing metrics are returned.
-     * @function
-     * @param {String|Array} [keys] - Key to get (string), keys to get (Array), or all (`undefined`).
-     * @returns {Object} An object of timings that can be used to infer browser render performance.
-     * @example
-     * it('should have loaded the page in less than two seconds', function () {
-     *     expect(encore.rxMisc.getPerformanceMetrics('loadTime')).to.eventually.be.under(2000);
-     * });
-     */
```
Droogans: Ah, I see now that this was not a private function. I was mistaken. I will update the wiki to reflect this and make a PR against the master branch.

utils/rx-page-objects/src/rxDatePicker.page.js
Droogans: The protractor conf file includes it in the `onPrepare` hook. It's meant for midway files only, but globals, you know?

.gitignore
CITguy: Are we doing any of this with our current grunt tasks?

utils/rx-page-objects/package.json
```diff
@@ -10,8 +10,7 @@
...
+    "lodash": "4.13.1"
```
CITguy: Will this force consumers to use lodash 4.x for their midways?

Droogans: No, it will force any code running under the `rx-page-objects` directory to use lodash 4.x though.

CITguy:
LGTM

parlarjb:
LGTM

utils/rx-page-objects/package.json
100stacks: @Droogans, it looks like we do need lodash 4.x to run midways, unless I'm missing something. 😕 

I had to:
 * `$ cd utils/rx-page-objects`
 * ` $ npm install`
 *  `$ cd ../..`
 *  `$ ./node_modules/.bin/protractor`

Midways work with those extra steps, which we should document.

PR #1739, opened 2016-07-08
Ng disabled

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-717

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxModalAction/scripts/rxModalAction.js
+++ b/src/components/rxModalAction/scripts/rxModalAction.js
@@ -55,18 +55,18 @@ angular.module('encore.ui.rxModalAction')
  * cancel the given action.  For example, you may use this to return an
  * indeterminate component to a previous state (e.g. toggle switches).
  *
- * @param {Function=} [preHook]
+ * @param {Function=} preHook
  * Function to call when a modal is opened
- * @param {Function=} [postHook]
+ * @param {Function=} postHook
  * Function to call when a modal is submitted (not called when modal cancelled)
- * @param {Function=} [dismissHook]
+ * @param {Function=} dismissHook
  * Function to call when a modal is dismissed (not called when modal submitted)
- * @param {String=} [templateUrl]
+ * @param {String=} templateUrl
  * URL of template to use for modal content
- * @param {*=} [disable-esc]
+ * @param {*=} disable-esc
  * If the `disable-esc` attribute is present, then "Press Esc to close" will be
  * disabled for the modal. This attribute takes no values.
- * @param {Expression=} [ngDisabled=false]
+ * @param {Boolean=} [ngDisabled=false]
  * If the expression evaluates to true, then the link for opening the modal will
  * be disabled.
  *
--- a/src/components/rxModalAction/scripts/rxModalAction.js
+++ b/src/components/rxModalAction/scripts/rxModalAction.js
@@ -5,55 +5,70 @@ angular.module('encore.ui.rxModalAction')
  * @restrict E
  * @scope
  * @description
- * Link which will show a modal window on click, and handle callbacks for pre/post modal actions
+ * Link which will show a modal window on click, and handle callbacks for
+ * pre/post modal actions
  *
  * # Template URL
  *
- * Due to the complex nature of the information passed into modal windows, HTML is handled via a template (versus
- * transcluding the content).
+ * Due to the complex nature of the information passed into modal windows,
+ * HTML is handled via a template (versus transcluding the content).
  *
- * One benefit is that this allows for multiple actions to re-use the same template. It also allows modal content to
- * live in a separate file, which can be helpful for maintainability if the modal HTML is complex. While this can be
- * done via `ng-include`, it would be a little extra work for a common scenario.
+ * One benefit is that this allows for multiple actions to re-use the same
+ * template. It also allows modal content to live in a separate file, which
+ * can be helpful for maintainability if the modal HTML is complex. While this
+ * can be done via `ng-include`, it would be a little extra work for a common
+ * scenario.
  *
  * # Hooks
  *
- * `rxModalAction` allows you to define functions to be called at different points of the modal lifecycle.  These hooks
- * are optional and the modal window is fully functional without any being defined.
+ * `rxModalAction` allows you to define functions to be called at different
+ * points of the modal lifecycle.  These hooks are optional and the modal window
+ * is fully functional without any being defined.
  *
  * ## Pre-hook
  *
- * The `pre-hook` function is called before the modal window is shown.  Use a this to populate field information inside
- * of the modal.  This is useful when you have information you don't want loaded when the page is first opened, but do
- * need for the modal.  It is also useful for dynamic information that is based on actions taken.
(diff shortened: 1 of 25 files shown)


src/elements/Buttons/scripts/rxButton.js
```diff
@@ -50,8 +50,17 @@ angular.module('encore.ui.elements')
...
+            disable: '=?', // **DEPRECATED** - Remove in 3.0.0
+            ngDisabled: '=?',
             classes: '@?'
         },
+        link: function (scope) {
+            // Support [ng-disabled] OR [disabled] (for backward compatibility)
+            scope.$watch(function () {
+                return (scope.disable || scope.ngDisabled);
+            }, function (newVal) {
+                scope.isDisabled = newVal;
```
CITguy: Use `isDisabled` to disable the button in the template.

src/components/rxModalAction/scripts/rxModalAction.js
```diff
@@ -93,7 +108,13 @@ angular.module('encore.ui.rxModalAction')
             scope.classes = attrs.classes;
 
             attrs.$observe('disabled', function (newValue) {
-                scope.isDisabled = scope.$eval(newValue);
+                if (typeof newValue === 'boolean') {
+                    scope.isDisabled = newValue;
+                } else {
+                    // TODO: remove in 3.0.0
+                    // DEPRECATED: override [disabled] behavior
+                    scope.isDisabled = scope.$eval(newValue);
```
CITguy: The ~~wrong~~ old way using `disabled="true"`` doesn't result in a proper Boolean value, so we have to assume that it's an angular expression and we `$eval()` to get the evaluated result.

src/components/rxModalAction/scripts/rxModalAction.js
```diff
@@ -93,7 +108,13 @@ angular.module('encore.ui.rxModalAction')
             scope.classes = attrs.classes;
 
             attrs.$observe('disabled', function (newValue) {
-                scope.isDisabled = scope.$eval(newValue);
+                if (typeof newValue === 'boolean') {
+                    scope.isDisabled = newValue;
```
CITguy: `ng-disabled` (the correct way to disable) will result in a proper boolean value, so we check here for the proper value

src/elements/Buttons/scripts/rxButton.js
```diff
@@ -38,7 +38,7 @@ angular.module('encore.ui.elements')
  * @param {String} loadingMsg Text to be displayed when an operation is in progress.
  * @param {String} defaultMsg Text to be displayed by default when no operation is in progress.
  * @param {Boolean=} [toggle=false] When true, the button will display the loading text.
- * @param {Boolean=} [disable=false] When true, the button will be disabled.
+ * @param {Boolean=} [ngDisabled=false] When true, the button will be disabled.
```
CITguy: the old `disable` param still works, but we've changed the documentation to the desired param to deter new code from using the old param

src/components/rxModalAction/scripts/rxModalAction.js
```diff
@@ -5,55 +5,70 @@ angular.module('encore.ui.rxModalAction')
...
+ * @param {Function=} [postHook]
+ * Function to call when a modal is submitted (not called when modal cancelled)
+ * @param {Function=} [dismissHook]
+ * Function to call when a modal is dismissed (not called when modal submitted)
+ * @param {String=} [templateUrl]
+ * URL of template to use for modal content
+ * @param {*=} [disable-esc]
+ * If the `disable-esc` attribute is present, then "Press Esc to close" will be
+ * disabled for the modal. This attribute takes no values.
+ * @param {Expression=} [ngDisabled=false]
```
CITguy: the old `disabled` param still works, but the documentation was updated to the new `ng-disabled` param to deter new code from using the old param

100stacks:
@CITguy, did you want this update to push to `2.x (master)` or `3.x`?  The JIRA story has it listed under the `3.x` Epic.

src/components/rxToggleSwitch/scripts/rxToggleSwitch.js
```diff
@@ -65,8 +66,15 @@ angular.module('encore.ui.rxToggleSwitch')
                 scope.state = ngModelCtrl.$viewValue ? 'ON' : 'OFF';
             };
 
+            scope.$watch(function () {
+                return (scope.disabled || scope.ngDisabled);
+            }, function (newVal) {
+                // will be true, false, or undefined
+                scope.isDisabled = newVal;
```
parlarjb: Is this what we're using for the initialization of `isDisabled`? We don't give it an explicit initial value anywhere else, so I'm assuming so.

parlarjb:
LGTM, aside from the one question above

src/utilities/Permission/scripts/Permission.js
```diff
@@ -30,8 +30,8 @@ angular.module('encore.ui.utilities')
      * @description Takes a function and a list of roles, and returns the
      * result of calling that function with `roles`, and comparing to userRoles().
      *
-     * @param {function} fn - Comparison function to use. _.any, _.all, etc.
-     * @param {array} roles - List of desired roles
+     * @param {Function} fn Comparison function to use. _.any, _.all, etc.
+     * @param {String[]} roles List of desired roles
      */
     var checkRoles = function (roles, fn) {
```
Droogans: Should the documentation for the params be listed in the same order that they're passed into the `checkRoles` function?

Also, I believe there should be a name after the `{String[]} roles - ` param.

src/components/rxToggleSwitch/scripts/rxToggleSwitch.js
CITguy: Yes, the first argument of `$watch()` will always return a boolean value and always executes at least once. So, when the evaluation of `scope.disabled || scope.ngDisabled` returns, `scope.isDisabled` will be initialized with the correct value. When we remove support for `scope.disabled` in 3.0.0, we'll explicitly initialize its value instead.

src/utilities/Permission/scripts/Permission.js
CITguy: Given that some of the arguments can be defaulted if not passed (making them optional), I documented the required params before the optional ones. This can be easily updated, though. I'm not married to the order.

The name after `{String[]}` is `roles`

CITguy:
@100stacks, This PR is fine for 2.x updates because it is backward compatible and does not break existing functionality. The existing JIRA story should be updated so that it is for `2.x` and we'll need a story to remove the old functionality in 3.0.0, later.

src/utilities/Permission/scripts/Permission.js
Droogans: Ugh, ok, I just reread that and realized that the dash is optional.

100stacks:
@CITguy, I added a story to remove the deprecated `disabled` attribute during our `EncoreUI 3.x` release cycle.

https://jira.rax.io/browse/FRMW-979

PR #1741, opened 2016-07-15
chore(node): Use version 6.3.0

Droogans:
JIRA: N/A

### LGTMs
- [ ] Dev LGTM


Droogans:
This would force consumers of rx-page-objects to upgrade to node 6.3.0, which is premature.

PR #1748, opened 2016-07-18
test(typescript): For rx-page-objects

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-743

### LGTMs
- [x] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [ ] QE LGTM


utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -0,0 +1,273 @@
...
+     * @description
+     * Will select the last day of the current month seen on the picker calendar.
+     * Does not accept arguments. Set the calendar month yourself with
+     * {@link rxDatePicker#month} and {@link rxDatePicker#year} first, so that logic
+     * picks the correct date.
+     */
+    private _selectLastDayOfCurrentMonth(): Promise<void> {
+        this.open();
+        return (this.month as Promise<string>).then(function (month) {
+            return (this.year as Promise<string>).then(function (year) {
```
parlarjb: Does Typescript know enough to automatically assign a `string` type to the `year` parameter, based on the `Promise<string>`?

Droogans: If you do something like

```ts
datepicker.year = 1;
```

You'll get a compilation error, saying that the `year` setter expects type `String | Promise<string>`.

Droogans: Also, this is me `get`ting both the month and the year, which returns a `Promise<string>`.

The setter expects a string, though.

parlarjb: Sorry, that's not what I meant. Let me be more clear. (And this is all just a curiosity thing on my part)

You pass an anonymous function to the `.then`. This anonymous function has the signature `function (year) {`, with no explicit typing.

Will typescript implicitly assign a `string` type to that `year` argument, because it's knowledgeable enough about promises to know that `(this.year as Promise<string>)` implies an eventual promise resolution that will pass in a `string`?

parlarjb:
LGTM

utils/rx-page-objects/src/rxDatePicker.page.ts
Droogans: :100:

100stacks:
@Droogans, do you think we can ask @jj4th to review?  I recall he was interested in refactoring tests to  `TypeScript`.

utils/rx-page-objects/test/rxDatePicker.midway.js
```diff
@@ -116,48 +105,37 @@ describe('rxDatePicker', function () {
...
-                var lastOfMonth = moment(today).endOf('month').format(isoFormat);
-                expect(yearMonthDayString(date)).to.equal(lastOfMonth);
-            });
+            var lastOfMonth = moment(today).endOf('month').format(isoFormat);
+            picker.date = lastOfMonth;
+            expect(picker.date).to.eventually.equal(lastOfMonth);
         });
 
         it('should update the date to one month ago', function () {
             var previousMonth = moment(today).subtract(1, 'months').format(isoFormat);
```
CITguy: Why not convert to using `lastMonth` like you did with `nextMonth` above?

utils/rx-page-objects/test/rxMetadata.midway.js
```diff
@@ -1,4 +1,5 @@
 var _ = require('lodash');
+var promise = require('selenium-webdriver').promise;
```
CITguy: Can this be made global rather than defining it in every file?

utils/rx-page-objects/test/rxDatePicker.midway.js
Droogans: That'd make it fit on one line, that's for sure.

```js
expect(picker.date).to.eventually.equal(lastMonth.format(isoFormat));
```

utils/rx-page-objects/test/rxMetadata.midway.js
Droogans: This was written against protractor 4.0.0, but with the latest (4.0.3), `protractor.promise.all` works again, so this can go away.

It was originally filed as a defect against the project when I was first making this PR.

utils/rx-page-objects/test/rxDatePicker.midway.js
CITguy: Specifically, I was referring to [this change](https://github.com/rackerlabs/encore-ui/pull/1748/files?diff=unified#diff-7446cc08a727b0b9df60e5ac29b53fd6L119)

utils/rx-page-objects/.gitignore
```diff
@@ -2,4 +2,11 @@ doc/*
...
 !doc/layout.tmpl
 !doc/README.md
-node_modules/
\ No newline at end of file
+node_modules/
+typings/
+
+# these are typescript now
+# eventually, all src/*.page.js files
+# can be ignored with a glob.
```
CITguy: Non-blocking, but I'd be more inclined to separate source and build files by use of a modified directory structure, but this would be an update for later. I'd recommend doing...

```shell
src/scripts/   # *.ts or *.js
build/scripts/ # *.js only
```

utils/rx-page-objects/src/rxComponent.ts
```diff
@@ -0,0 +1,19 @@
...
+
+'use strict';
+
+import {browser} from 'protractor/globals';
+import {ElementFinder} from 'protractor';
+let ElementFinderConstructor: typeof ElementFinder = require('protractor/built/element').ElementFinder;
+
+export class rxComponentElement extends ElementFinderConstructor {
+    constructor(rootElement: ElementFinder) {
+        super(browser, rootElement.elementArrayFinder_);
```
CITguy: Is `elementArrayFinder_` a private function/property that is likely to change? (i.e. should we be using it?)

utils/rx-page-objects/src/rxComponent.ts
```diff
@@ -0,0 +1,19 @@
+///<reference path="../typings/globals/node/index.d.ts"/>
+
+'use strict';
+
+import {browser} from 'protractor/globals';
+import {ElementFinder} from 'protractor';
+let ElementFinderConstructor: typeof ElementFinder = require('protractor/built/element').ElementFinder;
```
CITguy: Is there any advantage to specifying the type as `typeof ElementFinder` here? We're only using it for class inheritance.

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -0,0 +1,273 @@
...
+    public rootElement: ElementFinder;
+    private tblCurrentMonthDays: ElementArrayFinder;
+
+    /**
+     * @param {ElementFinder} [rxDatePickerElement=$('rx-date-picker')]
+     * ElementFinder to be transformed into an rxDatePicker page object
+     * @returns {rxDatePicker}
+     */
+    constructor(rxDatePickerElement) {
+        if (rxDatePickerElement === undefined) {
```
CITguy: I'd add a note about this `if()` block. It exists because the version of Node we use doesn't support ES6 [Default Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters). Otherwise we'd have...

```javascript
constructor(rxDatePickerElement = $$('rx-date-picker').first()) { ... }
```

utils/rx-page-objects/test/rxDatePicker.midway.js
CITguy: Any update?

utils/rx-page-objects/test/rxMetadata.midway.js
CITguy: Can we make bump to 4.0.3 and make this go away?

utils/rx-page-objects/src/rxComponent.ts
Droogans: We shouldn't. This got fixed in 4.0.1 and beyond. They allow you to pull in the `ElementFinder` class object for extension purposes that doesn't require that you provide a `parentElementArrayFinder` parameter here!

utils/rx-page-objects/src/rxComponent.ts
Droogans: You have to since it's getting pulled from a javascript file. If you `require` something from a javascript file you have to tell typescript what type it is (as opposed to using `import` to pull in typescript files that have type information included in them.

Again, this is now in 4.0.1 and beyond. This can go away on an update to the newest version of protractor.

utils/rx-page-objects/src/rxDatePicker.page.ts
Droogans: I don't know if I should reply with the 👍 emoji or the 😕 since I'm bummed that I can't use it today.

jj4th: Might not be the place to draw this line, but there are some pretty good arguments for `typeof variable == "undefined"`.  Conveniently... `_.isUndefined()` performs this style of check.

Droogans: I think that way makes the most sense. I'll update it to use a typed parameter in the constructor, with a `if (!DatePickerElement)` test for existence.

utils/rx-page-objects/test/rxDatePicker.midway.js
Droogans: I'm going to bump this whole thing to 4.0.3 and clean it up right now and rebase my other PRs on top of it to slim this one up a bit.

I'll also revisit the tests to make sure there's no unnecessary re-declaration of the `lastMonth`/`nextMonth` vars.

Droogans:
Moving this to https://github.com/rackerlabs/encore-ui/pull/1781

utils/rx-page-objects/src/rxComponent.ts
Droogans: @CITguy This was the only thing I was unable to change in protractor 4.0.3. It is still required.

Also, this is a foundational property on all instances of ElementFinders. If this property goes away or changes, then the call signature of the ElementFinder constructor would have to be updated in protractor....and here as well. I have no issues using it.

PR #1754, opened 2016-07-20
FRMW-761: Refactor rxCharacterCount to element

100stacks:
JIRA: https://jira.rax.io/browse/FRMW-761

### LGTMs
- [x] Dev/Dev+Vidyo LGTM
- [x] Design LGTM
- [x] QE LGTM

---
Refactor `Components > rxCharacterCount` to `Elements > Forms > Character Count`


100stacks:
## Before `Components` > `rxCharacterCount`

rxcharactercount - encoreui 2016-07-20 11-32-58

## After `Elements` > `Forms` > `Character Count` (section added)

character count - encoreui 2016-07-20 11-41-07

## API Docs 
encore ui developer documentation rxcharactercount 2016-07-20 11-43-08

## After `Elements` > `Forms` (full demo page)
forms - encoreui 2016-07-20 11-39-48

Droogans:
We pass in a `?debug=true` param in urls that we visit during the e2e tests. It's intended to hide/show extra demo components so that we can show just one or two instances to users, but then test every permutation of feature combinations in test.

Do you think you could do that here? I think showing users half a dozen character count elements isn't a good idea.

100stacks:
@Droogans: I checked locally and we are passing that parameter:
```
http://localhost:9001/#/elements/Forms?debug=true
```
In this block I wonder if it's the `Array.join()` adding a char at a time that's causing the issue...it's weird because it wasn't an issue when `rxCharacterCount` was on it's own page...it just zipped through. 🎱 

https://github.com/rackerlabs/encore-ui/blob/master/utils/rx-page-objects/src/rxCharacterCount.exercise.js#L69-L101

Droogans:
@100stacks What I meant is that we are already passing in that debug url parameter to the e2e tests, do you think you could hide all but say, one character count instance if the parameter is not there?

That way, users navigating to the page see one rxCharacterCount, but the tests will see many more.

100stacks:
@Droogans, okay I'll give that a try! 👍

100stacks:
@khaziwallis @manojdas2212 @satyamkotakonda: While @Droogans and I work through updating the test for `Forms` -> `Character Count`, I pushed the PR to show you the basic structure of the PRs.  

Please see:
* `Before`, `After`, and  `API` screenshots in https://github.com/rackerlabs/encore-ui/pull/1754#issuecomment-234030091.  **PLEASE NOTE:** That the screenshots are added in a `comment` AFTER pushing the PR, so that the screenshots do not overly expand the initial commit message in Slack.
* [`2 commits`](https://github.com/rackerlabs/encore-ui/pull/1754/commits) - we do this to maintain the `git` history of the files.
* The original `rxCharacterCount` was a single demo on a single page, which was refactored into `6` single atomic demo examples.
* Also notice the `Labels` added:

    screen shot 2016-07-20 at 3 11 59 pm
* I also prefer to add a short description in the initial comment block so that others get a quick summary/purpose in Slack when pushed:

> Refactor Components > rxCharacterCount to Elements > Forms > Character Count



Please let me know if you have any questions or feedback.  👍

glynnis:
screen shot 2016-07-21 at 2 48 52 pm

Looks like we've gone from too many examples to too few :) Especially given the content of the last bullet point. I'd add back just the first example with demo code (the one that said "Default Values" on the component page, but it doesn't need a header now), and maybe add a bullet point to the current list that says, "For more options regarding rxCharacterCount functionality, see API docs."

Once we have one demo + code block back here, then I think we're good on the design front.

100stacks:
@glynnis, no problem can do. 👍 

I added a `testing` label as I'm testing out why local builds pass, but not on the CI Servers.

src/elements/Forms/docs/forms.docs.html
```diff
@@ -405,11 +410,37 @@ <h2 class="clear"><rx-permalink>Directives</rx-permalink></h2>
...
+    <h4>Count leading and trailing spaces</h4>
+    <rx-example name="rxCharacterCount.trailingSpaces"></rx-example>
+
+    <h4>Accounts for initial values</h4>
+    <rx-example name="rxCharacterCount.initialValues"></rx-example>
+
+    <h4>With highlighting</h4>
+    <rx-example name="rxCharacterCount.highlighting"></rx-example>
+  </div>
+</rx-debug>
```
100stacks: @Droogans, we're still taking performance hits when I merge `rxCharacterCount` on to the `Forms` demo page, even when I try wrapping the `Character Count` demos in `<rx-debug>`.

100stacks:
## `Forms` -> `Character Count` Midways
screen shot 2016-07-22 at 5 18 05 pm
**7 Minutes!**

## Standalone `rxCharacterCount` Midways
screen shot 2016-07-22 at 5 18 18 pm
**29 Seconds!**

@Droogans, I'm leaning toward your theory that the Angular `$digest` cycle is getting overloaded on the `Forms` demo page, or possibly Protractor as it iterates over the exercises below.

The major time increases are with these exercises using `Array.join()` to append `a` character over xx iterations : 
```js
        var belowNearLimitLength = options.maxCharacters + 1 - options.nearLimit;
        it('should not set the near-limit class when ' + belowNearLimitLength + ' characters are entered', function () {
            component.comment = Array(belowNearLimitLength).join('a');
            expect(component.isNearLimit()).to.eventually.be.false;
        });

        var atNearLimitLength = options.maxCharacters + 2 - options.nearLimit;
        it('should set the near-limit class when ' + atNearLimitLength + ' characters are entered', function () {
            component.comment = Array(atNearLimitLength).join('a');
            expect(component.isNearLimit()).to.eventually.be.true;
        });

        var aboveNearLimitLength = options.maxCharacters + 3 - options.nearLimit;
        it('should set the near-limit class when ' + aboveNearLimitLength + ' characters are entered', function () {
            component.comment = Array(aboveNearLimitLength).join('a');
            expect(component.isNearLimit()).to.eventually.be.true;
        });

        var atLimit = options.maxCharacters + 1;
        it('should not set the over-limit class when ' + atLimit + ' characters are entered', function () {
            component.comment = Array(atLimit).join('a');
            expect(component.isOverLimit()).to.eventually.be.false;
        });

        it('should have zero remaining characters', function () {
            expect(component.remaining).to.eventually.equal(0);
        });

        var overLimit = options.maxCharacters + 2;
        it('should set the over-limit class when ' + overLimit + ' characters are entered', function () {
            component.comment = Array(overLimit).join('a');
            expect(component.isOverLimit()).to.eventually.be.true;
        });
```
Source: https://github.com/rackerlabs/encore-ui/blob/master/utils/rx-page-objects/src/rxCharacterCount.exercise.js#L69-L101

Droogans:
@100stacks Great work on narrowing the scope for troubleshooting.

I think we can clear this up by attempting to use a couple different approaches next week. Namely, dropping character literals like

```js
component.comment = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
```

I am going to go on the record as of now and guess that this will not save us any time. `Array(255).join('a')`, as an operation, might take a few milliseconds. However, entering that text (and triggering digests!) in the comment field is what I assume is causing the performance hit. That's why it's been exacerbated so much on a larger demo page.

We should start by making a text area, and doing a lot of the same types of large input tests to it as well. If that causes significant delays in the test execution time, then we know it's probably a Selenium thing.

Droogans:
I'm not going to rebase this since I saw you and Ryan working on the performance issues for this PR yesterday. @100stacks please remember to rebase this against master when you push up any new work, otherwise codeship will continue to give you a failed build!

glynnis:
This has a design LGTM :+1:

100stacks:
@glynnis, based on feedback from @CITguy, I wrapped the `rxCharacterCount` example in a `<rx-form>` to show a simple implementation rather just a code snippet.

Update screenshots above in https://github.com/rackerlabs/encore-ui/pull/1754#issuecomment-234030091, provided here for quick reference:

screen shot 2016-07-26 at 6 18 25 pm

src/elements/Forms/docs/forms.docs.html
```diff
@@ -405,12 +410,92 @@ <h2 class="clear"><rx-permalink>Directives</rx-permalink></h2>
...
+    Please see the example below for implementing a character counter on a
     textarea or text input field.
   </li>
+  <li>
+    For more options regarding rxCharacterCount functionality, see API docs.
+  </li>
 </ul>
 
+<rx-example name="rxCharacterCount.simple"></rx-example>
+
```
100stacks: Wrapped our `rxCharacterCount` tests in `<rx-debug>` to decouple test demo snippets from consumer view.

We did this to improve test performance, by adding:

```js
ng-model-options="{ debounce: 100 }"
```

src/elements/Forms/docs/forms.docs.html
```diff
@@ -405,12 +410,92 @@ <h2 class="clear"><rx-permalink>Directives</rx-permalink></h2>
...
+<rx-debug>
+  <div ng-controller="rxCharacterCountCtrl">
+    <h4>Character Count: Default Values</h4>
+    <div>
+      <textarea
+        ng-model="data.comment1"
+        ng-model-options="{ debounce: 100 }"
```
100stacks: Wrapped our `rxCharacterCount` tests in `<rx-debug>` to decouple test demo examples from consumer view.

We did this to improve test performance, by adding:

```js
ng-model-options="{ debounce: 100 }"
```

glynnis:
LGTM still

utils/rx-page-objects/src/rxCharacterCount.exercise.js
```diff
@@ -43,81 +51,130 @@ exports.rxCharacterCount = function (options) {
...
 
-        var atLimit = options.maxCharacters + 1;
-        it('should not set the over-limit class when ' + atLimit + ' characters are entered', function () {
-            component.comment = Array(atLimit).join('a');
-            expect(component.isOverLimit()).to.eventually.be.false;
+
+        describe('when ' + belowNearLimitLength + ' characters are entered', function () {
+
+            beforeEach(function () {
+                component.comment = 'a'.repeat(belowNearLimitLength);
```
CITguy: Yay! No more `Array(n+1).join('a')`

CITguy:
LGTM

CITguy:
@Droogans, is it too late to get your LGTM?

CITguy:
LGTM (as @Droogans proxy)

100stacks:
@CITguy, thanks for pairing up on refactoring! 👍 💯

PR #1773, opened 2016-08-18
Grunt pipeline refactor

CITguy:
JIRA: n/a

The work in this PR came about after noticing that the generated javascript was duplicating DI dependencies. This lead to tweaking the grunt configs, which lead to the discovery that we're generating distributable files that aren't being used. Additionally, several other little things were discovered and corrected.

### LGTMs
- [ ] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [ ] QE LGTM

### DIFF
--- a/grunt-tasks/templates/encore-ui.ejs.js
+++ b/grunt-tasks/templates/encore-ui.ejs.js
@@ -1,0 +1,9 @@
+angular.module('encore.ui', [
+    'cfp.hotkeys',
+    'ui.bootstrap',
+    'encore.ui.tpls',
+    <%= _.uniq(config.srcModules).join(',\n    ') %>
+])
+angular.module('encore.ui.tpls', [
+    <%= _.sortBy(_.flatten(config.tplModules)).join(',\n    ') %>
+]);
--- a/grunt-tasks/util/config.js
+++ b/grunt-tasks/util/config.js
@@ -1,3 +1,11 @@
+var _banner = `/*
+ * <%= pkg.name %>
+ * <%= pkg.homepage %>
+ *
+ * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>
+ * License: <%= pkg.license %>
+ */`;
+
 module.exports = {
     dir: {
         app: 'src',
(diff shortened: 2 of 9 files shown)


grunt-tasks/templates/encore-ui.ejs.js
```diff
@@ -0,0 +1,9 @@
+angular.module('encore.ui', [
+    'cfp.hotkeys',
+    'ui.bootstrap',
+    'encore.ui.tpls',
+    <%= _.uniq(config.srcModules).join(',\n    ') %>
+])
+angular.module('encore.ui.tpls', [
+    <%= _.sortBy(_.flatten(config.tplModules)).join(',\n    ') %>
+]);
```
CITguy: Rather than defining all of this in `grunt-tasks/utils/config.js` using javascript, this file makes it easier to explicitly define how the module and its dependencies are written.

grunt-tasks/util/config.js
```diff
@@ -18,26 +26,9 @@ module.exports = {
...
-            '\'<%= config.dependencies.join("\',\'") %>\']);',
-        banner: [
-            '/*',
-            ' * <%= pkg.name %>',
-            ' * <%= pkg.homepage %>\n',
-            ' * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>',
-            ' * License: <%= pkg.license %>',
-            ' */\n'
-        ].join('\n')
-    },
```
CITguy: The majority of the `meta` property was used to write out the generated JS files. But now that we have `grunt-tasks/templates/encore-ui.ejs.js`, most of it is no longer necessary.  The banner configuration was the only piece that was being used in multiple locations, so it was renamed from `config.meta.banner` to `config.banner`.

grunt-tasks/util/config.js
```diff
@@ -1,3 +1,11 @@
+var _banner = `/*
+ * <%= pkg.name %>
+ * <%= pkg.homepage %>
+ *
+ * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>
+ * License: <%= pkg.license %>
+ */`;
```
CITguy: Instead of ugly `Array.prototype.join` nonsense, we can use ES6, multi-line, backtick strings to define this.

grunt-tasks/util/config.js
```diff
@@ -18,26 +26,9 @@ module.exports = {
         port: 9001
     },
     latestTag: '', // filled in by shipit task
-    moduleName: 'encore.ui',
```
CITguy: This was only being used in `config.meta`. Now it's explicit in `grunt-tasks/templates/encore-ui.ejs.js`.

CITguy:
Crap! Submitted from fork. Will resubmit.

PR #1774, opened 2016-08-18
Grunt pipeline refactor

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-969

The work in this PR came about after noticing that the generated javascript was duplicating DI dependencies. This lead to tweaking the grunt configs, which lead to the discovery that we're generating distributable files that aren't being used. Additionally, several other little things were discovered and corrected.

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

### DIFF
--- a/grunt-tasks/modules.js
+++ b/grunt-tasks/modules.js
@@ -131,18 +131,21 @@ module.exports = function (grunt) {
 
         // Set the concat task to concatenate the given src modules
         var srcFiles = _.pluck(modules, 'srcFiles'); // OK
+        var tplJsFiles = _.pluck(modules, 'tplJsFiles'); // OK
 
         srcFiles = [
+            'grunt-tasks/templates/encore-ui.ejs.js',
             // Prepend override files in order to achieve custom functionality outside of the encore ecosystem
             'src/domainOverride.js',
             // Prepend ngBootstrap overrides to suppress v0.14.3 deprecation warnings
             'src/ngBootstrapOverrides.js'
-        ].concat(srcFiles);
+        ].concat(srcFiles).concat(tplJsFiles);
 
         grunt.config('concat.dist.src', grunt.config('concat.dist.src').concat(srcFiles));
 
-        // Set the concat-with-templates task to concat the given src & tpl modules
-        var tplJsFiles = _.pluck(modules, 'tplJsFiles'); // OK
-        grunt.config('concat.distTpls.src', grunt.config('concat.distTpls.src').concat(srcFiles).concat(tplJsFiles));
+        // TODO: remove in 3.0.0
+        // Both dist.src and distTpls.src should be the same to retain
+        // backward compatibility
+        grunt.config('concat.distTpls.src', grunt.config('concat.dist.src'));
     });
 };
--- a/grunt-tasks/options/concat.js
+++ b/grunt-tasks/options/concat.js
@@ -1,16 +1,20 @@
 module.exports = {
+    // Both dist and distTpls will contain the same content, to retain
+    // backward compatibility
     dist: {
         options: {
-            banner: '<%= config.meta.banner %><%= config.meta.modules %>\n'
+            banner: '<%= config.banner %>\n',
+            process: true
         },
-        src: [], //src filled in by build task
+        src: [], //src filled in by 'modules' task
         dest: '<%= config.dir.dist %>/<%= config.dist.fileName %>.js'
     },
-    distTpls: {
+    distTpls: { // TODO: Remove in 3.0.0
         options: {
-            banner: '<%= config.meta.banner %><%= config.meta.all %>\n<%= config.meta.tplmodules %>\n'
+            banner: '<%= config.banner %>\n',
+            process: true
         },
-        src: [], //src filled in by build task
+        src: [], //src filled in by 'modules' task
         dest: '<%= config.dir.dist %>/<%= config.dist.fileNameTpl %>.js'
     },
     tmpDemosLess: {
--- a/grunt-tasks/templates/encore-ui.ejs.js
+++ b/grunt-tasks/templates/encore-ui.ejs.js
@@ -1,0 +1,9 @@
+angular.module('encore.ui', [
+    'cfp.hotkeys',
+    'ui.bootstrap',
+    'encore.ui.tpls',
+    <%= _.uniq(config.srcModules).join(',\n    ') %>
+])
+angular.module('encore.ui.tpls', [
+    <%= _.sortBy(_.flatten(config.tplModules)).join(',\n    ') %>
+]);
--- a/grunt-tasks/util/config.js
+++ b/grunt-tasks/util/config.js
@@ -1,3 +1,11 @@
+var _banner = `/*
+ * <%= pkg.name %>
+ * <%= pkg.homepage %>
+ *
+ * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>
+ * License: <%= pkg.license %>
+ */`;
+
 module.exports = {
     dir: {
         app: 'src',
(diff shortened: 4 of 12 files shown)


grunt-tasks/templates/encore-ui.ejs.js
```diff
@@ -0,0 +1,9 @@
+angular.module('encore.ui', [
+    'cfp.hotkeys',
+    'ui.bootstrap',
+    'encore.ui.tpls',
+    <%= _.uniq(config.srcModules).join(',\n    ') %>
+])
+angular.module('encore.ui.tpls', [
+    <%= _.sortBy(_.flatten(config.tplModules)).join(',\n    ') %>
+]);
```
CITguy: Rather than defining all of this in `grunt-tasks/utils/config.js` using javascript, this file makes it easier to explicitly define how the module and its dependencies are written.

grunt-tasks/util/config.js
```diff
@@ -18,26 +26,9 @@ module.exports = {
...
-        banner: [
-            '/*',
-            ' * <%= pkg.name %>',
-            ' * <%= pkg.homepage %>\n',
-            ' * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>',
-            ' * License: <%= pkg.license %>',
-            ' */\n'
-        ].join('\n')
-    },
+    banner: _banner,
```
CITguy: The majority of the `meta` property was used to write out the generated JS files. But now that we have `grunt-tasks/templates/encore-ui.ejs.js`, most of it is no longer necessary.  The banner configuration was the only piece that was being used in multiple locations, so it was renamed from `config.meta.banner` to `config.banner`.

grunt-tasks/util/config.js
```diff
@@ -1,3 +1,11 @@
+var _banner = `/*
+ * <%= pkg.name %>
+ * <%= pkg.homepage %>
+ *
+ * Version: <%= pkg.version %> - <%= grunt.template.today("yyyy-mm-dd") %>
+ * License: <%= pkg.license %>
+ */`;
```
CITguy: Instead of ugly `Array.prototype.join` nonsense, we can use ES6, multi-line, backtick strings to define this.

grunt-tasks/util/config.js
```diff
@@ -18,26 +26,9 @@ module.exports = {
         port: 9001
     },
     latestTag: '', // filled in by shipit task
-    moduleName: 'encore.ui',
```
CITguy: This was only being used in `config.meta`. Now it's explicit in `grunt-tasks/templates/encore-ui.ejs.js`.

grunt-tasks/options/concat.js
```diff
@@ -1,16 +1,20 @@
...
-            banner: '<%= config.meta.banner %><%= config.meta.modules %>\n'
+            banner: '<%= config.banner %>\n',
+            process: true
         },
-        src: [], //src filled in by build task
+        src: [], //src filled in by 'modules' task
         dest: '<%= config.dir.dist %>/<%= config.dist.fileName %>.js'
     },
-    distTpls: {
+    distTpls: { // TODO: Remove in 3.0.0
```
CITguy: `bower.json` has `encore-ui-tpls.js` as its main JS file, but we were also distributing `encore-ui.js` with no template cache dependency. The latter is likely not being used, so we're going to transition to have `encore-ui.js` as the main JS (with template cache) and we'll drop `encore-ui-tpls.js`.

This will not officially happen until the 3.0.0 release, because this has the potential to break existing applications that may be hard-coding `encore-ui-tpls.js` in their `index.html`. However, the demo and the karma configurations will use the new name, because it won't break consumers of the framework.

grunt-tasks/modules.js
```diff
@@ -131,18 +131,21 @@ module.exports = function (grunt) {
...
+        var tplJsFiles = _.pluck(modules, 'tplJsFiles'); // OK
 
         srcFiles = [
+            'grunt-tasks/templates/encore-ui.ejs.js',
             // Prepend override files in order to achieve custom functionality outside of the encore ecosystem
             'src/domainOverride.js',
             // Prepend ngBootstrap overrides to suppress v0.14.3 deprecation warnings
             'src/ngBootstrapOverrides.js'
-        ].concat(srcFiles);
+        ].concat(srcFiles).concat(tplJsFiles);
```
CITguy: All source files will include the template js files.

src/utilities/PaginatedItemsSummary/scripts/PaginatedItemsSummary.js
```diff
@@ -11,14 +12,14 @@ angular.module('encore.ui.utilities')
...
     return function (pager) {
-        var template = '<%= first %>-<%= last %> of <%= total %>';
+        var template = '{{first}}-{{last}} of {{total}}';
         if (pager.showAll || pager.itemsPerPage > pager.total) {
-            template = '<%= total %>';
+            template = '{{total}}';
         }
         var firstAndLast = rxPaginateUtils.firstAndLast(pager.currentPage(), pager.itemsPerPage, pager.total);
-        return _.template(template, {
+        return $interpolate(template)({
```
CITguy: Make use of $interpolate, so that `<%= %>` syntax doesn't get processed by grunt

src/elements/AccountInfo/scripts/rxAccountInfo.js
```diff
@@ -62,7 +63,7 @@ angular.module('encore.ui.elements')
                 scope.showCurrentUser = isCloud;
             });
 
-            scope.accountPageUrl = _.template('/accounts/<%= accountNumber %>', scope);
+            scope.accountPageUrl = $interpolate('/accounts/{{accountNumber}}')(scope);
```
CITguy: Make use of $interpolate, so that `<%= %>` syntax doesn't get processed by grunt.

src/utilities/rxFavicon/scripts/rxFavicon.spec.js
```diff
@@ -7,8 +7,8 @@ describe('utilities:rxFavicon', function () {
         local: 'local.png'
     };
 
-    var baseTemplate = _.template('<link rel="icon" type="image/png" href="<%= prod %>"' +
-        ' rx-favicon="{ staging: \'<%= staging %>\', local: \'<%= local %>\' }" />');
+    var baseTemplate = _.template('<link rel="icon" type="image/png" href="${prod}"' +
+        ' rx-favicon="{ staging: \'${staging}\', local: \'${local}\' }" />');
```
CITguy: Convert to `${}` syntax to match ALL other `_.template()` formatting.

grunt-tasks/options/concat.js
```diff
@@ -1,16 +1,20 @@
 module.exports = {
+    // Both dist and distTpls will contain the same content, to retain
+    // backward compatibility
     dist: {
         options: {
-            banner: '<%= config.meta.banner %><%= config.meta.modules %>\n'
+            banner: '<%= config.banner %>\n',
+            process: true
```
CITguy: The only drawback to this option, is that we can no longer use `<%= var %>` style ejs templates in the source code. This is fine, because we can make use of `$interpolate()` in angular or `${var}` syntax in lodash.

Droogans:
> JIRA: n/a

My favorite PRs!

grunt-tasks/options/concat.js
100stacks: @CITguy, suggest you add a check for  `encore-ui-tpls.js` to your `Encore Auditor` for `EncoreUI` `v3.0.0` - https://github.com/rackerlabs/encore-auditor-cli

100stacks:
LGTM 👍

100stacks:
@CITguy, not sure if you want to capture this effort here - https://jira.rax.io/browse/FRMW-969.

CITguy:
@100stacks: yeah, using FRMW-969 as JIRA story
@Droogans: sorry, no longer `JIRA: n/a`

parlarjb:
LGTM

grunt-tasks/options/concat.js
CITguy: Good call. [FRMW-1017](https://jira.rax.io/browse/FRMW-1017) created.

PR #1779, opened 2016-08-24
feat(navItem): Checks url before passing to origin

lentusignavus:
JIRA: https://jira.rax.io/browse/FRMW-952

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

Checks to see if the iframe has url that has been clicked
If it does, do nothing, if it doesn't load into iframe

Defaults to regular link behavior if not in origin

### DIFF
--- a/src/components/rxApp/scripts/rxAppNavItem.js
+++ b/src/components/rxApp/scripts/rxAppNavItem.js
@@ -181,6 +181,28 @@ angular.module('encore.ui.rxApp')
                 }
                 // otherwise, let the default nav do it's thing
             };
+
+            $scope.navigateToApp = function (ev, url) {
+                // We want to control what the click to the <a> tag does
+                // If it is Origin prevent the default click action
+                // otherwise handle the click as normal (implied by the lack of else block)
+                if ($injector.has('oriLocationService')) {
+                    var oriLocationService = $injector.get('oriLocationService');
+                    var currentIframeUrl = oriLocationService.getCanvasURL();
+                    var finalUrl = $scope.getUrl(url);
+
+                    ev.preventDefault();
+                    // Only change the iFrame if the urls are different
+                    if (!_.isEmpty(finalUrl) && currentIframeUrl !== finalUrl) {
+                        oriLocationService.setCanvasURL(finalUrl);
+                    }
+                }
+            };
+
+            $scope.navClickHandler = function (clickEvent, item) {
+                $scope.toggleNav(clickEvent, item.href);
+                $scope.navigateToApp(clickEvent, item.url);
+            }
         }
     };
 });
--- a/src/components/rxApp/scripts/rxAppNavItem.spec.js
+++ b/src/components/rxApp/scripts/rxAppNavItem.spec.js
@@ -5,6 +5,7 @@ describe('encore.ui.rxApp', function () {
 
         var menuItem = {
             href: { tld: 'example', path: 'myPath' },
+            url: 'someapp.com',
             linkText: '1st',
             directive: 'fake-directive',
             visibility: function () {
@@ -295,5 +296,49 @@ describe('encore.ui.rxApp', function () {
                 expect(item.attr('target')).to.equal('_self');
             });
         });
+
+        describe('Origin Navigation', function () {
+            var navigateToAppSpy;
+            var setCanvasURLSpy;
+            var currentCanvasURL = 'someapp.com';
+
+            var fakeOriLocationService = {
+                getCanvasURL: function () {
+                    return currentCanvasURL;
+                },
+                setCanvasURL: function () {}
+            };
+
+            beforeEach(function () {
+                setCanvasURLSpy = sinon.spy(fakeOriLocationService, 'setCanvasURL');
+                provide.factory('oriLocationService', function () {
+                    return fakeOriLocationService;
+                });
+                el = helpers.createDirective(template, compile, scope);
+                navigateToAppSpy = sinon.spy(el.isolateScope(), 'navigateToApp');
+            });
+
+            afterEach(function () {
+                navigateToAppSpy.restore();
+                setCanvasURLSpy.restore();
+            });
+
+            it('should call navigateToApp on click', function () {
+                el.find('.item-link').click();
+                expect(navigateToAppSpy).to.have.been.called;
+            });
+
+            it('should not call setCanvasURL when urls are the same', function () {
+                el.find('.item-link').click();
+                expect(setCanvasURLSpy).to.have.not.been.called;
+            });
+
+            it('should call setCanvasURL when urls are different', function () {
+                currentCanvasURL = 'someotherapp.com';
+                el.find('.item-link').click();
+                expect(setCanvasURLSpy).to.have.been.called;
+                expect(setCanvasURLSpy).to.have.been.calledWith('someapp.com');
+            });
+        })
     });
 });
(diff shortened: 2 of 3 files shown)


src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -181,6 +181,23 @@ angular.module('encore.ui.rxApp')
...
             };
+
+            $scope.navigateToApp = function (ev, url) {
+                //Check if we are in origin;
+                var isOrigin = $injector.has('oriLocationService');
+
+                // We want to control what the click to the <a> tag does
+                // If it is Origin prevent the default click action
+                // otherwise handle the click as normal (implied by the lack of else block)
+                if (isOrigin) {
```
halkon: Just do the check here, no need to use a variable

halkon:
LGTM

src/components/rxApp/templates/rxAppNavItem.html
```diff
@@ -1,6 +1,6 @@
 <li class="rx-app-nav-item" ng-show="isVisible(item.visibility, item.roles)"
 ng-class="{'has-children': item.children.length > 0, active: item.active, 'rx-app-key-{{ item.key }}': item.key }">
-    <a ng-href="{{ getUrl(item.url) }}" ng-attr-target="{{ getTarget() }}" class="item-link" ng-click="toggleNav($event, item.href)">{{item.linkText}}</a>
+    <a ng-href="{{ getUrl(item.url) }}" ng-attr-target="{{ getTarget() }}" class="item-link" ng-click="toggleNav($event, item.href); navigateToApp($event, getUrl(item.url))">{{item.linkText}}</a>
```
parlarjb: Rather than 
```
ng-click="toggleNav($event, item.href); navigateToApp($event, getUrl(item.url))"
```

could you do something like 
```
ng-click="toggleNav($event, item.href); navigateToApp($event, getUrl(item.url))"
```

could you just call `navigateToApp()` inside of the `toggleNav()`?

I don't like putting two functions calls like this in a template, because it means you're encoding business logic into the template itself. An `ng-click` should signify the template saying "I need a thing to happen when this is clicked, I don't care how it's done, the function will figure it out". But here you're asking for two separate things to happen

halkon: rather than pass `item.href` why don't we just pass `item` and then inside of the `toggleNav` call we just call `getUrl(item.url)` since the href and url are different.

lentusignavus: I do like the idea of having a `navClickHandler($event, item)` where both functions get called separately

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -181,6 +181,25 @@ angular.module('encore.ui.rxApp')
...
+            $scope.navigateToApp = function (ev, url) {
+                // We want to control what the click to the <a> tag does
+                // If it is Origin prevent the default click action
+                // otherwise handle the click as normal (implied by the lack of else block)
+                if ($injector.has('oriLocationService')) {
+                    ev.preventDefault();
+                    var currentIframeUrl = $injector.get('oriLocationService').getCanvasURL();
+                    // Only change the iFrame if the urls are different
+                    if (typeof url !== 'undefined' && currentIframeUrl !== url) {
+                        $injector.get('oriLocationService').setCanvasURL(url);
```
parlarjb: Rather than doing `$injector.get('oriLocationService')` a second time (you also do it on line 191), can you just do it one time above line 191, store the service into a variable, and reuse it?

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -181,6 +181,25 @@ angular.module('encore.ui.rxApp')
...
+                    }
+                }
+            };
+
+            $scope.navClickHandler = function (clickEvent, item) {
+                $scope.toggleNav(clickEvent, item.href);
+                $scope.navigateToApp(clickEvent, $scope.getUrl(item.url));
```
halkon: can you just pass `item.url` here? and call `$scope.getUrl` in line 194?

src/components/rxApp/scripts/rxAppNavItem.js
```diff
@@ -181,6 +181,28 @@ angular.module('encore.ui.rxApp')
...
+                // If it is Origin prevent the default click action
+                // otherwise handle the click as normal (implied by the lack of else block)
+                if ($injector.has('oriLocationService')) {
+                    var oriLocationService = $injector.get('oriLocationService');
+                    var currentIframeUrl = oriLocationService.getCanvasURL();
+                    var finalUrl = $scope.getUrl(url);
+
+                    ev.preventDefault();
+                    // Only change the iFrame if the urls are different
+                    if (typeof url !== 'undefined' && currentIframeUrl !== finalUrl) {
```
jonkruse00: @lentusignavus @halkon Is `typeof url !== 'undefined'` necessary ? Or can `url` be another type of falsie value and thats why you make sure that its the `undefined` type ?

jonkruse00:
If Im allowed to LGTM I would.

halkon:
@jonkruse00 can you write javascript? can you write code? 😛

src/components/rxApp/scripts/rxAppNavItem.js
halkon: actually there are arguments online about using `typeof url !...` as a check for undefined, that being said, @jonkruse00 is right.  

Perhaps it's better to move the `finalUrl` logic inside of this if, and just do: `if (!_.isEmpty...` 

@CITguy thoughts?

jonkruse00:
@lentusignavus Can you do the dew ?

halkon:
LGTM

src/components/rxApp/scripts/rxAppNavItem.spec.js
```diff
@@ -295,5 +296,49 @@ describe('encore.ui.rxApp', function () {
...
+            });
+
+            afterEach(function () {
+                navigateToAppSpy.restore();
+                setCanvasURLSpy.restore();
+            });
+
+            it('should call navigateToApp on click', function () {
+                el.find('.item-link').click();
+                sinon.assert.called(navigateToAppSpy);
```
CITguy: Please use the `expect()` syntax for consistency. Same goes for the other assertions below.

lentusignavus: roger roger

src/components/rxApp/scripts/rxAppNavItem.spec.js
```diff
@@ -295,5 +296,49 @@ describe('encore.ui.rxApp', function () {
...
+            });
+
+            afterEach(function () {
+                navigateToAppSpy.restore();
+                setCanvasURLSpy.restore();
+            });
+
+            it('should call navigateToApp on click', function () {
+                el.find('.item-link').click();
+                expect(navigateToAppSpy.called).to.be.true;
```
CITguy: Because we use [`sinon-chai`](https://github.com/domenic/sinon-chai), you can actually use `expect(navigateToAppSpy).to.have.been.called`

CITguy:
Dev LGTM

CITguy:
@lentusignavus: please squash this PR

PR #1780, opened 2016-08-24
FRMW-455: Add Copy Element (rxCopy)

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-455

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM
- [x] QE LGTM

### DIFF
--- a/.eslintrc.test.json
+++ b/.eslintrc.test.json
@@ -8,6 +8,7 @@
         "expect": true,
         "it": false,
         "inject": false,
+        "process": true,
         "screenshot": true,
         "sinon": false
     },
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   - "5.1.1" # same as .node-version
 addons:
- - firefox: "41.0"
+  firefox: "41.0"
 before_install:
 - node ./utils/quitIfFork.js
 - npm install -g grunt-cli bower
--- a/src/elements/Copy/scripts/rxCopy.js
+++ b/src/elements/Copy/scripts/rxCopy.js
@@ -1,0 +1,80 @@
+angular.module('encore.ui.elements')
+/**
+ * @name elements.directive:rxCopy
+ * @ngdoc directive
+ * @restrict E
+ * @scope
+ *
+ * @requires $interpolate
+ * @requires $interval
+ * @requires $window
+ * @requires utilities.service:rxCopyUtil
+ *
+ * @description
+ * The rxCopy directive is designed to provide programmatic copy-to-clipboard
+ * functionality for plain text on the screen.
+ *
+ * @example
+ * <pre>
+ * <rx-copy>
+ *   This text will be copied to the system clipboard when you click the
+ *   clipboard icon found nearby on the page.
+ * </rx-copy>
+ * </pre>
+ */
+.directive('rxCopy', function ($window, $interval, $interpolate, rxCopyUtil) {
+    return {
+        restrict: 'E',
+        transclude: true,
+        scope: {},
+        templateUrl: 'templates/rxCopy.html',
+        link: function (scope, element, attrs, controllers, transclude) {
+            // first span in the template (aka. '.rxCopy__text')
+            var copyTextNode = element.find('span')[0];
+
+            var CopyState = {
+                waiting: function () {
+                    scope.copyState = 'waiting';
+                    scope.tooltip = 'Click to Copy';
+                },
+
+                passed: function () {
+                    scope.copyState = 'success';
+                    scope.tooltip = 'Copied!';
+                },
+
+                failed: function () {
+                    scope.copyState = 'fail';
+
+                    if ($window.navigator.platform.match(/mac/i)) {
+                        scope.tooltip = 'Press &#x2318;-C to copy.';
+                    } else {
+                        scope.tooltip = 'Press Ctrl-C to copy.';
+                    }
+                }//failed
+            };//CopyState
+
+            CopyState.waiting();
+
+            // We use transclude so that we can trim any leading/trailing
+            // whitespace from the wrapped text.  This is very difficult,
+            // using the selection API alone.
+            transclude(function (clone) {
+                var cloneText = clone.text().trim();
+                // $interpolate to ensure that angular expressions are evaluated
+                // before setting the trimmed content
+                scope.trimmedContent = $interpolate(cloneText)(scope.$parent);
+            });
+
+            scope.copyText = function () {
+                // document.execCommand() requires a selection
+                // before it can perform a copy operation
+                rxCopyUtil.selectNodeText(copyTextNode);
+
+                rxCopyUtil.execCopy(CopyState.passed, CopyState.failed);
+
+                $interval(CopyState.waiting, 3000, 1);
+            };//copyText()
+        }
+    };
+});//rxCopy
--- a/utils/rx-page-objects/src/tooltip.page.js
+++ b/utils/rx-page-objects/src/tooltip.page.js
@@ -1,0 +1,69 @@
+'use strict';
+
+/* CSS HIERARCHY
+ * ----------------------------------------
+ * div.tooltip <-- tooltipElement for constructor
+ *   div.tooltip-arrow
+ *   div.tooltip-inner
+ */
+
+/**
+ * @class
+ */
+class Tooltip {
+    /**
+     * @param {ElementFinder} tooltipElement
+     * ElementFinder to be transformed into a Tooltip page object
+     */
+    constructor (tooltipElement) {
+        this.rootElement = tooltipElement;
+
+        // Private selectors
+        this.eleInnerText = this.rootElement.$('.tooltip-inner');
+    }//constructor
+
+    /**
+     * @description
+     * (READ-ONLY) Text value of tooltip (if present).
+     *
+     * Warning: This property is known to be unstable in many
+     * Selenium end to end test runs in EncoreUI. Returns the tooltip text.
+     * Will automatically hover over the tooltip for you to retrieve the text.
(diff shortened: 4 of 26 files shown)


CITguy:
waiting to see what CI services do with midways, before continuing

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   - "5.1.1" # same as .node-version
 addons:
- - firefox: "41.0"
+  firefox: "41.0"
```
CITguy: Because... https://docs.travis-ci.com/user/firefox/

CITguy:
I'll be debugging the issue with Travis, but this PR is ready to review.

-----

encoreui-elements-copy-full-page

encoreui-elements-rxcopy-ngdocs

encoreui-elements-rxcopyutil-ngdocs

## Full vs Degraded Behavior

### Full Functionality
Chrome 42+ and Firefox 41+ should provide full functionality. rxCopy will behave as follows.

* ✅  Text will be selected
* ✅  Text will be copied to clipboard
* ✅  Tooltip will display "Copied!"

encoreui-elements-copy-chrome-full-functionality

### Degraded Functionality
Older browsers, or even current browsers that are not Chrome or Firefox, may have degraded functionality. rxCopy will behave as follows.

* ✅  Text will be selected
* ❌  Text will not be copied to clipboard
* ✅  Tooltip will display "Press Ctrl-C to Copy" or "Press ⌘-C to Copy"

encoreui-elements-copy-ff40-degraded-functionality

src/elements/Copy/scripts/rxCopy.js
```diff
@@ -0,0 +1,80 @@
...
+ * </rx-copy>
+ * </pre>
+ */
+.directive('rxCopy', function ($window, $interval, $interpolate, rxCopyUtil) {
+    return {
+        restrict: 'E',
+        transclude: true,
+        scope: {},
+        templateUrl: 'templates/rxCopy.html',
+        controller: function ($scope, $element, $attrs, $transclude) {
```
parlarjb: Why a `controller` instead of `link`?

utils/rx-page-objects/src/tooltip.page.js
```diff
@@ -0,0 +1,69 @@
...
+     * If there is no tooltip present on hover, returns `null`.
+     *
+     * @example
+     * it('should have the correct tooltip text', function () {
+     *     expect(myTooltip.getText()).to.eventually.equal('Some Value');
+     * });
+     *
+     * @type {Promise<String|null>}
+     */
+    get text () {
```
Droogans: Your documentation example is correct, but this name is not!

I think this should be a function, `getText`.

utils/rx-page-objects/test/rxCopy.midway.js
```diff
@@ -0,0 +1,218 @@
+// Until we upgrade to protractor 4, we have to skip these tests
+// for firefox to allow Travis to pass checks.
+browser.getCapabilities().then((capabilities) => {
+    if (capabilities.get('browserName') === 'firefox') {
+        // bail before testing
+        return;
```
Droogans: Pretty sure this is in another test.

Think it's time to make a global var for the current browser under test to avoid unnecessary promise callback indentation? I'd love it if I could reach up and grab the current browser name without any promise manipulation required.

CITguy: The only other test I found this type of logic was in `rxNotify.midway.js`, but it was wrapping individual test logic instead of the entire suite.  I agree that we should make a global at some point, though.  I tried using mocha 3, to try and make use of `this.skip()`, but there was strange behavior happening that I'd rather not sink a lot of time into right now.  As such, I opted to use the more straightforward (though ugly) solution you see here.

utils/rx-page-objects/src/tooltip.page.js
CITguy: Yeah, good call. Getters/setters should be reserved for simple values, not promises. I'll update.

CITguy:
Note to self: Update VR section to be responsive.

src/elements/Copy/scripts/rxCopy.js
CITguy: Can't remember. Let me try `link` again.

utils/rx-page-objects/test/rxCopy.midway.js
```diff
@@ -0,0 +1,218 @@
+// Until we upgrade to protractor 4, we have to skip these tests
+// for firefox to allow Travis to pass checks.
+browser.getCapabilities().then((capabilities) => {
```
CITguy: Note to Self: this is preventing all of the tests from running regardless of CI service. Find fix.

utils/rx-page-objects/test/rxStatusColumn.midway.js
```diff
@@ -44,7 +44,7 @@ var tablePageObject = Page.create({
                 // goo.gl/OJdysF
                 title: {
                     get: function () {
-                        return rowElement.$('td+td').getText();
+                        return rowElement.$('td+td').text;
```
CITguy: Note to Self: Revert changes to this file.

.eslintrc.test.json
```diff
@@ -8,6 +8,7 @@
         "expect": true,
         "it": false,
         "inject": false,
+        "process": true,
```
CITguy: Required for midway tests to determine which Protractor Key Chord to use when pasting into TextArea.

utils/rx-page-objects/test/rxCopy.midway.js
Droogans: I'm pretty sure this will work just fine if this becomes a promise + global var assignment in the protractor conf's `onPrepare` block.

However, you will need to add it to both `protractor.conf.js` and `protractor.chrome.conf.js`.

ElementE:
Design LGTM
-nice work sir

utils/rx-page-objects/test/rxCopy.midway.js
CITguy: I'm switching to use `if (!process.env.TRAVIS) { ... }` instead. I cannot find an elegant solution purely in protractor.

utils/rx-page-objects/test/rxCopy.midway.js
```diff
@@ -0,0 +1,212 @@
+// Until we upgrade to protractor 4 (to get selenium 2.53+),
+// we'll skip these tests on Travis to pass checks.
+if (!process.env.TRAVIS) {
```
CITguy: @Droogans: The previous way, using the `browser.getCapabilities()` method, didn't provide the ability to skip the entire file. As a result, I ended up checking for the `TRAVIS` environment variable, at Node runtime, before any protractor logic executes.

CITguy: This works as expected.

## Travis
*Skips rxCopy*
screen shot 2016-08-26 at 3 41 44 pm

## Codeship
*Does not skip rxCopy*
screen shot 2016-08-26 at 3 41 29 pm

utils/rx-page-objects/test/rxCopy.midway.js
```diff
@@ -0,0 +1,212 @@
...
+                expect(subject.isWaiting()).to.eventually.be.true;
+            });
+
+            it('should not be successful', function () {
+                expect(subject.isSuccessful()).to.eventually.be.false;
+            });
+
+            it('should not be failed', function () {
+                expect(subject.isFailed()).to.eventually.be.false;
+            });
```
Droogans: This test, particularly the test name, highlights some awkward grammar used in the API. However, I have no immediate suggestions as to what would work well in this case.

@jj4th got any ideas?

CITguy: Yeah, I'm not attached to the name either, but it was the best I could come up with at the time. I basically need to test that it's in state A, and not states B and C.

CITguy: We don't have any existing cases in the tests for elements with more than two states. So, I did the best I could.

Droogans: I've thought about this. Nobody should be manually interacting with success/failure states in application-specific end to end tests. If they want to, they can use an exercise (which I'm about to start writing).

Otherwise, they might, at most, copy something from the input. Maybe. Every other method exists solely to facilitate exercises, so `isFailed` isn't that big of a deal to me.

I can try to fix it with the exercises, but if not I'll mark the method private in jsdoc to keep it from cluttering the documentation.

halkon:
LGTM

utils/rx-page-objects/test/rxCopy.midway.js
```diff
@@ -0,0 +1,212 @@
...
+
+                    it('should not be successful', function () {
+                        expect(subject.isSuccessful()).to.eventually.be.false;
+                    });
+
+                    it('should not be failed', function () {
+                        expect(subject.isFailed()).to.eventually.be.false;
+                    });
+
+                    it ('should have default tooltip', function () {
```
parlarjb: extra space between `it` and `(`

parlarjb:
One small typo above, otherwise LGTM

utils/rx-page-objects/test/rxCopy.midway.js
Droogans: There's an exercises PR inbound for this branch, I can make sure this gets cleaned up.

100stacks: Adding reference to `rxCopy` exercises PR https://github.com/rackerlabs/encore-ui/pull/1791

CITguy:
QE LGTM - By proxy of #1791 being written by @Droogans himself and by passing CI checks

Issue #1767, opened 2016-08-11
rxAge filter doesn't handle very small time differences properly.


jj4th:
**Observed with:**
`Encore UI: 1.45.0` + `Moment: 2.5.1` AND
`Encore-UI: master` + `Moment: 1.13.0`

**Steps to recreate:**
- Invoke rxAge filter with `new Date();` or a date within a few milliseconds of the current date.

**Result:**
rxAge returns an improper date, typically `1 minute, -22 hours`

This happens because of the behavior of the line: https://github.com/rackerlabs/encore-ui/blob/master/src/utilities/rxAge/scripts/rxAge.js#L76

```javascript
var days = parseInt(duration.asDays(), 10);
```

Here is the example node console output using very small duration values with the current version of moment to demonstrate the behavior that causes this error:
![screen shot 2016-08-11 at 6 05 34 pm](https://cloud.githubusercontent.com/assets/1107895/17607977/fba83a22-5fee-11e6-8ca9-6ede1b116bb1.png)

jj4th:
It is worth noting that `moment.duration` generates a float, not a string, so `parseInt` is probably wrong here.

CITguy:
The issue is that `moment.duration(...).asDays()`, `.asHours()`, and `.asMinutes()` are returning an exponent for small values and `parseInt()` is interpreting the value incorrectly.

So consider the following:

```javascript
var d = moment.duration(1);
d.asDays(); // 1.1574074074074074e-8
parseInt(d.asDays()); // 1 (incorrect)
```

CITguy:
Captured in JIRA: https://jira.rax.io/browse/FRMW-1000

PR #1787, opened 2016-08-26
fix(rxAge): Fixes parseInt bug

lentusignavus:
JIRA: https://jira.rax.io/browse/FRMW-1000

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

With really small durations moment returns a small float
parseInt handles this weirdly returning 1 and sometimes 2
Changes implementation to use Math.floor

Issue: https://github.com/rackerlabs/encore-ui/issues/1767

PR #1788, opened 2016-08-26
fix(rxAge): Fixes parseInt bug

lentusignavus:
JIRA: https://jira.rax.io/browse/FRMW-1000

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

With really small durations moment returns a small float
parseInt handles this weirdly returning 1 and sometimes 2
Changes implementation to use Math.floor

Issue: https://github.com/rackerlabs/encore-ui/issues/1767

### DIFF
--- a/src/utilities/rxAge/scripts/rxAge.js
+++ b/src/utilities/rxAge/scripts/rxAge.js
@@ -73,9 +73,9 @@ angular.module('encore.ui.utilities')
         var date = moment(new Date(dateString));
         var diff = now.diff(date);
         var duration = moment.duration(diff);
-        var days = parseInt(duration.asDays(), 10);
-        var hours = parseInt(duration.asHours(), 10);
-        var mins = parseInt(duration.asMinutes(), 10);
+        var days = Math.floor(duration.asDays());
+        var hours = Math.floor(duration.asHours());
+        var mins = Math.floor(duration.asMinutes());
         var age = [];
 
         if (_.isBoolean(maxUnits)) {


src/utilities/rxAge/scripts/rxAge.js
```diff
@@ -73,9 +73,9 @@ angular.module('encore.ui.utilities')
         var date = moment(new Date(dateString));
         var diff = now.diff(date);
         var duration = moment.duration(diff);
-        var days = parseInt(duration.asDays(), 10);
-        var hours = parseInt(duration.asHours(), 10);
-        var mins = parseInt(duration.asMinutes(), 10);
+        var days = Math.floor(duration.asDays(), 10);
```
lentusignavus: Removing the 10, next commit

jonkruse00: just saying from MDN : 
<blockquote>
<b>Always specify this parameter</b> to eliminate reader confusion and to guarantee predictable behavior. Different implementations produce different results when a radix is not specified, usually defaulting the value to 10.
</blockquote>

jonkruse00:
If I can LGTM I do

Droogans:
The commit message "Update rxAge.js" is not very informative.

Can you change it to be more explanatory? Also, following the commit message format guidelines is always a plus.

CITguy:
LGTM

CITguy:
Marking "DO NOT MERGE" until #1787 is ready for merge.

PR #1791, opened 2016-08-29
Frmw 455 rx copy exercises

Droogans:
JIRA: https://jira.rax.io/browse/FRMW-455

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM

This can be closed and cherry picked if you want. Or merged.

### DIFF
--- a/utils/rx-page-objects/src/rxCopy.exercise.js
+++ b/utils/rx-page-objects/src/rxCopy.exercise.js
@@ -78,7 +78,7 @@ exports.rxCopy = function (options) {
                 expect(component.isSuccessful()).to.eventually.be.false;
             });
 
-            it('should not be failed', function () {
+            it('should not have failed', function () {
                 expect(component.isFailure()).to.eventually.be.false;
             });
 
@@ -122,7 +122,7 @@ exports.rxCopy = function (options) {
                     expect(component.isSuccessful()).to.eventually.be.true;
                 });
 
-                it('should not be failed', function () {
+                it('should not have failed', function () {
                     expect(component.isFailure()).to.eventually.be.false;
                 });
 
@@ -143,7 +143,7 @@ exports.rxCopy = function (options) {
                         expect(component.isSuccessful()).to.eventually.be.false;
                     });
 
-                    it('should not be failed', function () {
+                    it('should not have failed', function () {
                         expect(component.isFailure()).to.eventually.be.false;
                     });
 
--- a/utils/rx-page-objects/src/rxCopy.exercise.js
+++ b/utils/rx-page-objects/src/rxCopy.exercise.js
@@ -1,0 +1,149 @@
+var _ = require('lodash');
+
+/**
+ * @description rxCopy exercises.
+ * @see rxCopy
+ * @exports exercise/rxCopy
+ * @param {Object} options - Test options. Used to build valid tests.
+ * @param {rxSelect} options.instance - Component to exercise.
+ * @param {Boolean} [options.isPresent=true] - Determines if the component is present.
+ * @param {Boolean} [options.isDisabled=false] - Determines if the component is disabled.
+ * @param {Boolean} [options.isDisplayed=true] - Determines if the component is displayed.
+ * @param {ElementFinder} [testCopyArea=] - An input element somewhere that rxCopy contents can be pasted to.
+ * @param {String} [expectedText=] - The expected text content inside the input element to be copied.
+ *
+ */
+exports.rxCopy = function (options) {
+    if (options === undefined) {
+        options = {};
+    }
+
+    options = _.defaults(options, {
+        isPresent: true,
+        isEnabled: true,
+        isDisplayed: true
+    });
+
+    return function () {
+        var component;
+
+        /**
+         * @returns {Promise<String>}
+         */
+        function getPastedValue () {
+            return options.testCopyArea
+                .clear()
+                .sendKeys(protractor.Key.chord(
+                    (process.platform === 'darwin' ? protractor.Key.META : protractor.Key.CONTROL),
+                    'v'
+                ))
+                .getAttribute('value');
+        }
+
+        before(function () {
+            component = options.instance;
+        });
+
+        it('should be present', function () {
+            expect(component.isPresent()).to.eventually.eq(options.isPresent);
+        });
+
+        if (!options.isPresent) {
+            return;
+        }
+
+        it(`should ${options.isDisplayed ? 'be' : 'not be'} displayed`, function () {
+            expect(component.isDisplayed()).to.eventually.eq(options.isDisplayed);
+        });
+
+        if (!options.isDisplayed) {
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


utils/rx-page-objects/src/rxCopy.page.js
```diff
@@ -100,6 +125,9 @@ class rxCopy {
      */
     _hoverOverAction () {
         browser.actions().mouseMove(this.eleAction).perform();
+        // I know what you're thinking -- don't. Just leave it.
+        // Otherwise, tooltips in tables in Chrome will not actually appear.
+        browser.actions().mouseMove(this.eleAction).perform();
```
CITguy: Dear Selenium...

wtf

CITguy:
LGTM

utils/rx-page-objects/src/rxCopy.exercise.js
```diff
@@ -0,0 +1,157 @@
...
+                .clear()
+                .sendKeys(protractor.Key.chord(
+                    (process.platform === 'darwin' ? protractor.Key.META : protractor.Key.CONTROL),
+                    'v'
+                ))
+                .getAttribute('value');
+        }
+
+        before(function () {
+            component = options.instance;
```
lentusignavus: Should we throw an error if an instance is not provided? Or is this `exercise` only meant to be consumed in these files?

I imagine a default value would be hard to come up with

utils/rx-page-objects/src/rxCopy.exercise.js
```diff
@@ -0,0 +1,157 @@
...
+
+            it('should be waiting', function () {
+                expect(component.isWaiting()).to.eventually.be.true;
+            });
+
+            it('should not be successful', function () {
+                expect(component.isSuccessful()).to.eventually.be.false;
+            });
+
+            it('should not be failed', function () {
```
100stacks: @Droogans, I think this `should not fail` 💭

utils/rx-page-objects/src/rxCopy.exercise.js
```diff
@@ -0,0 +1,157 @@
...
+
+                    it('should be waiting', function () {
+                        expect(component.isWaiting()).to.eventually.be.true;
+                    });
+
+                    it('should not be successful', function () {
+                        expect(component.isSuccessful()).to.eventually.be.false;
+                    });
+
+                    it('should not be failed', function () {
```
100stacks: I think this `should not fail` 💭

utils/rx-page-objects/src/rxCopy.exercise.js
```diff
@@ -0,0 +1,157 @@
...
+ * @param {rxSelect} options.instance - Component to exercise.
+ * @param {Boolean} [options.isPresent=true] - Determines if the component is present.
+ * @param {Boolean} [options.isDisabled=false] - Determines if the component is disabled.
+ * @param {Boolean} [options.isDisplayed=true] - Determines if the component is displayed.
+ * @param {ElementFinder} [testCopyArea=] - An input element somewhere that rxCopy contents can be pasted to.
+ * @param {String|Regex} [expectedText=] - The expected text content inside the input element to be copied.
+ * Can be a regular expression if the text content is very large.
+ */
+exports.rxCopy = function (options) {
+    if (options === undefined) {
```
lentusignavus: Related to comment below. If it's somewhat expected to get an undefined value for options then `options.instance` should have a default value or inform the dev somehow

**This comment was written after the one below**

utils/rx-page-objects/src/rxCopy.exercise.js
CITguy: No exercise can work without an element. It'll be painfully obvious if someone attempts to use exercises without an instance. IMO, I don't think it's worth the effort to put in error handling for exercises, because the consumers are developers, not end users.

utils/rx-page-objects/src/rxCopy.exercise.js
CITguy: `fail` is an action, while `failed` is a state. This test is checking the state of the element.

100stacks:
LGTM

utils/rx-page-objects/src/rxCopy.exercise.js
CITguy: see previous comment

utils/rx-page-objects/src/rxCopy.exercise.js
Droogans: This is already a well covered topic by myself and @CITguy. Right now, we just have some guidelines.

- If you have a component where there is (by design's input/requirements) only supposed to be one on the page at a time (breadcrumbs is a good example), then it is ok to provide default constructor parameters for `new rxBreadcrumbs()`. If you don't pass in anything, it will use the "one" breadcrumb you're supposed to have on the page. These types of components can default to using the "one" instance in the exercises, too! So a breadcrumbs exercise wouldn't need an `options.instance` argument to be required.

- If you have something like `rxPaginate`, where it's very likely that you could have two or more instances on a page, you must always pass in the root element to the constructor (`new rxPaginate($('#my-table .pagination'));`, or whatever). In these cases, the exercise requires the user to pass in an `options.instance` to the exercise.

In this case, rxCopy usage allows for multiple instances on the page, so we require that you tell us which rxCopy to test, everytime. I could assume that you're asking about the *first* rxCopy that I can find on the page, but we decided that it's better to guide the user towards proper usage, even if it could be more convenient for us not to.

utils/rx-page-objects/src/rxCopy.exercise.js
CITguy: Some exercises can do this, because there's naturally going to be one of an element on the page (rxBreadcrumbs, rxApp, etc.). The rxCopy element cannot default, because there's likely going to be more than one on a page and there's no way to reliably determine the correct element by default.

Droogans: The jsdoc comments here tell you two things that are really important: `options` is a required parameter (because the name of the argument is not wrapped in `[]`), and that `options.instance` is also required. If you don't obey the call signature that's documented, you're going to have a bad time.

Once we switch to typescript, this is all going to get much, much simpler. We could even theoretically kill off much of our documentation, but I wouldn't recommend it.

<sub>In an unrelated note -- hey it looks like git's syntax awareness has spread to jsdoc blocks! Required parameters are colored, optional ones are greyed out!</sub>

utils/rx-page-objects/src/rxCopy.exercise.js
lentusignavus: Purrfect, thanks for the explanation! 😄

utils/rx-page-objects/src/rxCopy.exercise.js
100stacks: @CITguy, sounds good.  The text could be clearer that it is testing the element's state rather than action.

PR #1799, opened 2016-09-06
3.x upgrade to lodash 4

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-445

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

Update lodash to 4.x

### DIFF
--- a/src/elements/Notifications/docs/examples/notifications.stack.js
+++ b/src/elements/Notifications/docs/examples/notifications.stack.js
@@ -22,7 +22,7 @@ angular.module('demoApp')
         var messageOptions = _.clone($scope.options);
 
         if ($scope.ondismiss.should) {
-            messageOptions.ondismiss = _.clone($scope.ondismiss.method);
+            messageOptions.ondismiss = $scope.ondismiss.method;
         }
 
         messageOptions.stack = stack;
--- a/src/elements/Spinner/scripts/rxSpinner.spec.js
+++ b/src/elements/Spinner/scripts/rxSpinner.spec.js
@@ -1,7 +1,5 @@
 describe('rxSpinner', function () {
-    var scope, compile, el;
-    var validTemplate = '<div rx-spinner toggle="toggle"></div>';
-    var darkTemplate = '<div rx-spinner="dark" toggle="toggle"></div>';
+    var scope, compile, el, template;
 
     beforeEach(function () {
         // load module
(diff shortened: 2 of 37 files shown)


src/elements/Notifications/docs/examples/notifications.stack.js
```diff
@@ -22,7 +22,7 @@ angular.module('demoApp')
         var messageOptions = _.clone($scope.options);
 
         if ($scope.ondismiss.should) {
-            messageOptions.ondismiss = _.clone($scope.ondismiss.method);
+            messageOptions.ondismiss = $scope.ondismiss.method;
```
CITguy: **Be forewarned!** This one threw me for a loop.

You can see how it differs between lodash 2, 3, and 4 [here](http://jsbin.com/xewavex/8/edit?html,js,console).

src/elements/Spinner/scripts/rxSpinner.spec.js
```diff
@@ -14,43 +12,64 @@ describe('rxSpinner', function () {
...
+            beforeEach(function () {
+                scope.toggle = false;
+                scope.$digest();
+            });
+
+            it('should not render a spinner', function () {
+                expect(el.children()).to.have.lengthOf(0);
+            });
+        });//not toggled
+    });//dark spinner
```
CITguy: This would've been fine with the fix to `helpers.getChildDiv`, but I fixed the tests to remove the dependency on that function anyway.

src/utilities/rxNotify/scripts/rxNotify.js
```diff
@@ -383,11 +383,11 @@ angular.module('encore.ui.utilities')
             stacks[stack] = [];
         }
 
-        // merge options with defaults (overwriting defaults where applicable)
+        // add defaults to options
         _.defaults(options, messageDefaults);
 
         // add options to message
-        _.merge(message, options);
```
CITguy: `_.merge` is recursive, but it didn't seem to be working in 4.x and `_.defaults` does what we need anyway

halkon:
LGTM

src/utilities/rxNotify/scripts/rxNotify.js
CITguy: I'm unable to pinpoint the exact issue with `_.merge`, but results seem to suggest that the scope on which a `$watch` is applied for dismissal is not the one being passed to `rxNotify.add()` (see below)

screen shot 2016-09-07 at 2 08 42 am

src/utilities/rxNotify/scripts/rxNotify.js
```diff
@@ -163,7 +163,7 @@ angular.module('encore.ui.utilities')
         loading: false,
         show: 'immediate',
         dismiss: 'next',
-        ondismiss: _.noop(),
```
CITguy: I'm surprised this was working, to begin with. We don't want the result of the noop. We want the function itself.

lentusignavus:
@CITguy We can do the vidyo review

PR #1805, opened 2016-09-08
chore(navItemTarget): Use null instead of _self as default target

halkon:
JIRA:

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM


lentusignavus:
@halkon Why not self?

halkon:
@lentusignavus from reading:
http://stackoverflow.com/questions/16837704/angularjs-normal-links-with-html5mode
and
It turns out that angular will omit taking over the html5 links when there's a `target="_self"` present.  

Interestingly enough this behavior was not exposed on other apps, but testing this change within ticketing proved to be the actual fix.

lentusignavus:
DEV LGTM

parlarjb:
DEV LGTM

PR #1822, opened 2016-09-23
chore(iframe): try/catch access to the top frame (1.x)

halkon:
JIRA:

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

### DIFF
--- a/src/components/rxApp/scripts/rxApp.js
+++ b/src/components/rxApp/scripts/rxApp.js
@@ -50,7 +50,7 @@ angular.module('encore.ui.rxApp')
                 // Element by attempting to set a property on it.  If we have no errors
                 // then this means that `document.domain` matches and we have no Cross
                 // Origin security constraints
-                $window.top['encore'] = true;
+                $window.top['hasSameDomain'] = true;
                 scope.isEmbedded = $window.self !== $window.top;
             } catch (e) {
                 scope.isEmbedded = false;
--- a/src/components/rxApp/scripts/rxApp.js
+++ b/src/components/rxApp/scripts/rxApp.js
@@ -44,7 +44,17 @@ angular.module('encore.ui.rxApp')
 
             scope.isWarning = scope.isPreProd || scope.isLocalNav;
 
-            scope.isEmbedded = $window.self !== $window.top;
+            scope.isEmbedded = false;
+            try {
+                // Checks to see if we have access to the global scope of a DOM Window
+                // Element by attempting to set a property on it.  If we have no errors
+                // then this means that `document.domain` matches and we have no Cross
+                // Origin security constraints
+                $window.top['encore'] = true;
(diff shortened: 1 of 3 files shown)


src/components/rxApp/scripts/rxApp.js
```diff
@@ -44,7 +44,17 @@ angular.module('encore.ui.rxApp')
...
             scope.isWarning = scope.isPreProd || scope.isLocalNav;
 
-            scope.isEmbedded = $window.self !== $window.top;
+            scope.isEmbedded = false;
+            try {
+                // Checks to see if we have access to the global scope of a DOM Window
+                // Element by attempting to set a property on it.  If we have no errors
+                // then this means that `document.domain` matches and we have no Cross
+                // Origin security constraints
+                $window.top['encore'] = true;
```
Droogans: Why not make this a little more explicit?

```js
$window.top['hasSameDomain'] = true
```

halkon: @Droogans done

src/utilities/rxLocalStorage/scripts/rxLocalStorage.js
```diff
@@ -22,8 +22,12 @@ angular.module('encore.ui.utilities')
...
 .service('rxLocalStorage', function ($window) {
     var localStorage = $window.localStorage;
-    if ($window.self !== $window.top && $window.top.localStorage) {
-        localStorage = $window.top.localStorage;
+    if ($window.self !== $window.top) {
+        try {
+            localStorage = $window.top.localStorage;
+        } catch (e) {
+            localStorage = $window.localStorage;
+        }
```
Droogans: Can you explain what happened here? https://github.com/rackerlabs/encore-ui/commit/ca66b158bb190660d3229b9e2dd9071ddb372733#diff-1046d29f8cc250327e7dfa57de04677fR23 shows that the previous version was written by you, and I liked the previous implementation better. What error was being thrown? How did you find out about it?

halkon: If the frame does not have access to the parent frame, both Chrome and Firefox will throw an exception that prevents the rest of the application to run.  By simply referring `$window.top.localStorage` you're requesting access and thus you get a security violation when both frames do not have the same domain

halkon: that link points to the factory definition. 

Are you asking strictly why `$window.top.localStorage` was in the conditional statement? If so that was due to PhantomJS & Karma.  The way the unit tests are run are through an `iframe`, and PhantomJS allows a child frame to retrieve it's parents `localStorage` (or properties) without issuing a security error, but instead of being an actual object, it's undefined.

lentusignavus:
LGTM

PR #1839, opened 2016-10-07
fix(rxAccountUsers): Fix directive in Origin's nav

parlarjb:
JIRA:

### LGTMs
- [x] Dev LGTM
- [x] QE LGTM

This fixes our usage of rxAccountUsers when oriLocationService
is available


halkon:
LGTM

Droogans:
I'm so glad there is no more shared nav, otherwise this would have to go back to 1.x too.

PR #1845, opened 2016-10-17
fix(Forms): Styling Issue with "Saving Form Data" example

kevincareer427:
JIRA: https://jira.rax.io/browse/FRMW-1011

### LGTMs
- [x] Dev LGTM

### DIFF
--- a/src/elements/Forms/docs/examples/forms.autoSave.html
+++ b/src/elements/Forms/docs/examples/forms.autoSave.html
@@ -3,6 +3,14 @@
   away from this page, and then return, whatever you previously typed will
   automatically be populated into the text input
 </p>
-<span class="field-input" ng-controller="formsAutoSaveExampleController">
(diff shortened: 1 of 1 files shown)


kevincareer427:
Before:

before

after:

after

src/elements/Forms/docs/examples/forms.autoSave.html
```diff
@@ -3,6 +3,14 @@
   away from this page, and then return, whatever you previously typed will
   automatically be populated into the text input
 </p>
-<span class="field-input" ng-controller="formsAutoSaveExampleController">
-  <input type="text" ng-model="forms.autosave">
-</span>
+<form ng-controller="formsAutoSaveExampleController" rx-form>
+  <rx-form-section>
```
CITguy: Please add the `controlled-width` attribute to the `<rx-form-section>` to limit the width of the input.

PR #1846, opened 2016-10-19
fix(rxAutoSave): Update rxAutoSave to use $location.path()

kevincareer427:
JIRA: https://jira.rax.io/browse/FRMW-819
### LGTMs
- [x] Dev LGTM

### DIFF
--- a/utils/rx-page-objects/test/rxLocalStorage.midway.js
+++ b/utils/rx-page-objects/test/rxLocalStorage.midway.js
@@ -19,7 +19,7 @@ describe('utilities:rxLocalStorage', function () {
         expect(rxLocalStorage.getItem('anInt')).to.eventually.equal(anInt);
     });
 
-    it('should set and return a float', function () {
+    it.skip('should set and return a float #FRMW-1130', function () {
         var aFloat = 42.0000002;
 
         rxLocalStorage.setItem('aFloat', aFloat);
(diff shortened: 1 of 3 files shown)


utils/rx-page-objects/test/rxLocalStorage.midway.js
```diff
@@ -19,7 +19,7 @@ describe('utilities:rxLocalStorage', function () {
         expect(rxLocalStorage.getItem('anInt')).to.eventually.equal(anInt);
     });
 
-    it('should set and return a float', function () {
+    it.skip('should set and return a float', function () {
```
Droogans: If you're going to skip this, you need to track it via a defect in Jira. Once you create it, you can link back to it for future reference by updating the name of this test to:

```js
it.skip('should set and return a float #FRMW-0000', function () {
    // ....
});
```

I can then see stories that are skipped due to open defects quickly and easily in the future!

kevincareer427: @Droogans sorry missed that, it's added.

Droogans:
LGTM need a Dev signoff though.

CITguy:
@kevincareer427: please submit PRs against `1.x-LTS` and `2.x-LTS` with these same changes.

PR #1847, opened 2016-10-21
fix(windowTop): Remove usage of window.top and use window.parent

halkon:
JIRA:

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM


Droogans:
http://stackoverflow.com/questions/3473946/what-is-the-difference-between-window-window-top-and-window-parent

- `window` refers to the current frame.
- `parent` refers to the parent of the current frame.
- `top` refers to the outermost frame.

This allows for users who put encore apps into an iframe to not have origin accidentally refer to _their_ iframe, instead of the one we put there.

PR #1874, opened 2016-12-08
chore(ghPages): 3.x support for versioned docs

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-1219

### Generated Documentation
**This is not disrupting existing documentation at http://rackerlabs.github.io/encore-ui/.**

Generated 3.x docs served at http://rackerlabs.github.io/encore-ui/3.x/#/overview.
_NOTE:_ The version selection is not available yet in the navigation. This will come in the next PR.

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM
- [x] QE LGTM (no change to functional javascript)

### Summary
With the local development server, the navigation will behave like this.
version-nav-demo-local

PR #1875 added the necessary logic for it to behave like this on production.
version-nav-demo

PR #1875, opened 2016-12-08
feat(switchDocs): add version switching directive

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-1220

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM

### Summary
* This PR adds the shared `<switch-docs>` directive to be used by the different versions of the docs.
* This PR does not add the automatic forwarding to the latest release documentation.
* This PR is expected to be merged into the `gh-pages` branch.

### DIFF
--- a/components/switchDocs/switch-docs.html
+++ b/components/switchDocs/switch-docs.html
@@ -1,0 +1,4 @@
+<select rx-select
+  ng-model="version"
+  ng-options="v as (v.path + (v.isReleased ? '' : ' (unreleased)')) for v in versions | orderBy:['-path']">
+</select>
--- a/components/switchDocs/switchDocs.js
+++ b/components/switchDocs/switchDocs.js
@@ -1,0 +1,52 @@
+(function () {
+    angular
+        .module('demoApp')
+        .directive('switchDocs', SwitchDocsDirective);
+
+    function SwitchDocsDirective ($location, DOC_VERSIONS) {
+        return {
+            restrict: 'E',
+            templateUrl: '../components/switchDocs/switch-docs.html',
+            controller: function ($scope) {
+                $scope.versions = DOC_VERSIONS;
+
+                // Determine which version is currently being viewed
+                $scope.version = $scope.versions.filter(function (version) {
+                    return $location.absUrl().search('/' + version.path) >= 0;
+                })[0];
+
+                /* Navigate to different version on change */
+                $scope.$watch('version', function (newVal, oldVal) {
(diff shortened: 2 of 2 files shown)


CITguy:
See #1874 for the 3.x documentation logic.

CITguy:
Codeship will fail, because there are no tests. Will merge once we have 2 LGTMs.

components/switchDocs/switch-docs.html
```diff
@@ -0,0 +1,4 @@
+<select rx-select
+  ng-model="version"
+  ng-options="v as (v.path + (v.isReleased ? '' : ' (unreleased)')) for v in versions | orderBy:['-path']">
```
CITguy: There's logic here to add `(unreleased)` to versions that haven't been released or are in release candidate status.

Version | Label
----- | -----
4.x | `4.x (unreleased)`
3.x | `3.x`
2.x | `2.x`
1.x | `1.x`

We will not be adding 0.x documentation.

components/switchDocs/switchDocs.js
```diff
@@ -0,0 +1,52 @@
...
+                $scope.$watch('version', function (newVal, oldVal) {
+                    if (oldVal && (oldVal.path !== newVal.path)) {
+                        window.location.href = '../' + newVal.path;
+                    }
+                });
+            }
+        };
+    };//SwitchDocsDirective
+
+    angular
```
CITguy: `DOC_VERSIONS` is defined here to keep the versioned `index.html` changes to a minimum.

components/switchDocs/switchDocs.js
```diff
@@ -0,0 +1,52 @@
+(function () {
+    angular
+        .module('demoApp')
+        .directive('switchDocs', SwitchDocsDirective);
+
+    function SwitchDocsDirective ($location, DOC_VERSIONS) {
+        return {
+            restrict: 'E',
+            templateUrl: '../components/switchDocs/switch-docs.html',
+            controller: function ($scope) {
```
halkon: `controller` vs `link`? Any reason for `controller`?

CITguy: @halkon: Considering that this has to be compatible with multiple versions of Angular, I'm trying to keep it close to compatible with best-known methods for ng1.5+ "component" syntax without actually using `.component()`. Component syntax recommends using `controller` (using `controllerAs: "$ctrl"` syntax), instead of `link`, but I've kept it as simple as possible.

Either way, we can refactor after version switch functionality is implemented.

PR #1891, opened 2016-12-20
refactor(rxTokenInterceptor): rx-prefixes update to TokenInterceptor

cathysiller:
JIRA: https://jira.rax.io/browse/FRMW-1162

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM


src/utilities/rxTokenInterceptor/scripts/rxTokenInterceptor.js
```diff
@@ -12,14 +12,53 @@ angular.module('encore.ui.utilities')
...
+ * @name utilities.service:TokenInterceptor
+ */
 .provider('TokenInterceptor', function () {
     var exclusionList = this.exclusionList = [ 'rackcdn.com' ];
 
     this.$get = function (Session, $document) {
+        console.warn (
+            'DEPRECATED: TokenInterceptor - Please use rxTokenInterceptor. ' +
+            'TokenInterceptor will be removed in EncoreUI 4.0.0'
+        );
```
CITguy: Because `.provider()` objects deal with defining properties and methods on `this` (without `return`ing), we can't just create an alias to return another provider. Instead, we add a console warning to the `$get()` on the old name and copy the logic for the new provider.

Also, the warning had to be added to `$get()` because provider's are evaluated on load, not on consumption (resulting in console warnings appearing regardless if you use the provider).

parlarjb:
LGTM

halkon:
LGTM

PR #1915, opened 2017-01-11
refactor(rxFieldName): split rxFieldName into page object its own file

cathysiller:
JIRA: https://jira.rax.io/browse/FRMW-442

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM

# rxForm

### Before
rxform

### After
rxform-after

# rxFieldName
rxfieldname

### DIFF
--- a/utils/rx-page-objects/src/rxFieldName.page.js
+++ b/utils/rx-page-objects/src/rxFieldName.page.js
@@ -1,0 +1,66 @@
+'use strict';
+
+/**
+ * @class
+ */
+class rxFieldName {
+    /**
+     * @param {ElementFinder} rxFieldNameElement
+     * ElementFinder to be transformed into an rxFieldName object 
+     */
+    constructor (rxFieldNameElement) {
+        this.rootElement = rxFieldNameElement;
+
+        this.eleRequiredSymbol = this.rootElement.$('.required-symbol');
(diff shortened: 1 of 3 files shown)


utils/rx-page-objects/src/rxFieldName.page.js
```diff
@@ -0,0 +1,66 @@
...
+rxFieldName.initialize = function (rxFieldNameElement) {
+    console.warn(
+        'DEPRECATED: rxFieldName.initialize() ' +
+        'Use new rxFieldName() instead. ' +
+        'This function will be removed in EncoreUI 4.0.0 '
+    );
+    return new rxFieldName(rxFieldNameElement);
+};
+
+exports.rxFieldName = rxFieldName;
```
halkon: Why not do `module.exports = rxFieldName` rather than making it a property so that we don't have to do `require('...').rxFieldName`

CITguy: This is because of the way that `utils/rx-page-objects/index.js` imports the page objects and exercises. It has to be defined this way.

halkon: LGTM

CITguy:
Removed QE LGTM, as @Droogans has stated that the big green checkmark (showing tests pass) is enough for him.

PR #1930, opened 2017-02-06
chore(rxPageObjects): FRMW-1264 - Remove rx-page-objects 3.x Deprecations

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1264

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
-  ~~Design LGTM~~ Not Needed
- [x] QE LGTM


jj4th:
Rebased for 4.x changes

utils/rx-page-objects/package.json
```diff
@@ -12,15 +12,23 @@
...
   "devDependencies": {
+    "@types/chai-as-promised": "0.0.29",
+    "@types/jasmine": "^2.5.41",
+    "@types/lodash": "^4.14.50",
+    "@types/mocha": "^2.2.38",
+    "@types/moment": "^2.13.0",
+    "@types/node": "^7.0.3",
+    "@types/protractor": "^4.0.0",
+    "@types/selenium-webdriver": "^2.53.39",
+    "protractor": "^4.0.14",
```
Droogans: This is already in `dependencies` above.

utils/rx-page-objects/package.json
```diff
@@ -12,15 +12,23 @@
...
     "type": "git",
     "url": "git@github.com:rackerlabs/encore-ui.git"
   },
   "devDependencies": {
+    "@types/chai-as-promised": "0.0.29",
+    "@types/jasmine": "^2.5.41",
+    "@types/lodash": "^4.14.50",
+    "@types/mocha": "^2.2.38",
+    "@types/moment": "^2.13.0",
+    "@types/node": "^7.0.3",
```
Droogans: Shouldn't we be using the types for the node version we're using? We're not using version 7, we use version 5.

utils/rx-page-objects/src/rxMetadata.page.js
```diff
@@ -8,22 +8,33 @@ var rxMetadata = {
...
+
+                var definitionElement = termElement.$('.definition');
                 if (page.transformFns[metadataTerm] !== undefined) {
                     return page.transformFns[metadataTerm](definitionElement);
                 }
                 return definitionElement.getText();
-            }, fallbackReturnValue);
+            }, function () {
+                return fallbackReturnValue;
+            });
```
Droogans: Let's not convolute the logic here, just grab the term and give it back and have callers manage error callbacks themselves.

Might as well address transform logic stuff too. We should reduce hand holding like that as much as possible.

utils/rx-page-objects/test/rxPaginate.midway.js
```diff
@@ -1,143 +0,0 @@
-var Page = require('astrolabe').Page;
```
Droogans: :o

utils/rx-page-objects/package.json
```diff
@@ -12,15 +12,23 @@
...
+    "@types/protractor": "^4.0.0",
+    "@types/selenium-webdriver": "^2.53.39",
+    "protractor": "^4.0.14",
     "protractor-console-plugin": "^0.1.1",
-    "typescript": "2.0.2",
-    "typings": "^1.3.1"
```
Droogans: Check the post-install hook for `npm i`, pretty sure there's a few tasks around setting up TS in rx-page-objects.

package.json
```diff
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "test": "grunt test",
-    "postinstall": "cd utils/rx-page-objects; npm i; ./node_modules/.bin/typings install; cd -; grunt hooks",
+    "postinstall": "cd utils/rx-page-objects; npm i; cd -; grunt hooks",
```
CITguy: 👍

utils/rx-page-objects/package.json
Droogans: @jj4th Is it alright to use typings from a version of node that's two major versions ahead of what's [in our .node-version file](https://github.com/rackerlabs/encore-ui/blob/master/.node-version)?

jj4th: Pretty sure it doesn't matter.  In this specific case we're only using it for `require()` anyway, and that goes away once the typescript conversion is complete.  We don't really use the node API.

jj4th: @Droogans ^^

utils/rx-page-objects/src/rxMetadata.page.js
jj4th: This should be addressed in a separate PR though.

Droogans: Yeah I would make a Jira for it but it's more of an ethos to keep close to the front of our minds as we strip out much of what rx-page-objects is responsible for.

Droogans:
@parlarjb was present for a vidyo review.

Can you give this one last once over and mark LGTM when you have a chance?

parlarjb:
Vidyo LGTM

PR #1945, opened 2017-02-27
feat(FlexboxGrid): add 12-column grid system

CITguy:
* Deprecated attribute-based grid system
* Added 12-column, CSS class-based grid system

JIRA: https://jira.rax.io/browse/FRMW-1271

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] Design LGTM

### TL;DR

* Mobile-First!
* CSS Class-based
* Flexbox-based
* Cross-browser compatible
  * As long as the browser natively supports flexbox, it should work.
  * IE 11, Edge, Firefox, Chrome, Opera, Safari

#### Rows & Columns
direction | guttered | non-guttered
----- | ----- | -----
row | `.grid-row` | `.flex-row`
column | `.grid-col` | `.flex-col`

#### Column Width
non-responsive | responsive
----- | -----
`.span-N` | `.span-B-N`

* 1 &le; N &le; 12
* B = `xs` | `sm` | `md` | `lg` | `xl`

#### Column Offset
non-responsive | responsive
----- | -----
`.offset-N` | `.offset-B-N`

* 1 &le; N &le; 11
* B = `xs` | `sm` | `md` | `lg` | `xl`

#### Column Ordering
non-responsive | responsive
----- | -----
`.order-N` | `.order-B-N`

* 1 &le; N &le; 12
* B = `xs` | `sm` | `md` | `lg` | `xl`

### Screenshots
guttered-docs

gutterless-docs


src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,107 @@
+/* ===== CSS Class-Based Flexbox Grid System ===== */
+.grid-row {
+  box-sizing: border-box;
+  display: flex;
+  flex-grow: 1;
+  flex-wrap: wrap;
+  margin-top: 0;
+  margin-bottom: 0;
+  margin-left: -(@grid-gutter-width * 0.5) !important; // guard against accidental change
```
ssalinas24: As i am trying this out, i am thinking the negative margins are not a good idea. One situation that comes up is if a row is used inside of a native element with no grid or measurement on it.

src/elements/FlexboxGrid/docs/FlexboxGrid.demo.less
```diff
@@ -0,0 +1,73 @@
...
+        font-weight: bold;
+      }
+    }
+
+    .flex-col {
+      background-color: rgba(0, 0, 0, 0.12);
+      border-left: 1px dotted @gray-800;
+    }
+
+    // Automatically Set Column Content based on breakpoint (lg-2, md-8, etc.)
```
CITguy: Mixin and breakpoints below are for the dynamic responsive demos.

src/elements/FlexboxGrid/docs/FlexboxGrid.demo.less
```diff
@@ -0,0 +1,73 @@
+.FlexboxGrid-demo {
+  .rx-example {
+    .flex-row,
```
CITguy: These styles are for visualizing the rows and columns (normally invisible).

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  flex-shrink: 0;
+  flex-grow: 1;
+}
+
+.grid-row {
+  &:extend(.flex-row);
+  background-clip: content-box;
```
CITguy: If a `background-color` is applied to a row/column, this prevents the background color from jutting to far out (retains column expectation).

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+  border: 0;
+  margin: 0 -(@grid-gutter-width * 0.5);
```
CITguy: Negative margin required to align child contents with parent geometry.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+  padding: 0;
+}
+
+.grid-col {
+  &:extend(.flex-col);
+  background-clip: content-box;
+  border: 0;
+  margin: 0;
+  padding: 0 (@grid-gutter-width * 0.5);
```
CITguy: Gutters implemented via padding of the columns.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/* Grid Mixins */
+.grid-columns (@prefix, @n) when (@n > 0) and (@prefix = xs) {
+  .span-@{n},
+  .span-@{prefix}-@{n} {
```
CITguy: The non-responsive `.span-N` classes behave exactly as the `.span-xs-N` classes. This is because the `0em` xs breakpoint is the minimum width needed to apply these classes, thus it applies to every breakpoint, by default.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+    flex-grow: 0;
+    min-width: @width;
+    width: @width;
+  }
+  .grid-columns(@prefix, @n - 1);
+}
+
+.grid-offsets (@prefix, @n) when (@n > 0) and (@prefix = xs) {
+  .offset-@{n},
+  .offset-@{prefix}-@{n} {
```
CITguy: The `.offset-N` and `.offset-xs-N` classes behave similar to the `.span-*` classes above.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+.grid-offsets (@prefix, @n) when (@n > 0) and not (@prefix = xs) {
+  .offset-@{prefix}-@{n} {
+    margin-left: (@n * @grid-column-unit) !important;
+  }
+  .grid-offsets(@prefix, @n - 1);
+}
+
+.grid-orders (@prefix, @n) when (@n > 0) and (@prefix = xs) {
+  .order-@{n},
+  .order-@{prefix}-@{n} {
```
CITguy: The `.order-N` and `.order-xs-N` classes behave similar to the `.span-*` classes above.

src/styles/vars.less
```diff
@@ -502,3 +502,19 @@
...
+  - 1em is calculated on browser-defined font size
+    - No CSS property can affect this calculation.
+  - As font size increases/decreases (zoom %), more/less room is available,
+    so breakpoints behave more like larger/smaller screens
+*/
+@grid-break-wrist: 0em;
+@grid-break-palm: 40em; /* ~640px */
+@grid-break-lap: 64em; /* ~1024px */
+@grid-break-desk: 75em; /* ~1200px */
+@grid-break-wall: 90em; /* ~1440px */
```
CITguy: `em` units are used because they are more responsive to page zoom

megalord:
👍 luv it

halkon:
LATM

cathysiller:
Dev Vidyo LGTM

src/elements/FlexboxGrid/docs/FlexboxGrid.demo.less
```diff
@@ -0,0 +1,73 @@
...
+    .grid-row {
+      border-top: 1px dashed @blue-700;
+    }
+
+    .grid-col {
+      position: relative;
+      background-color: rgba(0, 0, 0, 0.12);
+
```
ssalinas24: below i took out all margins and spacing from rxMetadata and solely relied on the grid for layout. starting fresh, we need to think about how we want to have the bottom gutters and spacing of things and to what extent we need to include them in the grid or not.

Option 1:
We add bottom margins to rx-meta and the last one would be the spacing that helps space apart the grid-col items when they wrap and stack below one another.
Con: by default we are still not including bottom margin/spacing (aka gutters) of grid-col. So if the grid is used with plain HTML elements or divs, then by default are grid doesn't have the appropriate bottom gutters that it should have.

Option 2: 
We add bottom margin of 16px here on this `.grid-col` CSS.
Con: If `rx-meta` or something else at the bottom of a `.grid-col` also has bottom margin then the space between `.grid-col` that are stacked (.grid-col + .grid-col) then visually the spacing will be that plus the 16px on this line. Of course being a total of more than 16px.

Option 3:
We create classes for gutter spacing that would only apply bottom margins to grid-col
Same for rows.
Like maybe rows would have a class called `section` with 32px of spacing. `content-block` with 16px of margin could be a class for something like rx-meta or another element that just needs spacing between an element.
If you have two `content-block` elements aligned on the bottom of a `section`, then same doubled spacing mentioned in Option 2 is still a problem.

Seems like Option 2 and 3 would present more issues.

If using Option 1, then we would need a way for a class to be applied that removes bottom margins from an element. Something like `remove-vertical-gutter` or `remove-bottom-gutter`. These would be used on bottom elements that align along the bottom of a row that would have pre-defined bottom margins as well.

Look at the example below and let me know what option you think would be better from what i listed above?

Below is the example i am using on my local:
screen shot 2017-03-07 at 1 15 27 pm

src/elements/FlexboxGrid/docs/FlexboxGrid.demo.less
```diff
@@ -0,0 +1,73 @@
+.FlexboxGrid-demo {
+  .rx-example {
+    .flex-row,
+    .grid-row {
```
ssalinas24: similar to my other large comment below... we need to either add bottom margins to rows and have a way for a class to be applied and optionally remove. Or we don't put margins here, but we need helper classes that would assist a dev in applying bottom margin/gutters or optionally not having them.

src/elements/FlexboxGrid/docs/FlexboxGrid.demo.less
```diff
@@ -0,0 +1,73 @@
...
+
+      &.demo {
+        height: 1.5em;
+        font-weight: bold;
+      }
+    }
+
+    .flex-col {
+      background-color: rgba(0, 0, 0, 0.12);
+      border-left: 1px dotted @gray-800;
```
ssalinas24: same as other comments. However, while reviewing i'm really circling back to the idea that we need a helper class that would also work with .flex-col and .flex-row to add bottom margins or not be applied if not wanting bottom margin/gutters.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+
+.grid-row {
+  &:extend(.flex-row);
+  background-clip: content-box;
+  border: 0;
+  margin: 0 -(@grid-gutter-width * 0.5);
+  padding: 0;
+}
+
+.grid-col {
```
ssalinas24: below i took out all margins and spacing from rxMetadata and solely relied on the grid for layout. starting fresh, we need to think about how we want to have the bottom gutters and spacing of things and to what extent we need to include them in the grid or not.

Option 1:
We add bottom margins to rx-meta and the last one would be the spacing that helps space apart the grid-col items when they wrap and stack below one another.
Con: by default we are still not including bottom margin/spacing (aka gutters) of grid-col. So if the grid is used with plain HTML elements or divs, then by default are grid doesn't have the appropriate bottom gutters that it should have.

Option 2: 
We add bottom margin of 16px here on this `.grid-col` CSS.
Con: If `rx-meta` or something else at the bottom of a `.grid-col` also has bottom margin then the space between `.grid-col` that are stacked (.grid-col + .grid-col) then visually the spacing will be that plus the 16px on this line. Of course being a total of more than 16px.

Option 3:
We create classes for gutter spacing that would only apply bottom margins to grid-col
Same for rows.
Like maybe rows would have a class called `section` with 32px of spacing. `content-block` with 16px of margin could be a class for something like rx-meta or another element that just needs spacing between an element.
If you have two `content-block` elements aligned on the bottom of a `section`, then same doubled spacing mentioned in Option 2 is still a problem.

Seems like Option 2 and 3 would present more issues.

If using Option 1, then we would need a way for a class to be applied that removes bottom margins from an element. Something like `remove-vertical-gutter` or `remove-bottom-gutter`. These would be used on bottom elements that align along the bottom of a row that would have pre-defined bottom margins as well.

Look at the example below and let me know what option you think would be better from what i listed above?

Below is the example i am using on my local:
screen shot 2017-03-07 at 1 15 27 pm

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
+/* ===== CSS Class-Based Flexbox Grid System ===== */
+.flex-row {
+  box-sizing: border-box;
+  display: flex;
+  flex-grow: 1;
+  flex-wrap: wrap;
+}
+
+.flex-col {
```
ssalinas24: same as other comments. However, while reviewing i'm really circling back to the idea that we need a helper class that would also work with .flex-col and .flex-row to add bottom margins or not be applied if not wanting bottom margin/gutters.

src/elements/FlexboxGrid/styles/grid.less
```diff
@@ -0,0 +1,111 @@
...
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  flex-shrink: 0;
+  flex-grow: 1;
+}
+
+.grid-row {
```
ssalinas24: similar to my other large comment below... we need to either add bottom margins to rows and have a way for a class to be applied and optionally remove. Or we don't put margins here, but we need helper classes that would assist a dev in applying bottom margin/gutters or optionally not having them.

PR #1948, opened 2017-03-03
fix(rxDatePicker): correct empty string validity

cathysiller:
JIRA: https://jira.rax.io/browse/FRMW-1274

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM


lentusignavus:
LGTM

PR #1951, opened 2017-03-06
refactor(rxEnvironment): stop using rxEnvironment.get()

cathysiller:
JIRA: https://jira.rax.io/browse/FRMW-1253

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM

### DIFF
--- a/src/utilities/rxFavicon/scripts/rxFavicon.js
+++ b/src/utilities/rxFavicon/scripts/rxFavicon.js
@@ -42,10 +42,10 @@ angular.module('encore.ui.utilities')
             var currentEnv;
             if (rxEnvironment.isLocal()) {
                 currentEnv = 'local';
-            } else if (rxEnvironment.isUnifiedPreProd()) {
-                currentEnv = 'staging';
-            } else {
+            } else if (rxEnvironment.isPreProd() || rxEnvironment.isUnifiedProd()) {
                 currentEnv = 'prod';
+            } else {
+                currentEnv = 'staging';
             }
 
             // update href to use new path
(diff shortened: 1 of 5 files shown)


src/utilities/rxUrlUtils/scripts/rxUrlUtils.js
```diff
@@ -106,6 +106,7 @@ angular.module('encore.ui.utilities')
         }
 
         // run the href through rxEnvironmentUrl in case it's defined as such
+        // remove in EncoreUI 4.0.0
         url = rxEnvironmentUrlFilter(url);
```
CITguy: As it currently stands, the `rxEnvironmentUrl` filter is just passing through the `url` value. However, removing this line causes errors in `rxAppRoutes`. Given that `rxAppRoutes` is slated for removal in 4.0.0, we'll revisit removing this line in 4.0.0 updates.

CITguy:
We'll wait for @parlarjb to give the 👍  before merging this one.

src/utilities/rxFavicon/scripts/rxFavicon.js
```diff
@@ -38,23 +38,18 @@ angular.module('encore.ui.utilities')
             favicons.prod = el.attr('href');
             favicons.staging = favicons.staging || favicons.prod;
             favicons.local = favicons.local || favicons.staging;
+            
+            var currentEnv;
+            if (rxEnvironment.isLocal()) {
+                currentEnv = 'local';
+            } else if (rxEnvironment.isUnifiedPreProd()) {
+                currentEnv = 'staging';
```
parlarjb: Unfortunately this logic isn't correct, as `isUnifiedPreProd()` will return true for staging _and_ preprod.

I'd do this:

```
currentEnv = 'staging';
if (rxEnvironment.isLocal()) {
    currentEnv = 'local';
} else if (rxEnvironemnt.isPreProd() || rxEnvironment.isUnifiedProd() {
    currentEnv = 'prod'
}
```

or something along those lines. Our preprod behaviour is actually undefined here, so I'm not sure what the right thing is to do

cathysiller: updated

parlarjb:
LGTM

PR #1962, opened 2017-03-14
Frmw 1273 convert rx page objects to type script

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1273

### LGTMs
- [ ] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [ ] Design LGTM
- [ ] QE LGTM

This is a huge PR, but it converts all rxPageObjects to TypeScript.  It also converts the mock tests for the page objects, and provides linting support.

One thing about this PR is there are multiple places where I remove comments.  This is because much of the commenting needs to be reformatted to work for TypeScript and so that it can support intellisense in vscode/other IDEs.  I started that work during this process, but due to time constraints, am leaving it until later to finish.  I may need @Droogans help/assistance to update the test environment/visreg to node 6 for this to pass.


jj4th:
Going to chunk this up a bit.

PR #1965, opened 2017-03-16
Frmw 1273 convert rx page objects to type script part1

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1273

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [x] QE LGTM

This PR covers four things:
- Initial setup of TypeScript and the TSLinter
- Removal of `rxBreadcrumbs` and `rxAccountInfo` page objects
- Porting of the following page objects/midways to TypeScript: `rxTimePicker(Util)`, `rxSelect`, `rxRadio`, `rxDatePicker`, `rxCheckbox`.
- Minimal changes to support the above in the following midways: `rxAutoSave`, `rxBulkSelect`; and page objects: `rxCopy`, `rxFeedback`, `rxForm`, and `rxMultiSelect`


package.json
```diff
@@ -15,7 +15,8 @@
     "test": "grunt test",
     "postinstall": "cd utils/rx-page-objects; npm i; cd -; grunt hooks",
     "eslint-dev": "eslint -c .eslintrc.json --ignore-path .eslintignore-dev {src,demo}",
-    "eslint-test": "eslint -c .eslintrc.test.json --ignore-path .eslintignore-test {src,demo} utils/rx-page-objects/{test,src}"
+    "eslint-test": "eslint -c .eslintrc.test.json --ignore-path .eslintignore-test {src,demo}",
+    "tslint-test": "cd utils/rx-page-objects; $(npm bin)/tslint -c tslint.json {test,src}/*.ts"
```
CITguy: Do we need the `$(npm bin)`? I thought that these scripts ran in the context of the locally install npm modules.

utils/rx-page-objects/src/rxCheckbox.page.ts
```diff
@@ -0,0 +1,113 @@
+'use strict';
+
+import * as _ from 'lodash';
+import {ElementFinder} from 'protractor';
+import {by} from 'protractor';
```
CITguy: Can 4 and 5 be combined into the following?

```typescript
import {by, ElementFinter} from 'protractor';
```

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -1,79 +1,67 @@
...
-import * as moment from 'moment';
 import * as _ from 'lodash';
-import * as webdriver from 'selenium-webdriver';
-import {ElementFinder, ElementArrayFinder} from 'protractor';
-import {$$, browser, by} from 'protractor';
-import {rxComponentElement, AccessorPromiseString, Promise} from './rxComponent';
-let rxSelect = require('./rxSelect.page').rxSelect;
+import * as moment from 'moment';
+import {ElementFinder} from 'protractor';
+import {by} from 'protractor';
```
CITguy: Can we combine line 5 and 6 to the following?

```typescript
import {by, ElementFinder} from 'protractor';
```

utils/rx-page-objects/src/rxCheckbox.exercise.ts
```diff
@@ -1,59 +1,58 @@
...
- * @param {Boolean} [options.selected=false] - Whether the checkbox is selected at the start of the exercise
- * @param {Boolean} [options.visible=true] - Whether the checkbox is visible at the start of the exercise
- * @param {Boolean} [options.valid=true] - Whether the checkbox is valid at the start of the exercise
  */
-exports.rxCheckbox = function (options) {
-    if (options === undefined) {
-        options = {};
-    }
-
+export function rxCheckbox(options: IRxCheckboxExerciseOptions) {
```
CITguy: Does this syntax make `options` required or will it default to an empty object like it did before?

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -1,79 +1,67 @@
...
+                this.close();
+            }
+
+            return month;
         });
     }
     set month(value: AccessorPromiseString) {
         this.open();
-        let dropdownExpectedValue = moment(`2000-${value}`, 'YYYY-MM').format('MMM');
+        let dropdownExpectedValue = moment(<string> value, 'MM').format('MMM');
```
CITguy: What is this syntax doing? I'm unfamiliar with this notation.

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -1,79 +1,67 @@
...
         if (dropdownExpectedValue === 'Invalid date') {
             throw new Error(
-                `Unexpected month value for month number "${value}". Months are not zero-indexed!`
+                `Unexpected month value for month number "${value}". Months are not zero-indexed!`,
```
CITguy: Why the trailing comma?

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -235,56 +183,36 @@ export class rxDatePicker extends rxComponentElement {
 
     /**
      * @description Whether or not the date requested is currently selected.
-     * @param {String} date
-     * `YYYY-MM-DD` formatted date to check if it is currently selected.
-     * @returns {Promise<Boolean>}
      */
-    isDateSelected(date: string): Promise<boolean> {
+    isDateSelected(date: string) {
```
CITguy: Why are we leaving off the return type here, but not in previous methods like `nextMonth()` and `previousMonth()`?

utils/rx-page-objects/src/rxRadio.exercise.ts
```diff
@@ -1,16 +1,12 @@
...
- * @exports exercise/rxRadio
- * @param {Object} options - Test options. Used to build valid tests.
- * @param {rxRadio} options.instance - Component to exercise.
- * @param {Boolean} [options.disabled=false] - Whether or not the radio button is disabled at the start of the exercise.
- * @param {Boolean} [options.selected=false] - Whether or not the radio button is selected at the start of the exercise.
- * @param {Boolean} [options.visible=true] - Whether or not the radio button is visible at the start of the exercise.
- * @param {Boolean} [options.valid=true] - Whether or not the radio button is valid at the start of the exercise.
  */
-exports.rxRadio = function (options) {
+export function rxRadio(options) {
```
CITguy: Did you want to define an interface for these options?

utils/rx-page-objects/src/rxRadio.page.ts
```diff
@@ -0,0 +1,93 @@
+'use strict';
+
+import * as _ from 'lodash';
+import {ElementFinder} from 'protractor';
+import {by} from 'protractor';
```
CITguy: Can we combine line 4 and 5 into the following?

```typescript
import {by, ElementFinder} from 'protractor';
```

utils/rx-page-objects/src/rxSelect.exercise.ts
```diff
@@ -1,52 +1,56 @@
...
- * @param {Object} options - Test options. Used to build valid tests.
- * @param {rxSelect} options.instance - Component to exercise.
- * @param {Boolean} [options.disabled=false] - Determines if the select is disabled
- * @param {Boolean} [options.visible=true] - Determines if the select is visible
- * @param {Boolean} [options.valid=true] - Determines if the select is valid
- * @param {String} [selectedText] - The expected selected text of the dropdown.
  */
-exports.rxSelect = function (options) {
+export function rxSelect (options: IRxSelectExerciseOptions) {
     if (options === undefined) {
```
CITguy: I'm seeing inconsistencies with how we deal with exercise options.  `rxCheckbox` doesn't default if `options` isn't defined, yet `rxRadio` and `rxSelect` leave the logic in place to default. Which should it be?

utils/rx-page-objects/src/rxSelect.page.ts
```diff
@@ -0,0 +1,138 @@
+'use strict';
+
+import * as _ from 'lodash';
+import {ElementFinder} from 'protractor';
+import {by} from 'protractor';
```
CITguy: Can we combine line 4 and 5 into the following?

```typescript
import {by, ElementFinder} from 'protractor';
```

package.json
jj4th: I added `$(npm bin)` because it didn't work without it.  I think that it's because `tslint` is only installed at the `utils/rx-page-objects` level, perhaps?

utils/rx-page-objects/src/rxCheckbox.page.ts
jj4th: Yes, I will do that, this is an artifact from an automatic optimization, these lines used to be a lot longer.

utils/rx-page-objects/src/rxDatePicker.page.ts
jj4th: Yes, I will do that, this is an artifact from an automatic optimization, these lines used to be a lot longer.

utils/rx-page-objects/src/rxDatePicker.page.ts
jj4th: It's casting `value` to a `<string>`. `moment()` doesn't accept the full type of `AccessorPromiseString`.  This is an artifact of typescript not really liking it when getters and setters have different types.

utils/rx-page-objects/src/rxDatePicker.page.ts
jj4th: My guess is tslint's optimizer thought it was a list.  I'll remove the comma.

utils/rx-page-objects/src/rxDatePicker.page.ts
jj4th: Good catch, the previous two should have the type removed, it's unnecessary.

utils/rx-page-objects/src/rxCheckbox.exercise.ts
jj4th: Options is always required, because you always have to pass in an instance to exercises.  There are no more default selectors for components.

utils/rx-page-objects/src/rxRadio.exercise.ts
jj4th: Good catch!  I'll fix this tomorrow.

utils/rx-page-objects/src/rxRadio.page.ts
jj4th: Same as above.

utils/rx-page-objects/src/rxSelect.exercise.ts
jj4th: This is an artifact.  Because options is not marked as an optional parameter, it actually can never be undefined.  I'll remove this conditional block.

utils/rx-page-objects/src/rxSelect.page.ts
jj4th: Same answer as above :)

package.json
jj4th: Verified, that is why this is happening.

jj4th:
Didn't we already have a Vidyo LGTM of this?

CITguy:
@jj4th, the `Vidyo LGTM` is for someone who was on the Vidyo session to provide their LGTM/Approval. It's not an automatic LGTM just by having a Vidyo session.

utils/rx-page-objects/src/rxComponent.ts
```diff
@@ -1,10 +1,14 @@
...
-import {ElementFinder, browser} from 'protractor';
 
 export class rxComponentElement extends ElementFinder {
-    constructor(rootElement: ElementFinder) {
-        super(browser, rootElement.elementArrayFinder_);
+    originalElement: ElementFinder;
+
+    constructor(originalElement: ElementFinder) {
+        super(browser, originalElement.elementArrayFinder_);
+        this.originalElement = originalElement;
```
Droogans: I think this should be `this._originalElement` to make it even more obvious that it is not a replacement for legacy `this.rootElement` usage.

utils/rx-page-objects/src/rxComponent.ts
```diff
@@ -13,3 +17,19 @@ export class rxComponentElement extends ElementFinder {
...
+ */
+export function OverrideWebdriver(target: Object, propertyKey: string | symbol): PropertyDescriptor {
+    let original = target[propertyKey];
+    return {
+        get() {
+            return original;
+        },
+        set() {
+            return null;
+        },
```
Droogans: Extra comma[.](http://izquotes.com/quotes-pictures/quote-this-is-the-sort-of-pedantry-up-with-which-i-will-not-put-winston-churchill-340357.jpg)

utils/rx-page-objects/src/rxCopy.exercise.js
```diff
@@ -101,7 +101,7 @@ exports.rxCopy = function (options) {
             });
 
             // Skip on Chrome for Mac: CMD-V is not working to paste clipboard contents
-            if (options.testCopyArea && !(isMac && isChrome)) {
+            if (options.testCopyArea && !(browser.params.isMac && browser.params.isChrome)) {
```
Droogans: This is getting exported to ~~all users~~ the user.

I say leave it -- but maybe just throw up some sane defaults around `isMac` (false), and `isChrome` (false), hard coded into the exercise. I think the `options` should look for the browser params, but then `||` out into the defaults.

utils/rx-page-objects/src/rxDatePicker.exercise.ts
```diff
@@ -182,63 +169,63 @@ export function rxDatePicker (options: rxDatePickerExerciseOptions) {
...
                 expect(datepicker.date).to.eventually.equal(firstOfMonth);
             });
 
-            it('should update the date to the last of the month', function () {
-                var lastOfMonth = moment().endOf('month').format(isoFormat);
+            it('should update the date to the last of the month', () => {
+                let lastOfMonth = moment().endOf('month').format(isoFormat);
                 datepicker.date = lastOfMonth;
                 expect(datepicker.date).to.eventually.equal(lastOfMonth);
             });
```
Droogans: This test isn't doing anything anymore now that there are strings being used to process dates.

The only reason this test is here is because the datepicker took date objects, which were notorious for giving you issues near day/month/etc. boundaries when running tests in environments that used UTC, for instance.

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -1,79 +1,64 @@
...
-
-                return moment(`${month} 2000`, 'MMM YYYY').format('MM');
+            let month = new rxSelect(this.currentMonth).selectedOption.getText().then(text => {
+                return moment(text, 'MMM').format('MM');
             });
+
+            // if datepicker was closed before starting, put it back
+            if (!isOpen) {
+                this.close();
+            }
```
Droogans: :1st_place_medal:

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -1,79 +1,64 @@
...
-        }
-
-        super(rxDatePickerElement);
-        this.rootElement = rxDatePickerElement; // deprecate in 3.x
+    private get currentMonth() {
+        return this.element(by.model('currentMonth'));
+    }
+    private get currentYear() {
+        return this.element(by.model('currentYear'));
+    }
```
Droogans: Shouldn't this return text of some sort? These appear to be `eleCurrentMonth` and `eleCurrentYear`, respectively.

Also, please add a blank line between members.

utils/rx-page-objects/src/rxDatePicker.page.ts
```diff
@@ -207,84 +154,62 @@ export class rxDatePicker extends rxComponentElement {
...
     };
 
     /**
-     * @private
      * @description Click back to the previous month in the calendar.
      */
-    previousMonth(): Promise<void> {
+    previousMonth() {
         this.open();
         return this.$('.arrow.prev').click();
```
Droogans: Think this (and `nextMonth()`) should close the datepicker when they're done? It may introduce a side effect if you don't, which would need to be documented.

I think it's best to check if it's already open, then interact with it, based on the initial state, leave it how it was when you found it.

utils/rx-page-objects/src/rxRadio.exercise.ts
```diff
@@ -0,0 +1,57 @@
...
+        selected: false,
+        visible: true,
+        valid: true,
+    });
+
+    return () => {
+        let component: component.rxRadio;
+
+        before(() => {
+            component = options.instance;
```
Droogans: Are you doing this to get around some undef compile errors for `component`?

```js
let component = component.rxRadio
```

isn't clicking for me.

utils/rx-page-objects/src/rxComponent.ts
jj4th: TSLint requires these commas, you'll see this in all object literal and list literal declarations.

utils/rx-page-objects/src/rxComponent.ts
jj4th: Fair.

utils/rx-page-objects/src/rxCopy.exercise.js
jj4th: Hmm, I get what you're saying, but how is it any different than what we had before?  I didn't change the use of these, just moved them from the onPrepare to the params in protractor.

utils/rx-page-objects/src/rxRadio.exercise.ts
jj4th: it's `let component: component.rxRadio` it's a type, not an assignment.  And no, this is just the way that literally every exercise was written.  if I were to revisit this we would not `import * as component from './rxRadio.page'`  Instead I would `import {rxRadio} from './rxRadio.page'`

utils/rx-page-objects/src/rxDatePicker.page.ts
jj4th: I'd be more inclined to remove the `this.open()` then add a bunch of additional logic.  I'd have to look at the tests to see why this works just fine there.

utils/rx-page-objects/src/rxRadio.exercise.ts
Droogans: Ah, totally overlooked that. My bad. :man_facepalming:

utils/rx-page-objects/src/rxSelect.exercise.ts
```diff
@@ -1,52 +1,52 @@
...
         disabled: false,
         visible: true,
-        valid: true
+        valid: true,
     });
 
-    return function () {
-        var component;
+    return () => {
+        let component;
```
Droogans: Shouldn't this have a type signature like the rxRadio exercise above?

```js
let component: component.rxSelect;
```

Also, I believe you need to add `instance` to the options interface on line 20.

utils/rx-page-objects/src/rxSelect.page.ts
```diff
@@ -0,0 +1,137 @@
...
+import {by, ElementFinder} from 'protractor';
+import {OverrideWebdriver, rxComponentElement} from './rxComponent';
+
+/**
+ * @class
+ */
+export class rxSelect extends rxComponentElement {
+
+    get eleWrapper() {
+        return this.element(by.xpath('..'));
```
Droogans:
```js
return this.parent;
```

utils/rx-page-objects/src/rxSelect.page.ts
```diff
@@ -0,0 +1,137 @@
...
+        return this.option(optionText).click();
+    }
+}
+
+/**
+ * @description Generates a getter and a setter for an rxSelect element on your page.
+ * @example
+ * class MyForm {
+ *     @rxSelectAccessor(element(by.model('states'))) state;
+ *     // you can specify a single dropdown's slow clicking globally this way
```
Droogans: Drop this line. Slow clicking is gone for now, if it does appear again, it'll be shoved down into consumers.

utils/rx-page-objects/src/rxTimePickerUtil.page.ts
```diff
@@ -0,0 +1,15 @@
...
+     * @description return offset value, if present in string
+     *
+     * **NOTE:** Logic in this function must match the logic in
+     * the rxTimePickerUtil service.
+     */
+    export function parseUtcOffset(stringValue: string) {
+        let regex = /([-+]\d{2}:?\d{2})/;
+        let matched = stringValue.match(regex);
+        return (matched ? matched[0] : '');
+    }
```
Droogans: @jj4th tell me your thoughts on whether or not you think this should be a part of rxTimePicker's class methods, anonymously used in rxTimePicker, or exported as a static function in rxTimePicker (or something else entirely).

utils/rx-page-objects/src/rxTimePicker.page.ts
```diff
@@ -0,0 +1,185 @@
...
+        return this.$$('rx-inline-error').getText();
+    }
+
+    get hours(): AccessorPromiseString {
+        return this.txtHour.getAttribute('value');
+    }
+    set hours(hour) {
+        this.txtHour.clear();
+        this.txtHour.sendKeys(hour);
+    }
```
Droogans: Can we create a decorator for creating text accessors, instead of writing it out this way every time?

utils/rx-page-objects/src/rxTimePicker.page.ts
```diff
@@ -0,0 +1,185 @@
...
+    get pagePeriod() {
+        return new rxSelect(this.selPeriod);
+    }
+    get pageUtcOffset() {
+        return new rxSelect(this.selUtcOffset);
+    }
+
+    getErrors() {
+        return this.$$('rx-inline-error').getText();
+    }
```
Droogans: @jj4th I know you didn't write this, but....

```js
get errors() {
    // ...
}
````

Thoughts?

File: utils/rx-page-objects/src/rxTimePicker.page.ts [hunk 26]
```diff
@@ -0,0 +1,185 @@
...
+        // force to time zone from input
+        date.utcOffset(offset);
+
+        // set via picker
+        this.open();
+
+        this.hours = date.format('hh'); // 12-hour format (no padding)
+        this.minutes = date.format('mm'); // padded minutes
+        this.selPeriod.element(by.cssContainingText('option', date.format('A'))).click();
+        this.selUtcOffset.element(by.cssContainingText('option', date.format('Z'))).click();
```
Droogans: I think this was a workaround because the dropdowns weren't working, IIRC.

Can this be changed to use the `pagePeriod` and `pageUtcOffset` members instead?

utils/rx-page-objects/test/rxTimePicker.midway.ts
```diff
@@ -0,0 +1,331 @@
...
+        });
+
+        describe('time', () => {
+            it('should be blank', () => {
+                expect(picker.time).to.eventually.be.empty;
+            });
+
+            it('should change using ISO 8601 time string', () => {
+                picker.time = '20:00-04:00';
+                expect(picker.time).to.eventually.eq('20:00-04:00');
```
parlarjb: The string in the `eq()` is identical to the assignment on the line above. Is that intentional? The test description implies it should change it

parlarjb: The string in the eq() is identical to the assignment on the line above. Is that intentional? The test description implies it should change it

utils/rx-page-objects/test/rxTimePickerUtil.midway.ts
```diff
@@ -0,0 +1,34 @@
...
+            ['8:00 (-06:00)', '-06:00'],
+            ['13:00 (UTC-0800)', '-0800'],
+            ['20:00-04:00', '-04:00'],
+            ['non-time string', ''],
+            ['20:00-0400', '-0400'],
+            ['20:00-400', ''],
+            ['20:00-4', ''],
+        ].forEach(strPair => {
+            let strInput = strPair[0];
+            let strOutput = strPair[1];
```
Droogans: You can use argument unpacking to avoid the chopping up of the output here in these later lines.

```js
].forEach(([strInput, strOutput]) => {
    // ...
});
```

utils/rx-page-objects/test/rxTimePicker.midway.ts
jj4th: @parlarjb it's because `picker.time` is a complex setter, this test is testing that the setter works.  In general this pattern is deprecated but it was kept for datePicker and timePicker because otherwise working with them is super annoying (or you end up having `getTime` `setTime` methods).

utils/rx-page-objects/src/rxRadio.page.ts
```diff
@@ -0,0 +1,92 @@
...
+import {by, ElementFinder} from 'protractor';
+import {OverrideWebdriver, rxComponentElement} from './rxComponent';
+
+/**
+ * @class
+ */
+
+export class rxRadio extends rxComponentElement {
+    get eleWrapper() {
+        return this.element(by.xpath('..'));
```
jj4th: Sure, I'd love to implement `this.parent`!

utils/rx-page-objects/src/rxSelect.exercise.ts
jj4th: It's there, but it shouldn't be optional so I'll change that.  And yes on the type dec missing... ugh.

utils/rx-page-objects/test/rxTimePicker.midway.ts
parlarjb: I get that it's a setter. But the test description says "should change". But this test is not showing any kind of change. From the outside, you're assigning a value and testing that the same value is still there.

Maybe just the description needs to change?

jj4th: I'm changing the date, then checking that the date was changed?  I'm confused by your confusion 😸

utils/rx-page-objects/src/rxRadio.page.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import * as _ from 'lodash';
+import {by, ElementFinder} from 'protractor';
+import {OverrideWebdriver, rxComponentElement} from './rxComponent';
+
+/**
+ * @class
+ */
+
+export class rxRadio extends rxComponentElement {
```
jj4th: Not really handling docs right now.  I'll remove the space.  tslint is failing me wrt spacing.

utils/rx-page-objects/test/rxTimePicker.midway.ts
parlarjb: Here's the "change"

```
picker.time = '20:00-04:00';
```

You're then checking if `picker.time` is equal to `'20:00-04:00'`, which, unless I'm mis-reading, is the same string I just pasted above. i.e. nothing is changed.

utils/rx-page-objects/src/rxTimePicker.page.ts
jj4th: It's resolving the promise to text, hence the `getPrefix()`

utils/rx-page-objects/src/rxTimePickerUtil.page.ts
jj4th: Judging by the midway, it seems like it needs to be exported, but we could definitely make it a static member of rxTimePicker.

utils/rx-page-objects/src/rxTimePicker.page.ts
jj4th: Remember that decorators require a static locator specified at compile time.  Still it's possible.

utils/rx-page-objects/test/rxTimePickerUtil.midway.ts
jj4th: Yeah missed that.

utils/rx-page-objects/test/rxTimePicker.midway.ts
jj4th: Here is what happens in order:

1. The picker has a pre-determined state, in this case empty.
2. I'm changing the date from empty, to '20:00-04:00'.
3. I'm verifying that the date changed to '20:00-04:00'.

I could use the word "set" if you prefer that? `should set using ISO 8601 time string`

utils/rx-page-objects/src/rxTimePickerUtil.page.ts
Droogans: I feel like it's currently mirroring the dev codebase setup too much. I agree that it should be a static method on the timepicker class.

utils/rx-page-objects/test/rxTimePicker.midway.ts
parlarjb: Ok, now I understand. `set` makes sense, and maybe just a comment along the lines of

"This test is to verify that the logic in the setter doesn't screw up, and what we pass in is what we'll get out of the getter:

utils/rx-page-objects/src/rxCopy.exercise.js
Droogans: I just realized that if left undefined, it'll be falsey. That'll work just fine on our standard Linux+FF setups.

utils/rx-page-objects/src/rxTimePicker.page.ts
jj4th: @Droogans and I determined this wasn't possible 🐱

PR #1970, opened 2017-03-23
fix(navHighlighting): correct multi-highlight [4.x]

CITguy:
JIRA: https://jira.rax.io/browse/FRMW-1238

### LGTMs
- [x] Dev LGTM
- [ ] Dev+Vidyo LGTM

### BEFORE
screen shot 2017-03-23 at 4 59 43 pm

### AFTER
screen shot 2017-03-23 at 4 59 33 pm

### DIFF
--- a/demo/scripts/demoApp.directive.js
+++ b/demo/scripts/demoApp.directive.js
@@ -92,17 +92,20 @@
                 } else if (!_.isEmpty(route.href)) {
                     // Prepend a "#" for easier matching
                     var normalizedUrl = '#' + currentUrl;
+                    var regexHref = new RegExp(route.href + '\\b', 'g');
 
                     /*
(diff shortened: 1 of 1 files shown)


demo/scripts/demoApp.directive.js
```diff
@@ -92,17 +92,20 @@
                 } else if (!_.isEmpty(route.href)) {
                     // Prepend a "#" for easier matching
                     var normalizedUrl = '#' + currentUrl;
+                    var regexHref = new RegExp(route.href + '\\b', 'g');
```
CITguy: Using the word boundary escape character (`\b`) we can fairly reliably match full routes and avoid partial route matches causing false positives.

cathysiller:
Dev LGTM

PR #1983, opened 2017-04-12
chore(rxPageObjects): Update docs README.md

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1331

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [ ] Design LGTM
- [x] QE LGTM

This PR:
- Updates the examples in the `README.md` file. (Removes rxBreadcrumbs which no longer exists)
- Adds a note on extending Protractor API's `ElementFinder`.

PR #1984, opened 2017-04-12
chore(rxPageObjects): Update root README.md

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1332

### LGTMs
- [ ] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [ ] Design LGTM
- [ ] QE LGTM

This PR:
- Updates the examples in the `README.md` file.
- Adds a note on extending Protractor API's `ElementFinder`.

This is the actual root readme file, not the one used by docs. (see: https://github.com/rackerlabs/encore-ui/pull/1983)

### DIFF
--- a/utils/rx-page-objects/README.md
+++ b/utils/rx-page-objects/README.md
@@ -8,47 +8,59 @@ Many of the components that ship with EncoreUI require end to end testing. Why s
 
 **good**:
 ```js
-it('should have actually sorted the column ascending', function () {
-    var column = element(by.cssContainingText('rx-sortable-column', 'Name'));
+it('should have actually sorted the column ascending', () => {
+    let names = ['Anne', 'Bob', 'Charlie'];
+    let column = element(by.cssContainingText('rx-sortable-column', 'Name'));
     column.$('i.sort-icon').click();
     column.$('i.sort-icon').click();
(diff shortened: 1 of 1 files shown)


utils/rx-page-objects/README.md
```diff
@@ -8,47 +8,59 @@ Many of the components that ship with EncoreUI require end to end testing. Why s
...
+}
 
-it('should have actually sorted the column ascending', function () {
-    var column = myPage.column('Name');
-    column.sortAscending();
-    column.data.then(function (names) {
-        expect(names).to.eql(names.sort());
-    });
+// This allows you to more easily override for specific columns with odd data requirements.
+class SpecialColumn extends Colunn {
```
Droogans: `extends Column`

jj4th:
@Droogans Is there a compelling reason not to use the docs readme here too?  What do you feel we should add to be able to just keep one `README.md`?

Droogans:
I am all for using the other readme in place of this one. It shows you everything you need to know.

The high level "use it directly" approach, how to dig specifics out of it, and exercises. That's really all this thing is doing anymore, and to spend much more time or effort on it would likely start distracting from those simple tenants.

PR #1991, opened 2017-05-19
fix(angular): Angular 1.6 upgrade

halkon:
JIRA: https://jira.rax.io/browse/FRMW-1337

### LGTMs
- [x] Dev LGTM
- [x] QE LGTM

This PR has the following:

* Upgrade of Angular to 1.6
* Removal of UI Bootstrap
* Removal of CFP Hotkeys

Helpful reference to understand the changes: https://docs.angularjs.org/guide/migration


bower.json
```diff
@@ -6,29 +6,30 @@
...
-    "angular-route": "~1.3.20",
-    "angular-messages": "~1.3.20",
-    "angular-bootstrap": "0.14.3",
+    "angular": "1.6.4",
+    "angular-animate": "1.6.4",
+    "angular-sanitize": "1.6.4",
+    "angular-resource": "1.6.4",
+    "angular-route": "1.6.4",
+    "angular-messages": "1.6.4",
+    "angular-bootstrap": "2.5",
```
CITguy: eww

bower.json
```diff
@@ -6,29 +6,30 @@
...
     "lodash": "4.x",
     "moment": "~2.9.0",
     "html2canvas": "~0.4.1",
-    "angular-hotkeys": "chieffancypants/angular-hotkeys#~1.4.0",
+    "angular-hotkeys": "chieffancypants/angular-hotkeys#1.7",
     "font-awesome": "~4.4.0",
-    "ng-debounce": "0.1.5"
+    "ng-debounce": "0.1.5",
+    "angular-mocks": "1.6.4"
```
CITguy: `angular-mocks` needs added?

src/elements/Typeahead/scripts/typeahead.js
```diff
@@ -24,7 +24,7 @@
  */
 angular.module('encore.ui.elements')
 .config(function ($provide) {
-    $provide.decorator('typeaheadDirective', function ($delegate, $filter) {
+    $provide.decorator('uibTypeaheadDirective', function ($delegate, $filter) {
```
CITguy: I have a feeling ngBootstrap is going to be a thorn in our side for the upgrade.

bower.json
halkon: It got moved, didn't see it in the commit until I put it up

src/elements/Typeahead/scripts/typeahead.js
halkon: Oh you know it.  This change alone however fixed 300 something tests

halkon: Errr allowed 300 something tests to run 😁

bower.json
```diff
@@ -6,29 +6,29 @@
...
     "moment": "~2.9.0",
     "html2canvas": "~0.4.1",
-    "angular-hotkeys": "chieffancypants/angular-hotkeys#~1.4.0",
+    "angular-hotkeys": "chieffancypants/angular-hotkeys#1.7",
     "font-awesome": "~4.4.0",
     "ng-debounce": "0.1.5"
   },
-  "devDependencies": {
-    "angular-mocks": "~1.3.20",
+  "devDependencies": [
```
keithcruz: Mixing a bracket and brace here

halkon: bahahaha, I was updating it with my phone

halkon:
Closing this as we don't really need a PR up just yet

utils/rx-page-objects/src/rxTabs.page.js
```diff
@@ -173,9 +173,9 @@ var tabs = {
     /**
      * @instance
      * @function
-     * @param {String} tabName - The tab to search for in the group of tabs.
+     * @param {String} tabName - The rxTab to search for in the rxTabset.
      * @returns {Boolean}
-     * @description Whether or not the tab by text `tabName` exists in the group of tabs.
+     * @description Whether or not the txTan by text `tabName` exists in the rxTabset.
```
Droogans: `rxTan`

grunt-tasks/templates/encore-ui.ejs.js
```diff
@@ -1,10 +1,14 @@
...
-    'cfp.hotkeys',
-    'ui.bootstrap',
     'encore.ui.tpls',
     'ngMessages',
     <%= _.uniq(config.srcModules).join(',\n    ') %>
 ])
+.config(function ($locationProvider) {
+    // Angular 1.6 changes the default value of the prefix to '!', this reverts to previous behavior
+    // https://github.com/angular/angular.js/commit/aa077e81129c740041438688dff2e8d20c3d7b52
+    $locationProvider.hashPrefix('');
```
CITguy: This configuration was moved to the distributed asset output, so that each consuming application doesn't have to apply this fix manually.

megalord:
Helpful reference to understand the changes: https://docs.angularjs.org/guide/migration

halkon:
Thanks @megalord 

Updated the main comment

megalord:
What do y'all think about aliasing the ported directives?  Basically defining both `rxTab` and `tab`, etc. with the same factory function.  The motivation would be to make it easier for any apps that have to migrate.  I guess it might just save a `sed` on each repo, but that is one step we'd save.

src/elements/Tooltips/scripts/rxTooltipHtmlPopup.js
```diff
@@ -0,0 +1,16 @@
+angular.module('encore.ui.elements')
+/**
+ * @ngdoc directive
+ * @name elements.directive:rxTooltipHtmlPopoup
```
julianbonilla: `s / rxTooltipHtmlPopoup / rxTooltipHtmlPopup`

halkon:
@megalord the services could be done, but I think we'll find it a challenge to alias the directives (unless there's some angular magic I know nothing of).

src/elements/Tabs/scripts/rxTabset.js
```diff
@@ -0,0 +1,19 @@
...
+ * @description
+ * Element for creating tabs.
+ */
+.directive('rxTabset', function () {
+    return {
+        restrict: 'EA',
+        transclude: true,
+        replace: true,
+        scope: true,
+        controller: 'rxTabsetController',
```
halkon: I don't think so.  Perhaps it was just how it was imported from `ui-bootstrap`. 

Also could be to clean up the clutter.

megalord:
It might be possible by pulling the DDO from the $injector and registering directly with $provide, but maybe the simple answer is the best:
```
['rxTab', 'tab'].forEach(function (name) {
  angular.module('...').directive(name, function () {
    ...
  });
});
```
You'd likely also have to modify the `require` section of the DDO if that gets used at all, but otherwise I can't think of any other places that refer to the directive name, can you?

halkon:
not at the moment, but that is a clever solution, which means, I LIKE IT

CITguy:
@megalord and @halkon, given that these changes will be going toward a 5.0.0 release (breaking changes) with a target release date of July 26:

* breaking changes are to be expected (new names for ported bootstrap directives, removing ngBootstrap, ng1.6, etc.)
* EncoreUI is being decommissioned and our resources to make additional updates are getting directed toward Helix work, so it's in our best interest to follow the simplest path (no need to add potential bugs with aliasing)

src/elements/Tooltips/scripts/rxTooltipHtmlPopup.js
CITguy: Good catch!

src/elements/Tabs/scripts/rxTabset.js
CITguy: The controller is pretty large, and it made more sense to separate it into Utilities. Otherwise, we would've included it with the directive.

megalord:
> breaking changes are to be expected

Yeah, that makes sense; it's certainly expected with a major release.  At the same time, having the ability to make a breaking change doesn't always mean we should.  Is the change worth the downstream impact it will have?  I'm just thinking ahead to if I were the person who had to go do all the ng 1.6 migrations, I'd prefer to make a hopefully small change to the framework rather than in every app.  Given that the framework and all the apps are owned by the same org, it makes sense to me to optimize our time and effort across all of them.  Having full standardization around the rx naming convention doesn't make me as happy as doing less work, but that is admittedly selfish.

To me, EncoreUI being decommissioned means that instead of having a centralized dev team, the ownership now falls to anyone working on the dozens of apps that still use it as a critical dependency.  There could be many reasons for more updates to the framework, e.g. ng 1.7 being released with a security-related patch.  When that time comes, the best strategy seems like the one that can be executed easiest/fastest.

As a single app owner, my stake here is pretty small.  I'm just hoping to represent the people who will be most affected by these decisions, many of whom are without a voice here afaik.  We're all just trying to do the best with the perspective we have, and whatever you decide will be fine!

src/utilities/rxPosition/styles/rxPosition.less
```diff
@@ -0,0 +1,19 @@
+.rx-position-measure {
```
cathysiller: Does anyone know if these styles breaks functionality with anything? @CITguy I know we discussed this file but can't recall what the solution was

cathysiller:
Dev LGTM, we did good!

src/utilities/rxPosition/styles/rxPosition.less
CITguy: Though it uses `!important`, this is probably fine as-is. Given that it was ported from ngBootstrap, there's probably a reason the devs on that project used it.

src/elements/Tooltips/scripts/rxTooltipHtmlPopup.js
CITguy: fixed

CITguy:
@megalord

> having the ability to make a breaking change doesn't always mean we should. Is the change worth the downstream impact it will have?

Having the ability to make breaking changes allows us to be fanatical stewards of our code. I believe we should take advantage of any opportunity that allows us to clean up or improve efficiency of existing code. The rx naming convention is a small step toward cleanup (with little to no impact of API functionality). 

The ported ngBootstrap code is a step toward more efficient assets (less javascript to load). Unfortunately, the ported ngBootstrap assets cannot be aliased with their old names, because it would conflict with official ngBootstrap assets for applications that still wish to use other functionality from that library. This is a breaking change that is necessary for us to be able to handle future angular releases (ng1.7+).

> To me, EncoreUI being decommissioned means that instead of having a centralized dev team, the ownership now falls to anyone working on the dozens of apps that still use it as a critical dependency.

As it is decommissioned, we will not be dropping support for EncoreUI anytime soon. It just means that no new elements will be added and we'll provide security and bug fixes only. Additional features to existing elements may still be possible, if the feature is absolutely necessary to provide functionality for existing applications. This will allow our resources to focus the majority of their time on a framework-agnostic component library, while still providing support for legacy applications.

Finally, the testing library (rxPageObjects) shouldn't hinder adoption of a 5.0 release. Your application's test code should guide you toward what needs to be corrected for a major version upgrade. As such, we've decided to hold off on including breaking changes to rxPageObjects, with the intent of having a more strategic discussion around potential changes after the 5.0 release.

I hope this helps alleviate some of your concerns.

halkon:
While understanding that when pushing forward we must adhere to quality and standardization, we must also look at the initial consumers (as well as future consumers) of these updates. 

Initially the work for this is for legacy UI's that barely have any developers supporting them (and on-going, upgrades will likely never happen). 

Given this, and the timeframes for these updates, we should perhaps go back to our original plan of 3.8.x -> 5.0 -> 6.0 

Giving 3.8 just the upgrade and removal of bootstrap-ui (with warnings of using the old directives, and letting them use the new directives).  

5.0 removal of old bootstrap directive names.

6.0 `rxPageObjects`?

The challenge of already having to upgrade to angular 1.6 is/should be enough for breaking changes, instead of imposing that they also have to make more changes due to directive renames. 

This is an open discussion, but I think we need to really weigh in the pros/cons of breaking changes introduced in this PR (and I am willing/wanting to do the work to not have so many breaking changes introduced)

halkon:
Also, I think we are re-aligning with Origin being the only one taking any framework upgrades, legacy apps are mainly in RTB mode.

jj4th:
QE 😸 👍 💯 ❤️

PR #1993, opened 2017-05-22
feat(rxTab): porting framework tab directive

JoseVillarreal:
JIRA: [FRMW-1348](https://jira.rax.io/browse/FRMW-1348)

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM
- [ ] Design LGTM
- [ ] QE LGTM


src/elements/Tabs/scripts/rxTab.js
```diff
@@ -0,0 +1,61 @@
...
+            heading: '@',
+            onSelect: '&select', //This callback is called in contentHeadingTransclude
+            //once it inserts the tab's content into the dom
+            onDeselect: '&deselect'
+        },
+        controller: function () {
+            //Empty controller so other directives can require being 'under' a tab
+        },
+        link: function (scope, elm, attrs, tabsetCtrl, transclude) {
+            scope.$watch('active', function (active) {
```
jonkruse00: @JoseVillarreal Can we format this for easier reading ? 
1. Collecting all variable assignments at the start of the link function
2. Collect all `event` watchers together. 
3. Encapsulate all code that happens on instantiation into a function then calling that function. 
```
activate ();
function activate () {
   if ( attrs.disable) {
      ....
   }
   tabsetCtrl.addTab(scope);
}
```

src/elements/Tabs/scripts/rxTab.js
```diff
@@ -0,0 +1,61 @@
...
+ * @ngdoc directive
+ * @name elements.directive:rxTab
+ * @requires elements.directive:rxTabset
+ * @restrict EA
+ * @param {Boolean=} active True to show active tab by default
+ * @param {String} heading Heading text to be displayed by default when rx-tab-heading tags are used
+ * @param {Boolean=} onSelect True when tab is selected
+ * @param {Boolean=} onDeselect True when tab is deselected
+ * @description
+ * Element for creating a tab.
```
jonkruse00: @JoseVillarreal Since this is a directive can we get 1 or more example of its usage ?

jonkruse00:
@JoseVillarreal Does `Encore-ui` have criteria for what gets spec test and what doesn't ? I noticed a few directives and a controller don't have them.

JoseVillarreal:
@jonkruse00 @halkon I was led to understand that we were merely porting the existing code to remove the external dependency, not refactoring the code.

I also don't know what the criteria for spec testing is for encore-ui, again, hopefully @halkon can clear that up.

halkon:
So we need to port their specs, keeping the tests they do as close as possible to what they have.  That being said we'll have to port them to `mocha/chai` as is, if there are missing tests for controllers/etc, we perhaps won't do them.  @cathysiller since you've done this already care to chime in?

src/utilities/rxTabsetController/scripts/rxTabsetController.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @ngdoc controller
+ * @name utilities.controller:rxTabsetController
+ * @description
+ * Controller for creating tabs.
+ */
+.controller('rxTabsetController', function ($scope) {
+    var ctrl = this,
+        tabs = ctrl.tabs = $scope.tabs = [];
+    var destroyed;
```
jonkruse00: @JoseVillarreal 
1. Why are you using both `ctrl` and `$scope` ?  `ctrl` should be sufficient. 
2. Also you shouldn't need to create `var tabs`,  just manipulate the `ctrl.tabs` itself.
3. Explicitly set `destroyed = false;` for readability. 
With those changes you can: 
```
var ctrl = this;
var destroyed = false: 
ctrl.tabs = [];
```

halkon: It is how ui-bootstrap was implemented @jonkruse00 
https://github.com/angular-ui/bootstrap/blob/0.14.x/src/tabs/tabs.js#L14

You will find a lot of this just looks and behaves and works the same as it was defined in the library

cathysiller:
@jonkruse00  -- Agreed with what @halkon and @JoseVillarreal  said, we're porting over the external specs to eliminate dependency and refactoring them with mocha/chai. For now our goal is to upgrade Angular successfully. Some of those specs may need further updates with Angular 1.6 in order to pass. 

As for criteria for specs in Encore-UI, have you found directives/controllers that may need further testing? Thanks!!! Hope that helps.

src/elements/Tabs/scripts/rxTab.js
cathysiller: @CITguy any feels about this?

src/utilities/rxTabsetController/scripts/rxTabsetController.js
```diff
@@ -0,0 +1,56 @@
...
+        angular.forEach(tabs, function (tab) {
+            if (tab.active && tab !== selectedTab) {
+                tab.active = false;
+                tab.onDeselect();
+                selectedTab.selectCalled = false;
+            }
+        });
+        selectedTab.active = true;
+        // only call select if it has not already been called
+        if (!selectedTab.selectCalled) {
```
jonkruse00: @JoseVillarreal
When you find the active tab above you set `selectedTab.selectCalled = false;` aren't you guaranteeing this if-block to be called ? So we could rewrite it such that :
``` 
angular.forEach(tabs, function (tab) {
              if (tab.active && tab !== selectedTab) {
                  tab.active = false;
                  tab.onDeselect();
                  selectedTab.active = true;
                  selectedTab.onSelect();
                  selectedTab.selectCalled = true;
              }
          });
```
Then would you really need  the `selectedTab.selectCalled` variable ?

src/elements/Tabs/scripts/rxTabset.js
```diff
@@ -0,0 +1,19 @@
...
+ * @restrict EA
+ * @requires utilities.controller:rxTabsetController
+ * @description
+ * Element for creating tabs.
+ */
+.directive('rxTabset', function () {
+    return {
+        restrict: 'EA',
+        transclude: true,
+        replace: true,
```
jonkruse00: @JoseVillarreal 
`replace` has been deprecated. 
https://docs.angularjs.org/api/ng/service/$compile

halkon: good catch!

src/elements/Tabs/scripts/rxTab.js
CITguy: I don't think we need to spend any time adding additional documentation to the ngDocs, considering that the code snippets in the example should suffice.

src/elements/Tabs/scripts/rxTab.js
CITguy: @jonkruse00: we shouldn't spend too much time refactoring code that we're porting. If we end up refactoring, we'll be adding more work than is necessary to obtain the same functionality.

src/elements/Tabs/scripts/rxTabset.js
CITguy: @jonkruse00: `replace` has been deprecated for a while. We don't have to worry about using it, considering that it will not be removed in ng 1.x

src/utilities/rxTabsetController/scripts/rxTabsetController.js
CITguy: As stated above, this is original code from ngBootstrap. Don't worry about refactoring.

src/utilities/rxTabsetController/scripts/rxTabsetController.js
```diff
@@ -0,0 +1,56 @@
...
+            selectedTab.onSelect();
+            selectedTab.selectCalled = true;
+        }
+    };
+
+    ctrl.addTab = function (tab) {
+        tabs.push(tab);
+        // we can't run the select function on the first tab
+        // since that would select it twice
+        if (tabs.length === 1 && tab.active !== false) {
```
jonkruse00: @JoseVillarreal 
This if-block is a bit confusing. 
1. Is it possible for `tab.active` to not be a boolean ? If no, then setting `tab.active = true` doesn't make sense because its already `true`.
2. Your last `else` sets it to false but to get to this else-block it should already be false.

jj4th:
That's a lot of skipped midways... what's going on here? @halkon

cathysiller:
Dev LGTM -- spoke with @JoseVillarreal about removing `src/tabs.page.js` and `tabs.midway.js`from utils/rx-page-objects folder since we are replacing with rxTabs in this pr

halkon:
@cathysiller I renamed em??

cathysiller:
@halkon dope, thanks!!

PR #2000, opened 2017-06-06
Frmw-1352 - notification

keithcruz:
JIRA: https://jira.rax.io/browse/FRMW-1352

### LGTMs
- [x] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [x] QE LGTM


jj4th:
rxNotify tests appear to be passing in this branch so.  QE-LGTM.

halkon:
@keithcruz I think I remember doing this work, and I'd like to perhaps to a vidyo review of this monday? What do you think?

I want to be really careful, as the idea before was that we would be able to add HTML elements into the notification bar.

halkon:
I am really scared specifically about having changed the unit test from `.eq(1)` vs `.eq(0)`

keithcruz:
@halkon That sounds good, I'll also look further into it in the mean time.

PR #2003, opened 2017-06-08
FRMW-1351 - multiselect

keithcruz:
JIRA: https://jira.rax.io/browse/FRMW-1351

### LGTMs
- [x] Dev LGTM
- [ ] QE LGTM

### DIFF
--- a/src/elements/Forms/scripts/rxMultiSelect.js
+++ b/src/elements/Forms/scripts/rxMultiSelect.js
@@ -112,7 +112,7 @@ angular.module('encore.ui.elements')
                     scope.preview = (function () {
                         function getLabel (option) {
                             var optionElement = rxDOMHelper.find(element, '[value="' + option + '"]');
-                            return optionElement.text().trim();
+                            return optionElement.text().trim() || option.trim();
                         }
 
                         if (_.isEmpty(scope.selected)) {
--- a/src/elements/Forms/scripts/rxMultiSelect.js
+++ b/src/elements/Forms/scripts/rxMultiSelect.js
@@ -41,7 +41,7 @@ angular.module('encore.ui.elements')
  * @param {String} ng-model The scope property that stores the value of the input
  * @param {Array=} options A list of the options for the dropdown
  */
-.directive('rxMultiSelect', function ($document, rxDOMHelper) {
+.directive('rxMultiSelect', function ($document, $timeout, rxDOMHelper) {
     return {
         restrict: 'E',
         templateUrl: 'templates/rxMultiSelect.html',
@@ -108,11 +108,11 @@ angular.module('encore.ui.elements')
             var ngModelCtrl = controllers[1];
 
             ngModelCtrl.$render = function () {
-                scope.$evalAsync(function () {
+                $timeout(function () {
                     scope.preview = (function () {
                         function getLabel (option) {
                             var optionElement = rxDOMHelper.find(element, '[value="' + option + '"]');
-                            return optionElement.text().trim() || option;
+                            return optionElement.text().trim();
                         }
 
                         if (_.isEmpty(scope.selected)) {
--- a/src/elements/Forms/scripts/rxMultiSelect.js
+++ b/src/elements/Forms/scripts/rxMultiSelect.js
@@ -112,7 +112,7 @@ angular.module('encore.ui.elements')
                     scope.preview = (function () {
                         function getLabel (option) {
                             var optionElement = rxDOMHelper.find(element, '[value="' + option + '"]');
-                            return optionElement.text().trim() || option.trim();
+                            return optionElement.text().trim() || option;
                         }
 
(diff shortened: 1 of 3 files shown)


src/elements/Forms/scripts/rxMultiSelect.js
```diff
@@ -112,7 +112,7 @@ angular.module('encore.ui.elements')
                     scope.preview = (function () {
                         function getLabel (option) {
                             var optionElement = rxDOMHelper.find(element, '[value="' + option + '"]');
-                            return optionElement.text().trim();
+                            return optionElement.text().trim() || option.trim();
```
halkon: Can you add a comment for why we need `|| option.trim()` in the code?

src/elements/Forms/scripts/rxSelectOption.js
```diff
@@ -29,9 +29,10 @@ angular.module('encore.ui.elements')
             value: '@'
         },
         require: '^^rxMultiSelect',
+        controller: function ($scope) {
+            $scope.transclusion = _.isEmpty($scope.$parent.options);
+        },
```
halkon: Can you explain the reasoning behind this?  Typically `controller` is only for exposing API's to other directives to use.  

Also why couldn't you do this in the `link` function?

src/elements/Forms/scripts/rxMultiSelect.js
```diff
@@ -112,7 +112,7 @@ angular.module('encore.ui.elements')
                     scope.preview = (function () {
                         function getLabel (option) {
                             var optionElement = rxDOMHelper.find(element, '[value="' + option + '"]');
-                            return optionElement.text().trim();
+                            return optionElement.text().trim() || option;
```
halkon: This feels really sketchy

src/elements/Forms/scripts/rxSelectOption.js
keithcruz: @halkon This comes back undefined in the link function.  Going to look into possible alternatives but for now this is a simple solution. I'll let you know what I come up with.

src/elements/Forms/scripts/rxMultiSelect.js
keithcruz: @halkon I agree, for some reason the transclusion appears empty on render when there are options already selected and the options binding is passed.  It is a bit of an edge case and I have a feeling it has to do with the eval async in the parent link function.  I'll look into this a bit deeper and update.

src/elements/Forms/scripts/rxSelectOption.js
halkon: Thank you, just feels weird

src/elements/Forms/scripts/rxMultiSelect.js
halkon: Yeah not sure, perhaps a `$timeout` might be better instead of the `$evalAsync` (it waits until things have  rendered). 

@CITguy @cathysiller do either of you have an opinion (or prior knowledge) on what may be happening here on `$render`?

src/elements/Forms/scripts/rxSelectOption.js
```diff
@@ -30,8 +30,7 @@ angular.module('encore.ui.elements')
         },
         require: '^^rxMultiSelect',
         link: function (scope, element, attrs, selectCtrl) {
-            scope.transclusion = rxDOMHelper.find(element, '[ng-transclude] > *').length > 0;
-
+            scope.transclusion = _.isEmpty(scope.$parent.options);
```
halkon: Can you add a comment here describing how it was done before and why this new way is better?

halkon:
Also you should rebase

keithcruz:
@halkon Addressed comments and rebased.

src/elements/Forms/scripts/rxSelectOption.js
```diff
@@ -30,8 +30,9 @@ angular.module('encore.ui.elements')
         },
         require: '^^rxMultiSelect',
         link: function (scope, element, attrs, selectCtrl) {
-            scope.transclusion = rxDOMHelper.find(element, '[ng-transclude] > *').length > 0;
-
+            // Previous implementation accessed the DOM and was always returning false after the upgrade to 1.6.
+            // By simply checking the scope's $parent we can avoid accessing the DOM and acheive the same result.
```
halkon: `achieve`

halkon: Can you also add a part explaining why the `options` in the parent essentially does the same as finding any transcluded elements?

keithcruz: 10-4

src/elements/Forms/scripts/rxSelectOption.js
halkon: perfect!

PR #2021, opened 2017-06-29
refactor(deprecations): Remove rxOptionTable and dependencies

cathysiller:
JIRA: https://jira.rax.io/browse/FRMW-1375

### LGTMs
- [x] Dev LGTM
- [x] Dev+Vidyo LGTM


demo/less/_shame.less
```diff
@@ -0,0 +1,34 @@
+.info-panel {
```
CITguy: File is unnecessary. Please remove.

cathysiller: removed

jj4th:
Doesn't appear to affect midways, but I looked it over anyway.  Looks good to me :smile_cat:

Side note, do we want to do this?  How does this affect legacy apps who need to consume this update.  What are the drawbacks of leaving this in?  @CITguy @parlarjb @halkon

PR #2028, opened 2017-07-31
chore(test): FRMW-1382 Update Protractor

jj4th:
JIRA: https://jira.rax.io/browse/FRMW-1382

Testing the viability of a protractor/node deps upgrade.

### LGTMs
- [ ] Dev LGTM
- [ ] Dev+Vidyo LGTM
- [ ] Design LGTM
- [ ] QE LGTM


jj4th:
Huh, I don't remember this PR.  going to rebase it anyway.

CITguy:
One reason Codeship failed is likely due to the hard coded `nvm install 5.11.1` in the project settings. You'll need to reconfigure codeship for this PR.

.node-version
```diff
@@ -1 +1 @@
-5.1.1
+8.2.1
```
CITguy: Does nvm support this dotfile for a `nvm install`?

Droogans: No, there's a pretty [good reason from the author here](https://stackoverflow.com/a/29545541/881224).

> `avn` supports `.node-version` and attempts to provide automatic version switching by hooking into `cd`, after `nvm` decided that was too invasive a behavior to include.

CITguy: Might we add a `.nvmrc` so that Codeship can install 8.2.1?

Droogans: I think it's better just to alter codeship's setup installation steps to use `nvm` directly, since it's provided.

https://documentation.codeship.com/basic/languages-frameworks/nodejs/#versions-and-setup

Not sure if this link works for you but https://app.codeship.com/projects/102410/tests/edit should allow you update the nvm version used, but that will need to be updated at the last possible moment before merging this into master to ensure other PRs don't try to use a newer version of node than what's supported.

CITguy: As it stands, the node version in the codeship configuration is disconnected from the requirements of the pull request.  I'd like to update the codeship configuration so that is installs the required node version for the PR itself, rather than for the project as a whole.

If we add an `.nvmrc` to the root of the repo defining 8.2.1, then all Codeship needs in its scripts is...

```shell
nvm install # same as "nvm install 8.2.1"
nvm use     # same as "nvm use 8.2.1"
```

because nvm natively supports `.nvmrc` for installing and using a default node version.

Droogans: Ah, I see now.

Yes, that makes a lot of sense. Keeps us from updating the codeship run script between merges in different branches.

jj4th:
We should look at the build scripts, it's running the unit tests like 7 times.

